PR #22, opened 2017-05-21
Improve query list formatting

maximbaz:
Hey, I improved a bit the formatting of the query list to distinguish numbers easier. What do you think? 🙂
I was also not sure, should I create PR against master or dev branch?

Before:

image

After:

image


Jguer:
I like it, will merge. Further on formatting, the "repo/name" should always begin on the same column (screenshots difference between results 9 and 10), would be interesting to implement.

Thanks

PR #31, opened 2017-06-17
Fix TimeUpdate param usage

maximbaz:
TimeUpdate is `false` by default. We want to compare modification time only when user changes this parameter to `true`. The current code seems to have inverted logic at the moment.

@Jguer please have a look, am I correctly interpreting this logic with TimeUpdate?

### DIFF
--- a/aur/aur.go
+++ b/aur/aur.go
@@ -167,7 +167,7 @@ func Upgrade(flags []string) error {
 
 		if _, ok := foreign[res.Name]; ok {
 			// Leaving this here for now, warn about downgrades later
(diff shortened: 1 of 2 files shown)


maximbaz:
I had another confusion when I saw in `yay.json` the `versionupdate` config, which matches to `timeupdate` property in the code. They have inverse meaning! 🙂 So I aligned the name in config to match the property name.

Jguer:
I finally have some time today, I'll take a look at it. On first glance it seems right.

Thanks Maxim

PR #37, opened 2017-07-06
Fill default config before trying to unmarshal config file

kshlm:
This keeps the default config values available if the config file only
has a few options.

Also, if the config is not found, save the default config in its place.
And pretty print it when saving to make it easier to edit for users.


Jguer:
Everything working here, well thought out.

Thanks kshlm

PR #44, opened 2017-09-05
Suffix BuildDir with uid to avoid perm problems

afg984:
Different users using the same `/tmp/yaytmp` has caused permission problems when another user tries to use yay when yaytmp is created by another user.

This PR changes `/tmp/yaytmp` to `/tmp/yaytmp-1000`, where 1000 is the uid.

This method is also used by Kwpolska/pkgbuilder and keenerd/packer.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"os/user"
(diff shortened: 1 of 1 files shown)


Jguer:
Great commit,
That seems like a clean way of doing it.

PR #68, opened 2017-11-27
Using '-' doesn't raise an error

PietroCarrara:
running 
```
yay -
```
is now safe


Jguer:
It was already safe on the master branch thanks to a length check but your way is much more go than the C approach in use. Will merge, I love commits that make the code more go.

PR #83, opened 2017-12-30
Count from 1 instead of 0 when searching packages

Morganamilo:
I feel like starting from 1 is much better ergonomically while not using
a numpad. This could be added as a config option instead if you would
prefer.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -323,11 +323,11 @@ func numberMenu(pkgS []string, flags []string) (err error) {
 	}
 
 	if config.SortMode == BottomUp {
-		aq.printSearch(numpq)
+		aq.printSearch(numpq + 1)
 		pq.printSearch()
 	} else {
 		pq.printSearch()
-		aq.printSearch(numpq)
+		aq.printSearch(numpq + 1)
 	}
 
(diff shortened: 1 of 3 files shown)


maximbaz:
Over time I got used to this, but I remember being confused and uncomfortable with yay starting indexes from zero, so I support this change. Config is probably an overkill for this :)

Jguer:
@maximbaz I'm away from my dev environment at the moment, can you test that this patch does not create any conflicts and selects the right package on searches containing only AUR, only repo and AUR and repo?

Otherwise I'll only be able to review it in 7 days.

Morganamilo:
When I submitted this I tested:

- 0 (invalid)
- Number packages  + 1 (invalid)
- The first package
- Both packages on the aur/repos boundary

All seemed to work as expected.

maximbaz:
Testing now, will get back in a few moments.

maximbaz:
Works well in all 3 cases, the patch itself looks good to me too.

Morganamilo:
I forgot to mention I also did test all these cases on both --bottomup and --topdown. so it should be good for those too

PR #84, opened 2017-12-30
Fix crash when package version is "invalid"

### LINKED ISSUE #78: Yay crashes when parsing package name
Here's the output:
```
:: 8 Packages to upgrade.
7  extra/libdrm                                             2.4.88-1 -> 2.4.89-1
6  community/dbus-c++                                        0.9.0-7 -> 0.9.0-8
5  community/docker-compose                                 1.17.1-1 -> 1.18.0-1
4  community/girara                                          0.2.7-1 -> 0.2.8-1
lash invalid version format: 0.6.0~rc2-10
lash invalid version format: 0.6.0~rc2-11
3  community/lash                               panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x63da68]

goroutine 1 [running]:
main.upSlice.Print(0xc420210000, 0x8, 0x8, 0x0)
        /home/psafont/.cache/pacaur/yay/src/yay-2.219/upgrade.go:82 +0x488
main.upgradePkgs(0x0, 0x0, 0x0, 0x0, 0x0)
        /home/psafont/.cache/pacaur/yay/src/yay-2.219/upgrade.go:293 +0x1de
main.main()
        /home/psafont/.cache/pacaur/yay/src/yay-2.219/cmd.go:264 +0x898
```

I'm using yay v2.219

I would recommend to fail in a more graceful name. For example, to a string comparison.

Morganamilo:
This will fix #78. And keep the formatting if a package version is invalid.

Before:
image

After:
image

The root of this issue comes from gopkgbuild which seems to throw an error whenever a pkver contains a `~` (and possibly other symbols) which should be a valid character.

Currently yay should work as expected apart from displaying "Invalid Version"  for effected packages. Once this is fixed in gopkgbuild it should display normally.

### DIFF
--- a/upgrade.go
+++ b/upgrade.go
@@ -60,14 +60,9 @@ func (u upSlice) Less(i, j int) bool {
 // Print prints the details of the packages to upgrade.
 func (u upSlice) Print(start int) {
 	for k, i := range u {
-		old, err := pkgb.NewCompleteVersion(i.LocalVersion)
-		if err != nil {
-			fmt.Println(i.Name, err)
-		}
-		new, err := pkgb.NewCompleteVersion(i.RemoteVersion)
-		if err != nil {
-			fmt.Println(i.Name, err)
-		}
+		old, errOld := pkgb.NewCompleteVersion(i.LocalVersion)
+		new, errNew := pkgb.NewCompleteVersion(i.RemoteVersion)
+		var left, right string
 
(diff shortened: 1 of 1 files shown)


Morganamilo:
I fixed this upstream. gopkgbuild should also be updated.

PR #89, opened 2018-01-04
Refactoring and rewriting the argument parsing system 

### LINKED ISSUE #62: Argument passing broken
I'm trying to build AUR packages from within pacstrap in a script:

`env PATH=/tmp/pacutilj396ttt0:$PATH sudo /usr/bin/pacstrap -c -G -M -d /tmp/pacutilj396ttt0/chroot/aic94xx-firmware aic94xx-firmware
`

`/tmp/pacutilj396ttt0/pacman` is a wrapper script where I call yay, disable updating of the db, and give a custom db path:
`sudo -u hrehfeld /usr/bin/yay ${@/'-Sy'/-S} --noconfirm --dbpath /tmp/pacutilj396ttt0/tmp-pacman -dd`

However, yay doesn't handle -dd and (if you switch the order) --dbpath correctly:

```
sudo -u hrehfeld /usr/bin/yay -r /tmp/pacutilesu2q6hw/chroot/aic94xx-firmware -S aic94xx-firmware --noconfirm -b /tmp/pacutilesu2q6hw/tmp-pacman -dd
error: invalid option '-d'
```

```
sudo -u hrehfeld /usr/bin/yay -r /tmp/pacutiladk6ol6n/chroot/aic94xx-firmware -S aic94xx-firmware --noconfirm --dbpath /tmp/pacutiladk6ol6n/tmp-pacman -dd
/usr/bin/pacman: option '--dbpath' requires an argument
```

Morganamilo:
This started out as an attempt to fix #62 (which is funny given that its still does not work fully) but strayed into a total redesign of the argument parsing system, which required a lot of changes to the code.

Currently there is a bit of a mess with the mix of old and new but generally it is better than it was before and should be much more versatile to user input:
        `-Syu', '-S -y -u', '--sync --refresh --sysupgrade`
Should all work as expected, as well as the parsing of other arguments befond S, y and u.

Features added/fixed/changed:
- Yay now only calls sudo if root is needed
- Yay will ensure more than one operation has not been selected
- Yay will handle arguments that take arguments correctly  (e.g. -b requires a path) 
- The alpm handle will respect the command line (--config, --root, --dbpath, ect.)
- Major refactoring, should be more scalable for further argument parsing
- Change some command line arguments to be more consistent with pacman's usage (see updated usage in cmd.conf)
-  Always make sure relevant arguments are always passed to pacman  (--config, --root, --dbpath, ect.)
- Always handle fatal errors in main and ensure the relevant cleanup is called even if an error occurs (free alpmhandle, save config, ect.)
- Yay will treat everything after `--` as a literal target
- Yay will try to read from stdin after a `-`

Things that still need to be done before merging:
- [x] Update manpage to reflect usage changes
- [ ] update shell completion to reflect usage changes

Things that need to be done eventually:
- [ ] Ensure calls to makepkg also respect command line arguments
- [ ] Convert more code to use the new  cmdArgs and simlar objects instead of converting back to slices
- [ ] More testing, can be done after merge but should be done before release
- [ ] Move some options from config.go into the argument parser and use config.go for options that save to disk
- [ ] Add a more detailed yay specific help for `yay -Yh`

Bugs that exist but are not important enough to require a fix right now:
- [ ]  --cachedir will not work for multiple cachedirs
- [ ]  conflicting operations (e.g. using -i and -s on sync) will pass by without error
- [ ] reading from stdin does not work as expected and will break keyboard input, should work as expected with --noconfrim
- [ ] passing an argument twice will work (-dd, --nodeps --nodeps) will work as expected but mixing them (--nodeps -d) will not work when yay tries to check if it should skip deps, passes  to pacman will work as expected

These should be opened as issues at some point.

I realise iv'e gone and changed quite a lot here although i believe that's for the better. I suggest you give the commit messages a look. Feedback would be greatly appreciated

### DIFF
--- a/clean.go
+++ b/clean.go
@@ -32,7 +32,7 @@ func removeVCSPackage(pkgs []string) {
 }
 
 // CleanDependencies removes all dangling dependencies in system
-func cleanDependencies(pkgs []string) error {
+func cleanDependencies() error {
 	hanging, err := hangingPackages()
 	if err != nil {
 		return err
@@ -49,12 +49,15 @@ func cleanDependencies(pkgs []string) error {
 }
 
 // CleanRemove sends a full removal command to pacman with the pkgName slice
-func cleanRemove(pkgName []string) (err error) {
-	if len(pkgName) == 0 {
+func cleanRemove(pkgNames []string) (err error) {
+	if len(pkgNames) == 0 {
 		return nil
 	}
+	
+	arguments := makeArguments()
+	arguments.addArg("R", "s", "n", "s", "noconfirm")
+	arguments.addTarget(pkgNames...)
 
-	//TODO
-	//err = passToPacman("-Rsnc", pkgName, []string{"--noconfirm"})
+	err = passToPacman(arguments)
 	return err
 }
--- a/clean.go
+++ b/clean.go
@@ -55,7 +55,7 @@ func cleanRemove(pkgNames []string) (err error) {
 	}
 	
 	arguments := makeArguments()
-	arguments.addArg("R", "s", "n", "s", "noconfirm")
+	arguments.addArg("R", "noconfirm")
 	arguments.addTarget(pkgNames...)
 
 	err = passToPacman(arguments)
--- a/clean.go
+++ b/clean.go
@@ -54,6 +54,7 @@ func cleanRemove(pkgName []string) (err error) {
 		return nil
 	}
 
-	err = passToPacman("-Rsnc", pkgName, []string{"--noconfirm"})
+	//TODO
+	//err = passToPacman("-Rsnc", pkgName, []string{"--noconfirm"})
 	return err
 }
--- a/cmd.go
+++ b/cmd.go
@@ -55,10 +55,9 @@ If no operation is provided -Y will be assumed
 `)
 }
 
-func init() {
+func initYay() (err error){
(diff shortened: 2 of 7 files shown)


Morganamilo:
As a side note. Given that this was originally a fix for #62.

`yay -S aic94xx-firmware --noconfirm -b /tmp/pacutilesu2q6hw/tmp-pacman -dd
error: invalid option '-d'`

Does now succeed but not do the entirely right thing.

it will:
1. see `-S` and run handleSync()
2. see ` aic94xx-firmware` as the target to install
3. see `-b /tmp/pacutilesu2q6hw/tmp-pacman` and use it as the database location
4. see `-dd` and check skip the dependency check (if it was checking deps it would correctly check against the correct databse specified in -b)
5. start the install process and download the PKGBUILD
6. now is when things go wrong, makepkg is called and has no idea about the arguments to yay so it will use the default database, default pacman.conf, ect.

Morganamilo:
Seems you merged this already. I was looking for feedback on it before I continued with it as some cleanup/bugfixes are needed. I wouldn't say its worth reverting this as long as it's all worked out before next release it doesn't really matter. Maybe it's a good thing to have it in master so other people can help test and contribute to it. Just wanted to give a fair warning.

PR #92, opened 2018-01-06
fix query info with -Si

l4foss:
this PR fixes:
 - error handling on passToPacman
 - <s>typo on handling -Qi command</s>


query.go
```diff
@@ -158,6 +158,15 @@ func syncInfo(pkgS []string, flags []string) (err error) {
 		return
 	}
 
+	//repo always goes first
+	if len(repoS) != 0 {
+		err = passToPacman("-Si", repoS, flags)
+
+		if err != nil {
+			return
```
Morganamilo: If -Si errors don't you think we, should continue on anyway?

l4foss: I don't think we should, I don't see any circumstance in which we should continue in case of error
Could you point it out to me? Or aren't they just the same?

Morganamilo: I was thinking how pacman has an exit status of 1 if any of the packages fed into -Si don't exist. But i forgot that's handled earlier on with  `aurS, repoS, err := packageSlices(pkgS)`. So yeah I think you're right, we should exit on error.

config.go
```diff
@@ -221,3 +221,4 @@ func passToPacman(op string, pkgs []string, flags []string) error {
 	err := cmd.Run()
 	return err
 }
+}
```
Morganamilo: You have a duplicate bracket here.

config.go
```diff
@@ -194,4 +194,30 @@ func continueTask(s string, def string) (cont bool) {
 	}
 
 	return true
-}
\ No newline at end of file
+}
+
+// PassToPacman outsorces execution to pacman binary without modifications.
+func passToPacman(op string, pkgs []string, flags []string) error {
```
Morganamilo: Looks like you merged this incorrectly. This function was moved but your branch has both. Have you actually checked your branch compiles? It does not compile for me at least.

l4foss: Sorry let me check it again

Morganamilo:
Seems to work now, as you saw the code kinda changed a lot after your original pr. Still these changed are still needed so happy to have them. One suggestion, you might want to rebase a little bit to clean the commit history but that just comes down to my personal ocd so feel free to ignore.

I'd say this is ready to merge now.

l4foss:
Thank you
I didn't notice the new commits so I messed everything up

Morganamilo:
Don't worry about it, all worked out in the end.

PR #97, opened 2018-01-14
Support for ranges when selecting packages

samosaara:
- Also, minor formatting to cmd.go

Adds the support to select a range of packages, (e.g.: `1-10` all packages from 1 to 10) when selecting packages through `numberMenu()`


Morganamilo:
Could you also add the invert logic that Pacman supports? 
`^1` = everything but 1
`^2-6` = everything outside of the range

Also you might want to add this to the -Syu menu where it asks what packages to not upgrade.

samosaara:
Ok I can easily replicate the logic in the upgrade part, but about the negate part, just so I can be clear:
- When there are only negate selections, select everything, and then remove the specified
- When with both kinds of selections, only remove from the specified by the normal selections/ranges?

Morganamilo:
Hmm never really thought about it. Seems like it you have a positive selection it ignores the negative selection. Its built into Pacman so feel free to `pacman -S base` and play around with it.

Morganamilo:
Some of my big changes just got merged (albeit a little earlier than I would have liked). You've got some merge conflicts now sorry :(

Morganamilo:
Done some tests and it seems to work fine. Hope to see it merged soon.

Issue #90, opened 2018-01-06
Confirm with "j" in german


Musikermomo:
"Yes" means "Ja" in German, so you should confirm with  "j" as in pacman:
`sudo pacman -S cowsay`
[...]
` (1) cowsay-3.03-9`
[...]
`:: Installation fortsetzen? [J/n] `

in yay:
`Edit PKGBUILD? [y/N]`

Alternating between pressing "y" and "j" is very confusing while doing a system upgrade.
Accepting "y" and "j" would solve this issue I think.

Many thanks for yay btw!

samosaara:
I think this could be a better opportunity then simply translating to german, This would be a great opportunity to internationalize the project to multiple languages, like Portuguese (😉 😉 @Jguer ) ... But that would demand a rework of the messages system if they're hard-coded, (which I suspect them to be) I would love to help... But that would demand a bunch of design decisions... Like where to store the strings file, the format of the files, how to parse them, a naming scheme for the messages... What's @Jguer opinion to this?

Jguer:
Every string is hardcoded in yay. I'm not the biggest fan of software localisation but yay is a wrapper of a program that is indeed localized.

And the J and the Y really would disturb me if it happened to me so I'm solidary with the idea of localizing at least the response commands.

On the technical side, a map seems like the easiest way to make this work and loading the map from the desired language.

wzrdtales:
this seems to be still not fixed

PR #103, opened 2018-01-18
Localizing yay!

samosaara:
I pretend to start to support a localization system to yay as a followup to #90, currently it is simply a stub without the strings actually put in place. But I wanted to open the PR so we can discuss on the how, and also for me to be able to build with your guys feedback.

I've looked on specialized ways to translate like GNU gettext, XML resources, but given go build and distribution mechanisms, and the ecosystem of the tools that didn't seem very mature, without even mentioning the inherent complexity such big tools bring, I chose to do it programmatically, with go's map literals. 

The obvious downside of that approach is that it might be too complex to translators to contribute, given they needed to understand (even though simple) golang's syntax, and (AFAIK, never used them) we also won't be able to tap into the established ecosystem of translation crowd sourcing tools such as [pootle](http://pootle.translatehouse.org/) (even though an argument could be made that such tools could be also be too complex in and of themselves).

But anyway I did come up with a simple stub in go, any comments, thoughts? May I actually start mapping the strings?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,9 @@
 *.so
 
 # Folders
-_obj
-_test
+_obj/
+_test/
+.vscode/
 
 # Architecture specific extensions/prefixes
 *.[568vq]
--- a/translations/english.go
+++ b/translations/english.go
@@ -1,0 +1,16 @@
+package translations
+
+var enUsMap = map[string]string{
+	"yes": "Y",
+}
+
+func (l langauge) EnUS() Translation {
+	return Translation{
+		Code:   "en_US",
+		values: enUsMap,
+	}
+}
+
+func (l langauge) En() Translation {
+	return l.EnUS()
+}
--- a/translations/language_selector.go
+++ b/translations/language_selector.go
@@ -1,0 +1,75 @@
+package translations
+
+import (
+	"fmt"
+	"os"
+	r "reflect"
+	"strings"
+	"unicode"
+)
+
+// Translation represent a translation in a language
+type Translation struct {
+	Code   string
+	values map[string]string
+}
+
+// Stub struct for grouping the Translation constructors together for reflection
+type langauge struct{}
+
+// findLang is the helper function to determine the langauge based on the
+// envoirment
+func findLang() string {
+	lang := os.Getenv("LANG")
+	if lang == "" {
+		lang = os.Getenv("LC_ALL")
+		if lang == "" {
+			// TODO Filter with --verbose
+			fmt.Printf("Could not determine the LANG based on envoirment, %s",
+				"Falling back to english.")
+			return "en_US"
+		}
+	}
+
+	encodingDot := strings.IndexRune(lang, '.')
+	if encodingDot > -1 {
+		lang = lang[:encodingDot]
+	}
+	return lang
+}
+
+// GetTranslation gets the current translation struct based on envoirment
+func GetTranslation() Translation {
+	lang := findLang()
+	// Removes the underscore in the ISO format because we can't create
+	// methods with underscores
+	lang = lang[:2] + lang[3:]
+	// Method has to be public
+	lang = string(unicode.ToUpper(rune(lang[0]))) + lang[1:]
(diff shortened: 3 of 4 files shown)


Morganamilo:
I still think this is a bit overly complicated and might be a way to halt development. If some one adds a feature that has new text what do we do? refuse the feature until we translate all the strings for the feature? Use the English string as a fallback and add translations over time? If some one wants to tweak the phrasing of something then they will also have to do this in x more languages.

Down the line it could be a good idea when Yay is more mature and feature complete and less likely to change.

As for my general thoughts on implementation the first problem I see is we have `go-alpm` `gopkgbuild` `aur` as dependencies and most of the time if any of them fail we take the error string it provides verbatim. So we would have to deal with translating the dependencies or generating out own string on an API error.

Ignoring the problems my thoughts on implementation off the top of my head would be a language struct.
```go
type language struct {
    NumbersToInstall string
    CantFindPackage string
    ....
}
```

Which we build for each language.

```go
english := language {
    NumbersToInstall: "Type numbers to install. Separate each number with a space.",
    CantFindPackage: "Cant find package %s",
    ....
}
```

Then detect the language we require and set that globally.

```go
currentLang = getLang()
```

Where currentLang would be a pointer to the English struct if we were using English.

Then using it would be as simple as.

```go
fmt.Printf(currentLang.CantFindPackage, pkg)
```

samosaara:
Oh please, don't get me wrong, I would hate to transform a feature into a chore, I'm definitely not trying to overly-complicate the system or anything, I intend it to make as unobtrusive as possible, and just falling back to english is indeed my number one way to do that.

The system you proposed was pretty much exactly like my stub implementation, just that I preferred a hash map over a structure value because it is much, much more flexible, like allowing to do stuff like fallback-language even though it doesn't provide 'type safety'. or convenient auto complete.

Morganamilo:
With the fallback method I feel like we would end up with a complete English translation and all other translations half done. I don't really know how many people use Yay so I cant really say what the size of the user-base is but going from the GitHub contributors getting translators might be hard. Verifying translations would also be very difficult.

I don't think there's enough gain to warrant a system like this. But it might be good to have this in a separate branch so it can be developed for when it's ready and needed. So feel free to develop it further, I might even contribute to the framework but I'll be unable to translate anything. I'm not an authority on this project though just voicing my views.

Morganamilo:
Also a bit early for implementation but I would probably much prefer structs over maps for this, both for stricter compile time checks and speed. Fallbacks could still be done as.

```go
french := language {
    NumbersToInstall: "some french stuff",
    CantFindPackage: english.CantFinePackage //fallback
    ....
}
```

Although that would call the string to be duplicated because they are pass by value so maybe its not the best solution.

samosaara:
> I feel like we would end up with a complete English translation and all other translations half done.

Yes I also agree that is very likely. But I can't see why this would be bad... people would still have a mostly localized yay that can use even though it is not translated fully. Something is better than nothing.

Well, I still believe that a structure are way too limiting (and also very redundant and verbose to implement fallbacks), just for the sake of being convenient, well if you still want correctness, we could create an father object containing all the possible translation keys as in a ENUM

simon04:
Why not make use of [an existing library](https://github.com/search?o=desc&q=i18n+language%3Ago&s=stars&type=Repositories) such as https://github.com/nicksnyder/go-i18n?

Jguer:
I agree on pulling a new lib for localisation, I think shell completions are not a priority right now to translate (and they're slow as it is already)

Morganamilo:
What about this? https://github.com/leonelquinteros/gotext

Jguer:
From my research I also concluded that would be the best one (sweet mother of bloat).

Morganamilo:
Well if you care about bloat so much we can always rewrite Yay in c. ;)

I'll look into setting up some sort of working example.

Jguer:
I know, I know, no point crying over one more lib when you have ~600 MiB of `noto-fonts` installed

Jguer:
Closing this in favour of #342

Issue #65, opened 2017-11-15
Can't reply Y to pacman questions


papey:
Sometimes a package is replaced by another and pacman asks user is this package should be replaced or not (y/N question).

With yay, this behavior is broken and yay pass N to pacman and update fails.

Jguer:
This is due to the default update behavior, I believe checking for package
conflicts before presenting the update menu and offering the option to
replace then should fix the problem.

On Nov 15, 2017 5:58 PM, "Wilfried OLLIVIER" <notifications@github.com>
wrote:

Sometimes a package is replaced by another and pacman asks user is this
package should be replaced or not (y/N question).

With yay, this behavior is broken and yay pass N to pacman and update fails.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/Jguer/yay/issues/65>, or mute the thread

papey:
Ok, I will try to submit a PR (not sure about the time I have to work on this)

Issue #27, opened 2017-06-11
Download everything and ask all the questions first, then build and install


maximbaz:
_I saw you have this on the [projects](https://github.com/Jguer/yay/projects/1) page, I'm creating this issue for tracking purposes, something to subscribe to, in order to get updates when this works begins._

When installing or upgrading packages, it would be really nice to get all the questions asked first (do I want to edit PKGBUILD, install dependencies, confirm conflicting package replacement, etc.), answer to all of them in one go and let the installation proceed without further interruption.

Jguer:
Yes, I wanted to get that working before releasing `yay v2` but exam season hit a little earlier than expected.
 
- Ask all edit PKGBUILD
- Install all dependencies (Package conflicts are all handled by pacman I think)
- AUR dependencies are installed with --noconfirm on pacman but
- final built packages are moved to `${yay-build-dir}` and one last confirm to install them all.

rmarquis:
>  (Package conflicts are all handled by pacman I think)

Unfortunately, you'll have to handle this by yourself then bypass pacman prompts if you want to ask all the questions first - otherwise you'll have to deal with a yaourt-kind of linear prompts through the whole building process.

AladW:
Even if you ask all questions first, you might have to confirm 100's of times on large dependency chains. Some better approach is to have the user deal with it (by having a "download only" stage, see https://github.com/trizen/trizen/issues/8) or integrate with a suitable file manager.

Also the issue with conflicts pretty much disappears when you use a local repo (I've hardly seen them building 2000+ packages anyway), but I guess that's out of scope for this issue.

maximbaz:
True, although in my personal experience I haven't seen such long dependency chains, I think in 99% of cases you don't need to confirm more than 10 packages at once, and having an integrated approach to answer up to 10 questions and let yay proceed would be very useful 🙂

132ikl:
Any ETA on this? Liking yay as an alternative to the now unmaintained pacaur but this is definitely my biggest issue with yay.

Mayurifag:
+1 here.

Also, sorry for offtopic, but is there any way to list searched packages the same way like it was in yaourt — I meant, starting from first number and list all them from top to bottom?

maximbaz:
@Mayurifag if you still seek the answer to the offtopic question: `--topdown` will change the sorting order, and since https://github.com/Jguer/yay/pull/83 the numbering always starts from `1` (not released yet, only in master).

[...]

Issue #10, opened 2017-02-15
Is yay a "reliable solver" of complex depends


SolarAquarion:

Jguer:
Not yet. Soon™.

https://wiki.archlinux.org/index.php/AUR_helpers#Comparison_table
The objective is to have it all green except git clone.

AladW:
I'm curious as to why not git clone? There's many benefits over regular tarballs, like git log, git diff, etc.

Jguer:
I have a local branch experimenting git clone but I feel it is slower. Noticeably slower, even using `--depth=1` on different machines, the tarball+untar approach ends up being faster. 
So I've been sitting on it unsure if I should merge it or not.

`err = exec.Command("git", "clone", "--depth=1", fmt.Sprintf("https://aur.archlinux.org/%s.git", a.PackageBase), dir).Run()`

AladW:
The way I see it downloading the package files only makes up a small part of the whole build, so imo it doesn't weigh up against the advantages. You can also make it optional or some such.

AladW:
To bump the dependency issue, what's preventing yay from being able to solve dependencies reliably? Should be as simple as recursively querying the RPC and running the results through tsort. Example implementation: https://github.com/AladW/aurutils/blob/master/bin/aurchain

z3ntu:
I just tried to install a package I am maintaining in the AUR and it fails to install with yay (also fails with yaourt btw). Works fine with pacaur though. Link to the package (can be used as a kind of benchmark 😉 if solver etc is working correctly): https://aur.archlinux.org/packages/openrazer-meta

z3ntu:
I still can't install openrazer-daemon with yay:
```
==> Finished making: openrazer 2.2.1-1 (Tue Feb 27 16:57:41 UTC 2018)
==> Cleaning up...
loading packages...
resolving dependencies...
warning: cannot resolve "python-notify2", a dependency of "openrazer-daemon"
:: The following package cannot be upgraded due to unresolvable dependencies:
      openrazer-daemon

:: Do you want to skip the above package for this upgrade? [y/N] 
error: failed to prepare transaction (could not satisfy dependencies)
:: openrazer-daemon: requires python-notify2
exit status 1
```

Morganamilo:
Try the git version if you are not already.

Morganamilo:
[...]

PR #105, opened 2018-01-20
New install algorithm

Morganamilo:
```I have replaced the old install and dependancy algorithms with a new
design that attemps to be more pacaur like. Mostly in minimizing user
input. Ask every thing first then do everything with no need for more
user input.

It is not yet fully complete but is finished enough so that it works,
should not fail in most cases and provides a base for more contributors
to help address the existing problems.

The new install chain is as follows:
	Source info about the provided targets
	Fetch a list of all dependancies needed to install targets
		I put alot of effort into fetching the dependancy tree
		while making the least amount of aur requests as
		possible. I'm actually very happy with how it turned out
		and yay wil now resolve dependancies noticably faster
		than pacaur when there are many aur dependancies.
	Install repo targets by passing to pacman
	Print dependancy tree and ask to confirm
	Ask to clean build if directory already exists
	Download all pkgbuilds
	Ask to edit all pkgbuilds
	Ask to continue with the install
	Download the sources for each packagebuild
	Build and install every package
		using -s to get repo deps and -i to install
	Ask to remove make dependancies

There are still a lot of things that need to be done for a fully working
system. Here are the problems I found with this system, either new or
existing:
	Formating
		I am not so good at formatting myself, I thought best to
		leave it until last so I could get feedback on how it
		should look and help implementing it.
	Dependancy tree
		The dependancy tree is usually correct although I have
		noticed times where it doesnt detect all the
		dependancies that it should. I have only noticed this
		when there are circular dependancies so i think this
		might be the cause. It's not a big deal currently
		because makepkg -i installed repo deps for us which
		handles the repo deps for us and will get the correct
		ones. So yay might not list all the dependancies. but
		they will get installed so I consider this a visual bug.
		I have yet to see any circular dependancies in the AUR
		so I can not say what will happend but I#m guessing that
		it will break.
	Versioned packages/dependencies
		Targets and dependancies with version constriants such
		as 'linux>=4.1' will not be checked on the aur side of
		things but will be checked on the repo side.
	Ignorepkg/Ignoregroup
		Currently I do not handle this in any way but it
		shouldn't be too hard to implement.
	Conflict checking
		This is not currently implemented either
	Split Paclages
		Split packages are not Handles properly. If we only
		specify one package so install from a split package
		makepkg -i ends up installing them all anyway. If we
		specify more than one (n) package it will actually build the
		package base n times and reinstall every split package
		n times.
	Makepkg
		To get things working I decided to keep using the
		makepkg -i method. I plan to eventually replace this
		with a pacman -U based method. This should allow passing
		args such as --dbpath and --config to aur packages
		aswell as help solve some problems such as the split
		packages.
	Clean build
		I plan to improve the clean build choice to be a little
		more smart and instead of check if the directory exists,
		check if the package is already build and if so skip the
		build all together.
```

Here is an example of the install process,  printing is non final and needs to be improved.
Note makepkg has been replaced by a dud program that prints the command issued as to simplify the output for this example.
 
image

image

This pr relates to #10 #27 #65 although might not directly fix them just yet. But they should be kept in mind. I will probably open an issue linking here and explaining the issue again after this is merged as a way to keep track of progress with this system.

### DIFF
--- a/install.go
+++ b/install.go
@@ -1,17 +1,28 @@
 package main
 
+
 import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
 
 	rpc "github.com/mikkeloscar/aur"
 	gopkg "github.com/mikkeloscar/gopkgbuild"
+	alpm "github.com/jguer/go-alpm"
 )
 
-// Install handles package installs
+// Install handles package installs	
 func install(parser *arguments) error {
-	aurs, repos, _ := packageSlices(parser.targets.toSlice())
+	aurs, repos, missing, err := packageSlices(parser.targets.toSlice())
+	if err != nil {
+		return err
+	}
+
+	if len(missing) > 0 {
+		fmt.Println(missing)
+		return fmt.Errorf("Could not find all Targets")
+	}
 
 	arguments := parser.copy()
 	arguments.delArg("u", "sysupgrade")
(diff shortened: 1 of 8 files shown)


Jguer:
I like what I see 😀 , I'll review it as soon as possible.

samosaara:
Wow a friend of mine @felipeac will have a stroke if this works, he always used to complain about how much he loved this pacaur feature and how [pacaur is now unmaintained](https://www.reddit.com/r/archlinux/comments/7k5suz/pacaur_now_unmaintained/), we can even end up filling a void

Morganamilo:
@samosaara It's still a little broken right now, in fact I realised I kinda broke the number menu which needs to be fixed. I'll be away for the next day or two though sorry. Feel free to fix it yourself, after that the git branch should be good enough for every day use if you decided to use it.

Morganamilo:
I took a look and it was simple enough to fix in five minutes so I went ahead and did it myself in #107

maximbaz:
Wow this is astonishingly cool, great job @Morganamilo! 

> To get things working I decided to keep using the makepkg -i method. I plan to eventually replace this with a pacman -U based method...

So much looking forward to this!

install.go
```diff
@@ -27,147 +38,216 @@ func install(parser *arguments) error {
...
+		fmt.Println()
+
+		fmt.Println()
+	
+		askCleanBuilds(dc.AurMake)
+		askCleanBuilds(dc.Aur)
+
+		fmt.Println()
+
+		if !continueTask("Proceed with download?", "nN") {
```
maximbaz: @Morganamilo this is a new confirmation comparing to the latest release, why did you add it? I was just asked about clean builds, if I wanted to cancel I would have simply pressed <kbd>Ctrl+C</kbd>. I'm happy that all questions are now being asked in the beginning, but I feel this is a question that can be avoided.

/cc @Jguer

Morganamilo:
The printing was just done no get a feel for the flow and test the new system. it's all non final. Eventually I would like Y/N/A (yes/no/abort) on each question.

I plan to open an issue for the new install and another issue for how printing should go. I'm not home right now though. Feel free to open one for me. Would be nice to have a centralized place to discuss this.

maximbaz:
Got it, thanks! Y/N/A would be the best, I agree. I'll let you create the issues yourself when you are back, no rush at all, I'm just trying this new code out to give you some early feedback 🙂

PR #107, opened 2018-01-21
Fix numbermenu not working with new install alg

Morganamilo:
Simple fix, should have caught this before pushing #105 though.


Jguer:
Should have seen it as well. I'm now finally in one place for more than 4 days so I can properly review the development version now.

Issue #85, opened 2018-01-03
CLion AUR package doesn't work in AUR


132ikl:
CLion is being built when trying to install through yay. When I use yay -S clion I get this error:
https://pastebin.com/tzcJjuAn

I can build it manually or with pacaur. This has carried over through every update of CLion.

Morganamilo:
This is fixed in master and should be closed.

132ikl:
When will the releases section be updated? Should I just manually compile and use that until the updated version comes to AUR? Speaking of, would you consider creating an yay-git AUR package?

Jguer:
The current version needs more testing and work to be promoted to a stable release, specially with the changes being done to syntax. 
I'll consider creating a yay-git package since it is going to take a while

Morganamilo:
The new install algorithm is the last thing i really wanted to see. It still needs more testing and a little more work but I'm hoping for a release once that's done.

Also here's the pkgbuild I use for yay-git, i haven't bothered to put it on the AUR though.

```bash
# Maintainer: Jguer <joaogg3@gmail.com>
pkgname=yay-git
_pkgname=yay
pkgver=v2.219.r53.g61065dc
pkgrel=1
pkgdesc="Yet another yogurt. Pacman wrapper and AUR helper written in go."
arch=('i686' 'x86_64' 'armv7h' 'aarch64')
url="https://github.com/Jguer/yay"
license=('GPL')
options=('!strip' '!emptydirs')
depends=(
  'sudo'
)
makedepends=(
  'git'
  'go'
)
conflicts=('yay' 'yay-bin')
source=("git+https://github.com/Jguer/yay")
sha1sums=('SKIP')

prepare() {
  export GOPATH="${srcdir}/.go"
  export GOBIN="$GOPATH/bin"
  mkdir -p "$GOPATH"
  rm -rf "$GOPATH/src"
  ln -sf "$srcdir/$_pkgname/vendor" "$GOPATH/src"
}

build() {
  export GOPATH="${srcdir}/.go"
  export GOBIN="$GOPATH/bin"
  cd "$srcdir/$_pkgname"
  go build -v -o ${_pkgname} -ldflags "-s -w -X main.version=${pkgver}"
}

package() {
  _output="${srcdir}/$_pkgname"
  install -Dm755 "${_output}/${_pkgname}" "${pkgdir}/usr/bin/${_pkgname}"

  # Install GLP v3
[...]

PR #118, opened 2018-01-29
add yay-git to the packages list

heyitswither:


Morganamilo:
I would just like to note a couple of things about your PKGBUILD
- User defined variables in a PKGBUILD should always begin with a `_`. `VCS` -> `_VCS`
- Personally I would drop the VCS thing altogether and have `pkgname = yay-git` and `_pkgname = yay`
- A more complete pkgver can be generated using `git describe --long --tags | sed 's/^v//;s/\([^-]*-g\)/r\1/;s/-/./g' `. This would result in `2.219.r67.gb65dd86` instead of `r286.b65dd86`

I uploaded an example PKGBUILD a while ago if you wish to take a look. https://github.com/Jguer/yay/issues/85#issuecomment-359241373

heyitswither:
Thank you, I have fixed these issues

Jguer:
Ok, adding it as a conflict to the yay packages

PR #124, opened 2018-01-31
Filter out yay command line options

### LINKED ISSUE #122: Argumment parsing passing yay options to pacman
yay config options are being carried over to pacman on some operations when they should be parsed by yay.

Expected:
`yay -Syu --devel` 

Actual Result:
`pacman: unrecognized option '--devel'`

Workaround:
`yay -Y --devel`
`yay -Syu`

Morganamilo:
Fixes #122. funnily enough I already built a function to do this I must have just forgotten to call it.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -447,6 +447,10 @@ func (parser *arguments) parseShortOption(arg string, param string) (usedNext bo
 
 	arg = arg[1:]
 
+	if isYayParam(arg) {
+		return
+	}
+
 	for k, _char := range arg {
 		char := string(_char)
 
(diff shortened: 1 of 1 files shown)


Jguer:
Thanks for the quick fix, I figured you would know better where to touch it up.

Morganamilo:
Just realised this doesn't really fix it fully. I'll send another pr soon.

Jguer:
Weird, tested in the go folder with `./yay -Syu --devel` and it worked and now in the package it fails.

Morganamilo:
It filters it out a little too much. Basically it just ignores it and takes the config file's value.

Morganamilo:
Yes that should be a thing. Not sure how to do it for web related stuff though but I could do it for the parser and all that.

Jguer:
If you're able to do it for the parser that would be great

Morganamilo:
Sure I'll try. I've never actually done testing in go so I'll just be copying the existing test format.

PR #126, opened 2018-01-31
Fixes #122 properly

### LINKED ISSUE #122: Argumment parsing passing yay options to pacman

Morganamilo:
Options such as --devel are now striped from the parser before handling
the command but the option is still processed so that config.devel would
be true.

Also changed `changedConfig` to a global in config.go

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -199,7 +199,6 @@ func initAlpm() (err error) {
 func main() {
 	var status int
 	var err error
-	var changedConfig bool
 
 	err = cmdArgs.parseCommandLine()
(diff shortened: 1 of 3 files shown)


Morganamilo:
This seems to work as should now. GitHub is rate limiting me now so its getting a little annoying to test.

PR #130, opened 2018-01-31
Clone ABS/AUR Git repository instead of tarball

simon04:
Meant for discussion. Instead of downloading and unpacking the package source tarball, clone the corresponding Git repository. Inspired by https://github.com/falconindy/asp. Adds `git` as package dependency.

### DIFF
--- a/download.go
+++ b/download.go
@@ -2,66 +2,12 @@ package main
 
 import (
 	"fmt"
-	"io"
-	"net/http"
 	"os"
 	"os/exec"
-	"strings"
+	"path"
 
 	rpc "github.com/mikkeloscar/aur"
 )
-
-func downloadFile(path string, url string) (err error) {
-	// Create the file
-	out, err := os.Create(path)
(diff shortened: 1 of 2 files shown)


Morganamilo:
Instead of ripping out the old system I would rather this be a config option.

Jguer:
Same, would also prefer this to be an option, I know that tarballs are meant to disappear soon (if memory serves correctly) but I still prefer them.

Morganamilo:
Also how does this handle migrating. If i have a package that I installed using the tarball method then switch to the git method will it handle it correctly?

Morganamilo:
Do you plan on going anywhere with this? I might pick up on it myself otherwise. Add the config option and other fixes.

Jguer:
I've been playing around with git cloning and git pulling for updates in a few branches, I think we can roll it out as an option after the maintenance.

Jguer:
Closing this in favour of #297

Issue #109, opened 2018-01-22
Formatting and Printing 


Morganamilo:
This is a centralised issue for discussing formatting and printing text to the user. This issue was created with the new install algorithm in mind but is not limited to that.

If you are suggesting a change It can be useful to provide images or markup examples of what you wish to see.

## Planned Changes

- [x] Refactor escape codes from strings to dedicated functions
- [x] Add option to disable coloured output
- [x] Number menu for clean build and edit pkgbuild
- [x] Remove excess/unnecessary questions during install  
- ~~[ ] Add a (abort) option to y/n choices~~
    - Added in the number menu instead

## Regressions
- [x] Colour formatting broken when showing difference in versions on upgrade menu

Morganamilo:
Firstly I think wee need a refactor for most of the printing. I hate seeing magic escape codes in strings all over the code base and would much rather have these mapped to either const strings or functions.

For example the arrow prefix 
```
==> 
```
Could be a constant.

While printing it in red could be achieved via a function.
```go
fmt.Println(red(arrow + "Edit PKGBUILD?"))
```

Jguer:
I would also add escape codes as an on by default option but an option nonetheless since I know some users prefer to have no color coding or it makes their custom setup display black on black or other weird contrasts.

Jguer:
![2018-01-22-082200_3840x1080_scrot](https://user-images.githubusercontent.com/8071073/35228911-d6e535aa-ff89-11e7-9bd7-33c0993d0150.png)

![2018-01-22-082136_3840x1080_scrot](https://user-images.githubusercontent.com/8071073/35228917-d9cf170e-ff89-11e7-8205-cf2384775af2.png)

Just updated a machine that had seen no updates for 4 months. Good news, it worked. On other regards, if we're installing several packages maybe displaying a number menu of which to edit is better.

Morganamilo:
Yeah that's a really good idea actually. I was just trying to follow along with pacaur, never thought of a menu.

maximbaz:
- [ ] Upgrade doesn't highlight correctly the exact difference in package versions anymore:

![image](https://user-images.githubusercontent.com/1177900/35247099-67068f78-ffc9-11e7-9c2c-d2d01d5b5117.png)

[...]

PR #136, opened 2018-02-05
Prettify repository download messages

reesmichael1:
A simple commit to improve the looks of the message saying what will be downloaded and from which repositories.

Before:

image

After:

screenshot from 2018-02-05 14-10-25


Jguer:
Tried it at least with AUR package updates, looks great. Relevant to #109

PR #139, opened 2018-02-08
Re-add support for PacmanBin

kshlm:


Jguer:
Little oops, thanks for noticing

PR #142, opened 2018-02-09
Respect --noconfirm in config

### LINKED ISSUE #132: --noconfirm doesn't work anymore with -Sc
Since last update to yay v2.303 `yay -Sc --noconfirm` asks for confirmation:

```
yay -Sc --noconfirm
Packages to keep:
  All locally installed packages

:: Do you want to remove all other packages from cache? [Y/n] 
```

Morganamilo:
Fixes #132

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -664,8 +664,14 @@ func passToPacman(args *arguments) error {
 
 	argArr = append(argArr, "pacman")
 	argArr = append(argArr, cmdArgs.formatGlobals()...)
+	
+	if config.NoConfirm {
+		argArr = append(argArr, "--noconfirm")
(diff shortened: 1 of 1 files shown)


Morganamilo:
Did not realise you had already done this so now it's been fixed twice. This should be reverted.

Jguer:
So silly of me, I thought I had it fixed in an unpushed branch but it seems not

Morganamilo:
That's what I kinda did. I wrote the fix a while ago with the intent to fix a couple more things before making a pr. When I got reminded on it I just pushed it without pulling so I didn't see the new code. Should have pulled. :(

PR #144, opened 2018-02-10
Make devel operations respect default branch

### LINKED ISSUE #80: vsc does not respect default repo branch
I use `antigen-git` AUR package, and I noticed that yay never prompts that an update is available for it. This happens because in [zsh-users/antigen](https://github.com/zsh-users/antigen) the default repo branch is called `develop` and not `master`, while yay always track the master branch:

https://github.com/Jguer/yay/blob/3913673c8c8ae41804a7139ba43201083e200db4/vcs.go#L83-L87

One way to find the default branch is to use the following API call and look for the `default_branch`:

https://api.github.com/repos/zsh-users/antigen

However it is also possible to choose the branch by editing PKGBUILD and appending `#branch=develop` to the `source` parameter, this should also be respected and override the default branch value.

BrianAllred:
Fixes #80.

### DIFF
--- a/vcs.go
+++ b/vcs.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"time"
 )
 
 // branch contains the information of a repository branch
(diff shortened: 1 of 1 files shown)


Jguer:
Great work, seems to be working fine in my limited examples

PR #149, opened 2018-02-14
Fix not passing enough args to pacman

### LINKED ISSUE #145: -Sii not working
`-Sii` only prints `-Si` output:

```
$ yay -Sii yubikey-personalization
Repository      : community                                                                                                  
Name            : yubikey-personalization                                                                                    
Version         : 1.18.1-2                                                                                                   
Description     : Yubico YubiKey Personalization library and tool                                                            
Architecture    : x86_64                                                                                                     
URL             : https://github.com/Yubico/yubikey-personalization                                                          
Licenses        : BSD                                                                                                        
Groups          : None                                                                                                       
Provides        : ykpers                                                                                                     
Depends On      : libusb  json-c  yubico-c-client                                                                            
Optional Deps   : None                                                                                                       
Conflicts With  : ykpers                                                                                                     
Replaces        : None                                                                                                       
Download Size   : 60.43 KiB                                                                                                  
Installed Size  : 205.00 KiB                                                                                                 
Packager        : Christian Hesse <arch@eworm.de>                                                                            
Build Date      : Fri 09 Feb 2018 19:53:20 NZDT                                                                              
Validated By    : MD5 Sum  SHA-256 Sum  Signature
```

```
$ pacman -Sii yubikey-personalization
Repository      : community                                                                                                  
Name            : yubikey-personalization                                                                                    
Version         : 1.18.1-2                                                                                                   
Description     : Yubico YubiKey Personalization library and tool
Architecture    : x86_64
URL             : https://github.com/Yubico/yubikey-personalization
Licenses        : BSD
Groups          : None
Provides        : ykpers
Depends On      : libusb  json-c  yubico-c-client
Optional Deps   : None
Required By     : keepassxc  yubico-pam  yubikey-manager  yubikey-personalization-gui
Optional For    : None
Conflicts With  : ykpers
Replaces        : None
Download Size   : 60.43 KiB
Installed Size  : 205.00 KiB
Packager        : Christian Hesse <arch@eworm.de>
Build Date      : Fri 09 Feb 2018 19:53:20 NZDT
MD5 Sum         : 914ec4d55be5ccab6c52a59f5a8419ce
SHA-256 Sum     : 1d9966c2d0dc13c695dd0e2495864c8799221045f98d2e11bb3fd507e5d1e685
Signatures      : A6234074498E9CEEyubikey-personalization
```

BrianAllred:
Fixes #145.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -455,7 +455,6 @@ func handleYogurt() (err error) {
 
 func handleSync() (err error) {
 	targets := cmdArgs.formatTargets()
-	options := cmdArgs.formatArgs()
 
 	if cmdArgs.existsArg("y", "refresh") {
 		arguments := cmdArgs.copy()
@@ -480,7 +479,7 @@ func handleSync() (err error) {
 	} else if cmdArgs.existsArg("u", "sysupgrade") {
 		err = upgradePkgs(make([]string, 0))
 	} else if cmdArgs.existsArg("i", "info") {
-		err = syncInfo(targets, options)
+		err = syncInfo(targets)
 	} else if len(cmdArgs.targets) > 0 {
 		err = install(cmdArgs)
 	}
--- a/query.go
+++ b/query.go
@@ -152,7 +152,7 @@ func syncSearch(pkgS []string) (err error) {
 }
 
 // SyncInfo serves as a pacman -Si for repo packages and AUR packages.
-func syncInfo(pkgS []string, flags []string) (err error) {
+func syncInfo(pkgS []string) (err error) {
 	aurS, repoS, _, err := packageSlices(pkgS)
 	if err != nil {
 		return
@@ -160,10 +160,8 @@ func syncInfo(pkgS []string, flags []string) (err error) {
 
 	//repo always goes first
 	if len(repoS) != 0 {
-		arguments := makeArguments()
-		arguments.addArg("S", "i")
-		//arguments.addArg(flags...)
-		arguments.addTarget(repoS...)
+		arguments := cmdArgs.copy()
+		arguments.delTarget(aurS...)
 		err = passToPacman(arguments)
 
 		if err != nil {


Morganamilo:
The standard way to do this is to copy the command line args and strip out stuff you don't want in there. It's a bit clunky but the best we have right now feel free to think of a better one. But this approach will fail for long options such as `--info`.

Currently syncInfo takes a string slice called flags. This is leftover from the old system which relied on slices for most stuff. The current system mostly uses sets and parsers. I would recommend either changing the parameter to a parser or just use the global cmdArgs and remove the parameter.

BrianAllred:
How's this?

Morganamilo:
You shouldn't need to strip anything so you should just be able pass cmdArgs right into `passToPacman()` you can get rid of the arguments variable altogether. And if `existsDouble()` is true it already has it twice so you don't need to add it again.

BrianAllred:
The problem is if you pass an AUR package to pacman, it will throw an error. So we have to strip the AUR packages from the cmdArgs somehow.

Morganamilo:
Then make a copy of cmdArgs and replace the targets with just aur targets

BrianAllred:
Got it.

Morganamilo:
You can't just copy a struct like that. There's a copy function already made for parsers.

BrianAllred:
You can in Go (I tested it), but I can use the copy function, I don't mind
either way.

Morganamilo:
No you really can't. You only copy the reference to the maps so they end up being the same, only op is actually copied.

just use the prebuilt copy function and filter out the aur packages. Sure cmdArgs is not used further on so it works but you'll still end up modifying the map like it or not.

BrianAllred:
Ah, my mistake. Coming from C#, I was wary about passing by value vs. reference, so the [StackOverflow answer](https://stackoverflow.com/questions/21011023/copy-pointer-values-a-b-in-golang) I found when I Googled it said you're fine as long as there's no pointers. I didn't see any pointer values in the `arguments` data structure, so I'm assuming the maps use pointers behind the scenes, or the SO answer is wrong, one of the two.

I'll fix it later today.

Morganamilo:
Yeah it looks good now. Very simple and works perfectly.

Jguer:
Let's merge it then

Issue #135, opened 2018-02-04
Cached aur packages for newly bootstrapped filesystem.


pauldotknopf:
My system is currently setup to use [Darch](https://godarch.com/).

I have a Travis-CI repo setup that builds and deploys fresh operating system images for me to boot.

To cut down on build time on the CI server, I'd like to cache downloads and AUR packages.

On each build, I intend on mounting a cached ```/var/cache/pacman/pkg``` directory. This will save time on downloading of packages from official repositories.

But what about the AUR? I'd like to only clone/build/install if there is a new package version. Otherwise, I'd like to get install a cached package.

Any thoughts?

Morganamilo:
This was actually a thing I was working on before my laptop decided to brake. It will be added eventually although it don't really have the time to work on much right now.

Also the default build dir is due to be changed to `~/.cache/yay` Although with your use case you would probably want something more like `/var/cache/yay` maybe even with a dedicated yay user. Root can not use makepkg and regular users wont have write perms in /var, unless you mount it to /var/tmp of course.

PR #152, opened 2018-02-16
Improvements to install algorithm

### LINKED ISSUE #119: implement --force
For example, if I install software manually, and then I found out that there is a package for the same soft, I would install it with yay/Pacman. But it will report `x file already exists in filesystem` . I then would force the install with --force.

[arch wiki conflicting files](https://wiki.archlinux.org/index.php/Pacman#.22Failed_to_commit_transaction_.28conflicting_files.29.22_error)

### LINKED ISSUE #141: AUR builds fail with cleanafter enabled
Installing `aur/code` fails, because the sources seem to be deleted before the package finishes building.

Morganamilo:
This pr tweaks the install algorithm quite a bit but the main focuses are:

- Removing makepkg -s and -i (fixes #119)
- Package caching (implements #135)
- Fixing cleanafter (fixes #141)
- Better split package handling 

Now when a package is stored in cache it can be reinstalled via `yay -S` without doing a rebuild. A rebuilt will still occur if there is a newer version of the package available.

Cleanafter is now part of `install()` and runs in batch at the end instead of after every `passToMakepkg()`

Split packages are now displayed together for nicer output. Each pkgbase is built and installed only once and in one transaction. This should be enough to get a yes for split package support on the wiki but more testing is needed before changing that.

Here is an example of the new output:
 - libc++ and libc++abi are bing installed which belong to the libc++ pkgbase
 - python-virtualfish is being installed which belongs to the virtualfish pkgbase
 - discord is being installed

printscreen-2018-02-16_15 48 48

Old output:
image

For more details see individual commits.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -702,11 +702,6 @@ func passToMakepkg(dir string, args ...string) (err error) {
 	err = cmd.Run()
 	if err == nil {
 		_ = saveVCSInfo()
-		if config.CleanAfter {
-			fmt.Println(boldGreenFg(arrow +
-				" CleanAfter enabled. Deleting source folder."))
-			os.RemoveAll(dir)
-		}
 	}
 	return
 }
--- a/dependencies.go
+++ b/dependencies.go
@@ -16,9 +16,10 @@ type depTree struct {
 
 type depCatagories struct {
 	Repo     []*alpm.Package
-	RepoMake []*alpm.Package
-	Aur      []*rpc.Pkg
-	AurMake  []*rpc.Pkg
+	Aur[]*rpc.Pkg
+	MakeOnly stringSet
+	Bases map[string][]*rpc.Pkg
+
 }
 
 func makeDepTree() *depTree {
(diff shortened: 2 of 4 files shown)


Jguer:
Merged it with minimal test, we can use it as a work base

Morganamilo:
It installed everything I threw at it and the caching seems to work very well. I can't say testing was that extensive though.

Jguer:
Ran a gendb through it without any problem.
Devel packages update.
I'm going to fix some annoyances from gometalinter, give it let's say until saturday night, and maybe do a release to stable. 2.303 isn't that stable anyway

Jguer:
Ran a gendb through it without any problem.
Devel packages update.
I'm going to fix some annoyances from gometalinter, give it let's say until saturday night, and maybe do a release to stable. 2.303 isn't that stable anyway

Morganamilo:
If you wanted you could send me your exact settings and I can lint it myself before creating pr's.

Issue #146, opened 2018-02-11
Yay does not handle `Replaces` package field


beanaroo:
```
$ yay -Syu --devel
:: Synchronizing package databases...
 core is up to date
 extra is up to date
 community is up to date
:: Searching databases for updates...
:: Searching AUR for updates...
:: Checking development packages...

There is nothing to do
```

```
$ trizen -Syu --devel
:: Synchronizing package databases...
 core is up to date
 extra is up to date
 community is up to date
:: Starting full system upgrade...
:: Replace compositeproto with extra/xorgproto? [Y/n] y
:: Replace damageproto with extra/xorgproto? [Y/n] 
:: Replace fixesproto with extra/xorgproto? [Y/n] 
:: Replace fontsproto with extra/xorgproto? [Y/n] 
warning: glade: local (3.21.0-1) is newer than extra (3.20.2-1)
:: Replace inputproto with extra/xorgproto? [Y/n] 
:: Replace kbproto with extra/xorgproto? [Y/n] 
:: Replace randrproto with extra/xorgproto? [Y/n] 
:: Replace recordproto with extra/xorgproto? [Y/n] 
:: Replace renderproto with extra/xorgproto? [Y/n] 
:: Replace scrnsaverproto with extra/xorgproto? [Y/n] 
:: Replace videoproto with extra/xorgproto? [Y/n] 
:: Replace xextproto with extra/xorgproto? [Y/n] 
:: Replace xf86vidmodeproto with extra/xorgproto? [Y/n] 
:: Replace xineramaproto with extra/xorgproto? [Y/n] 
:: Replace xproto with extra/xorgproto? [Y/n] 
resolving dependencies...
looking for conflicting packages...

Packages (16) compositeproto-0.4.2-3 [removal]  damageproto-1.2.1-3 [removal]  fixesproto-5.0+9+g4292ec1-1 [removal]  fontsproto-2.1.3-2 [removal]  inputproto-2.3.2-1 [removal]
              kbproto-1.0.7-1 [removal]  randrproto-1.5.0-1 [removal]  recordproto-1.14.2-2 [removal]  renderproto-0.11.1-3 [removal]  scrnsaverproto-1.2.2-2 [removal]
              videoproto-2.3.3-1 [removal]  xextproto-7.3.0-1 [removal]  xf86vidmodeproto-2.3.1-3 [removal]  xineramaproto-1.2.1-3 [removal]  xorgproto-2018.1-1
              xproto-7.0.31-1 [removal]

[...]

PR #154, opened 2018-02-16
Improve conflict handling

Morganamilo:
Renable conflict checking, was disabled for testing, forgot to reenable
Use pacman --ask to remove conflicting packages, dont do it directly.
Say what each package conflicts with
Check for conflicts before aking for clean build

Reverse conflict are still not checked

Example output:
image

Relies on Jguer/go-alpm#4

### DIFF
--- a/install.go
+++ b/install.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"strconv"
 
 	alpm "github.com/jguer/go-alpm"
 	rpc "github.com/mikkeloscar/aur"
@@ -65,6 +66,8 @@ func install(parser *arguments) error {
 			}
 		}
 
+
+
 		//printDownloadsFromRepo("Repo", dc.Repo)
 		//printDownloadsFromRepo("Repo Make", dc.RepoMake)
 		//printDownloadsFromAur("AUR", dc.Aur)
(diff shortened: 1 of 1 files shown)


Morganamilo:
@AladW You closed the talk section so I'll just link you here. This commit removed `-Rdd`, less that 10 days after it was added in the first place.

> I pointed this out and was ignored

 I don't remember you pointing it out anywhere here. May you're confused with trizen? https://github.com/trizen/trizen/commit/ba687bc3c3e306e6f3942e95f825ed6a55d3ad69

AladW:
I mentioned in the comments of the commit adding -Rdd

Morganamilo:
This commit? https://github.com/Jguer/yay/commit/28515c8b

AladW:
right, https://github.com/Jguer/yay/commit/28515c8beb5da851f8c37bf7aa70e210e0e2163d?diff=unified#r27587686

Morganamilo:
Ah right, was in the middle of the page so I missed it some how. Sorry about that. Would like to point out that It was removed the day after that comment.

AladW:
Changed the entry to "partial" then for now. If #130 is implemented yay should move back in the top section of the table

Morganamilo:
#146 Is fixed in the git only, that's why I had not changed the wiki myself. We're due a release in a day so I don't know if it's worth the effort of changing it back.

AladW:
Eh, don't worry about it. Thanks for making those improvements.

AladW:
Oh, I see what you mean now. The wiki table used to have entries by name only, not actual AUR packages. So it should probably reflect how the release version performs as you say.

As I agree it's not worth changing back the colors, I've just added a "yay-git" to the entry.

Morganamilo:
Yeah that seems like the best way to do it. As long as the table doesn't end up getting flooded with helper-git* exceptions dotted around.

PR #167, opened 2018-02-20
Remove unused code, prune dependencies, appease vet and gofmt

mvdan:


Morganamilo:
I would avoid touching anything in `vendor/` as it is automatically pulled from upstream dependencies.

mvdan:
I did not touch anything in the vendor directory. I simply re-ran `dep ensure` with the updated Gopkg.toml, which enables pruning.

Morganamilo:
Right yeah thought you ran `go fmt` on all files but only our files are changed in that commit. Looks fine then.

Morganamilo:
We're finally moving again. Resolve the conflicts and we should be able to merge this right away.

mvdan:
Took longer to rebase with all the conflicts than it did to write these patches to begin with; please do merge before there are another four dozen commits in master.

Morganamilo:
Yeah sorry about that. My plan was vendor update then merge every one else's stuff then resolve the conflicts on my end. But the vendor update caused conflicts so I couldn't merge this when I wanted to. All done now at least.

PR #168, opened 2018-02-20
add config.String() to pretty print default config in valid json

### LINKED ISSUE #72: Print config not human friendly
The flag `--printconfig` just prints the JSON in a single line with no trailing newline. It would be more helpful to pretty print it.

seankhliao:
implements fmt Stringer interface
fixes #72

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -273,11 +273,10 @@ func sudoLoopBackground() {
 
 func sudoLoop() {
 	for {
-		updateSudo()	
+		updateSudo()
 		time.Sleep(298 * time.Second)
 	}
 }
-
 
 func updateSudo() {
 	for {
@@ -398,7 +397,7 @@ func handleVersion() {
 func handlePrint() (err error) {
 	switch {
 	case cmdArgs.existsArg("d", "defaultconfig"):
-		fmt.Printf("%#v", config)
+		fmt.Printf("%v", config)
 	case cmdArgs.existsArg("n", "numberupgrades"):
 		err = printNumberOfUpdates()
 	case cmdArgs.existsArg("u", "upgrades"):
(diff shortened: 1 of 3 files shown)


Morganamilo:
Looks good although there's some weird stuff going on with the manpage and `--help`.
```
-d --defaultconfig   Print current yay configuration
```
would be nice to get this changed to print the default config instead.

Maybe have the current config be
```
-g --config
```
`-c` is taken, `-jc` was suggested but after thinking about it a two letter short switch feels weird.

Would be nice if you could implement and fix this. Your pr looks fine as is though if you'd rather not.

Should probably have @Jguer's opinion on `-g` vs `-js` first though.

seankhliao:
I've separated current and default config, with `-g --config` for now but I haven't updated the man pages or the completions,

Morganamilo:
completions are kinda behind right now I wouldn't worry about it too much. Apart from the manpage it looks perfect.

seankhliao:
should I also add in the `-c --complete` to the manpage since I noticed it is missing?

Jguer:
You can add -c to the man page but it's a very implementation specific option, not really useful unless you want to parse package names for display.

Right now I'm only focusing on completions for fish since it's the shell I use and the mvp for other shells for me is that it auto completes package names

seankhliao:
I updated the manpage, without -c

I'll have to leave completions to you since I'm unfamiliar with how to do that

Morganamilo:
Code, `--help` and manpage all look good. We could merge now but the meaning of `-d` is changed so maybe you want to wait until we are ready to bump to Yay v3.xxx.

seankhliao:
no problem, was just browsing through the issues looking easy fixes :smile:

PR #169, opened 2018-02-20
upgrade: be consistent about no colors

mvdan:
If pacman's color is off, don't use colors to mark the repos of
upgradeable packages.

### DIFF
--- a/upgrade.go
+++ b/upgrade.go
@@ -65,6 +65,9 @@ func (u upSlice) Print(start int) {
 		var left, right string
 
 		f := func(name string) (output string) {
+			if alpmConf.Options&alpm.ConfColor == 0 {
+				return name
+			}
 			var hash = 5381
 			for i := 0; i < len(name); i++ {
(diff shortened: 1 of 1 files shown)


upgrade.go
```diff
@@ -65,6 +65,9 @@ func (u upSlice) Print(start int) {
 		var left, right string
 
 		f := func(name string) (output string) {
```
Morganamilo: This should probably be moved to its own function, all other checks to `if alpmConf.Options&alpm.ConfColor == 0` have their own function.

Morganamilo: To clarify because it's already its own function technically. I meant move it to it's no long inside Print and give it a decent name.

mvdan: Sure, I agree. But this is a bug fix, so I wanted to keep the patch small. There's no reason to do a refactor in the same commit.

Morganamilo: You have a point, it was already like this all you did was add the missing check. Change is small and good so I'll add it to the next realise. We can refactor during the feature freeze.

Issue #163, opened 2018-02-19
Wrong option parsing


notramo:
- Yay ignores nonexisting option (e.g. `yay -Sű`).
- Conflicting (or ambigous) actions are ignored, and performed wrong.
  - `yay -Sy crystal --info exa` synchronizes the database, and prints info about `crystal` and `exa` packages, however this is ambigous, because the user may want to update the database, install `crystal`, and view info about `exa`. In this case, one operation should be permitted only.
  - `yay -Syui` ignores the `-i` option.

Morganamilo:
Flags are not ordered `yay -Sy crystal --info exa` is not ambiguous. You asked for `--info` so it give you info. It doesn't matter if `crystal` comes before it.

Jguer:
Seems like a non issue, closing until it generates real problems

PR #173, opened 2018-02-21
cmd: error on unknown command

mvdan:
Don't silently ignore it.

Before:

	$ yay --foo

After:

	$ yay --foo
	unknown command: -Y --foo

Updates #163.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -438,6 +438,9 @@ func handleYay() (err error) {
 		err = handleGetpkgbuild()
 	} else if len(cmdArgs.targets) > 0 {
 		err = handleYogurt()
+	} else {
+		args := cmdArgs.formatArgs()
+		err = fmt.Errorf("unknown command: %s", strings.Join(args, " "))
 	}
 
 	return


Morganamilo:
This wouldn't fully fix the problem would it not? Mixing invalid and valid commands would still work `yay -c --foo` And this only applies to `handleYay()` so `-S -U -G....` wont be affected by this.

mvdan:
I agree - the code is buggy in multiple places. I simply fixed the biggest offender. Hence "Updates" and not "Fixes" :)

Morganamilo:
This is mostly by design. Keeping track of every single Pacman option seems like a hassle and and over complication. Especially when each operation support a different set off commands.

If the user does enter an invalid option it's already the user's fault. It might be ideal to catch the errors but I don't think it's worth the complication or the effort.

mvdan:
If that's the design, then the fallback should be to pass the command/options to pacman and let it give the proper error. One way or another, ignoring arguments is not a good outcome.

Morganamilo:
That's what happens already, if Pacman errors we don't continue
```
morganamilo@Octavia ~ % yay -Syx xterm discord
error: invalid option '-x'
exit status 1
````
I still don't see how this really fixes anything. If the user enters more than one target it will never error so `yay --foo xterm` wont be caught. The only two cases that it will error on is when the user doesn't put `-c`, '-h', '-g' or '--gendb'. And  `-g` should be taken out in favour of `-G` really so it pretty much leaves the three case.

I'm not against trying to fix this more fully but it will probably require a little more thought and some code reworks. This method is not expandable so I don't see a use in merging it now.

I would much rather you think it through and come up with a better approach that can better be applied to the code as a whole.

mvdan:
Fair enough. I don't have the time to refactor/redesign all of this code though, so this pull request should be closed if it's not going to be merged.

Morganamilo:
Closing it then. Even if you don't have time to code it yourself you could open an RFC with a strategy you think could work.

Your other PR is fine though, As soon as @Jguer gets around to making a release I'll merge it for the one after.

PR #179, opened 2018-02-26
Ask if yay should try to import missing PGP keys

qrwteyrutiyoup:
When building a package from the AUR for which there is a missing key,
yay will now prompt the user whether it should try to import the missing
key using gpg:

[...]
==> Verifying source file signatures with gpg...
    ncurses-6.1.tar.gz ... FAILED (unknown public key 702353E0F7E48EDB)
==> ERROR: One or more PGP signatures could not be verified!
==> Unknown key "702353E0F7E48EDB". Import?  [Y/n]
==> Importing key "702353E0F7E48EDB" with gpg...
gpg: key 702353E0F7E48EDB: 142 signatures not checked due to missing keys
gpg: key 702353E0F7E48EDB: public key "Thomas Dickey <dickey@invisible-island.net>" imported
[...]

Default option is to try to import the key ([Y/n]).

### DIFF
--- a/install.go
+++ b/install.go
@@ -200,12 +200,17 @@ func install(parser *arguments) error {
 			return nil
 		}
 
-		err = downloadPkgBuildsSources(dc.Aur, dc.Bases)
+		err = parsesrcinfosGenerate(dc.Aur, srcinfos, dc.Bases)
 		if err != nil {
 			return err
 		}
 
-		err = parsesrcinfosGenerate(dc.Aur, srcinfos, dc.Bases)
+		err = checkPgpKeys(dc.Aur, srcinfos, dc.Bases, nil)
+		if err != nil {
+			return err
+		}
+
+		err = downloadPkgBuildsSources(dc.Aur, dc.Bases)
 		if err != nil {
 			return err
 		}
--- a/keys.go
+++ b/keys.go
@@ -1,0 +1,121 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	rpc "github.com/mikkeloscar/aur"
+	gopkg "github.com/mikkeloscar/gopkgbuild"
+)
+
+// pgpKeySet maps a PGP key with a list of PKGBUILDs that require it.
+// This is similar to stringSet, used throughout the code.
+type pgpKeySet map[string][]*rpc.Pkg
+
+func (set pgpKeySet) toSlice() []string {
+	slice := make([]string, 0, len(set))
+	for v := range set {
+		slice = append(slice, v)
+	}
+	return slice
+}
+
+func (set pgpKeySet) set(key string, p *rpc.Pkg) {
+	// Using ToUpper to make sure keys with a different case will be
+	// considered the same.
+	upperKey := strings.ToUpper(key)
+	if _, exists := set[upperKey]; !exists {
+		set[upperKey] = []*rpc.Pkg{}
+	}
+	set[key] = append(set[key], p)
+}
+
+func (set pgpKeySet) get(key string) bool {
+	upperKey := strings.ToUpper(key)
+	_, exists := set[upperKey]
+	return exists
+}
+
+// checkPgpKeys iterates through the keys listed in the PKGBUILDs and if needed,
+// asks the user whether yay should try to import them. gpgExtraArgs are extra
+// parameters to pass to gpg, in order to facilitate testing, such as using a
+// different keyring. It can be nil.
+func checkPgpKeys(pkgs []*rpc.Pkg, srcinfos map[string]*gopkg.PKGBUILD, bases map[string][]*rpc.Pkg, gpgExtraArgs []string) error {
+	// Let's check the keys individually, and then we can offer to import
+	// the problematic ones.
+	problematic := make(pgpKeySet)
+	args := append(gpgExtraArgs, "--list-keys")
+
+	// Mapping all the keys.
+	for _, pkg := range pkgs {
+		for _, key := range srcinfos[pkg.PackageBase].Validpgpkeys {
+			// If key already marked as problematic, indicate the current
+			// PKGBUILD requires it.
+			if problematic.get(key) {
+				problematic.set(key, pkg)
+				continue
+			}
+
+			cmd := exec.Command(config.GpgBin, append(args, key)...)
+			err := cmd.Run()
+			if err != nil {
+				problematic.set(key, pkg)
+			}
+		}
+	}
+
+	// No key issues!
+	if len(problematic) == 0 {
+		return nil
+	}
+
+	question, err := formatKeysToImport(problematic, bases)
+	if err != nil {
+		return err
+	}
+	if continueTask(question, "nN") {
+		return importKeys(gpgExtraArgs, problematic.toSlice())
+	}
+
+	return nil
+}
+
+// importKeys tries to import the list of keys specified in its argument. As
+// in checkGpgKeys, gpgExtraArgs are extra parameters to pass to gpg.
+func importKeys(gpgExtraArgs, keys []string) error {
+	args := append(gpgExtraArgs, "--recv-keys")
+	cmd := exec.Command(config.GpgBin, append(args, keys...)...)
+	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
+
+	fmt.Printf("%s Importing keys with gpg...\n", bold(cyan("::")))
+	err := cmd.Run()
+
+	if err != nil {
+		return fmt.Errorf("%s Problem importing keys", bold(red(arrow+" Error:")))
+	}
+	return nil
+}
+
+// formatKeysToImport receives a set of keys and returns a string containing the
+// question asking the user wants to import the problematic keys.
+func formatKeysToImport(keys pgpKeySet, bases map[string][]*rpc.Pkg) (string, error) {
+	if len(keys) == 0 {
+		return "", fmt.Errorf("%s No keys to import", bold(red(arrow+" Error:")))
+	}
+
+	var buffer bytes.Buffer
+	buffer.WriteString(bold(green(("GPG keys need importing:\n"))))
+	for key, pkgs := range keys {
+		pkglist := ""
+		for _, pkg := range pkgs {
+			pkglist += formatPkgbase(pkg, bases) + " "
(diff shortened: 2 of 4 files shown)


Morganamilo:
This is the totally wrong way to go about this. We already get the PGP keys when we parse the srcinfos, there's no need to parse the makepkg output.

qrwteyrutiyoup:
Thanks for the feedback. I will resubmit using the approach you described.

Morganamilo:
That looks pretty good although I think you could tweak it a bit to follow the "ask questions first then do" mentality.

Its probably common that the user will already have all the pgp keys in their keyring already. So you could do `gpg --list-key <key1> <key2> <key3...>`. That way you can check all the keys in a single gpg call and if it comes back okay just skip over the key checking. If not then you'll have to check one by one like you are now. This might be premature optimisation though we would probably need a lot of keys in one transaction for this to make a difference.

Instead of the "ask, add, ask, add..." method it would be better to do all the asking then import all the keys in a single gpg command.

The easiest way to do this is have a function to get all the pgp keys from the pkgbuilds. Probably return a string set. They can't contain duplicates which is nice. Then have an `askaddkeys()` type method, ask all the questions in a row then import at the end.

In fact the build is going to fail if they decline to import a key so we could just list all the keys and provide one y/n question at the end. Or we could do a number menu but I wouldn't worry about that for now.

qrwteyrutiyoup:
Yeah, sounds better to check them all at once, and if it fails, list them all and provide one y/n question at the end, as you suggest. I will try to submit another iteration tonight or tomorrow.

Morganamilo:
Have you made or plan to make any progress on this? It's quite a good feature and I can add this myself if you no longer plan to.

qrwteyrutiyoup:
Sorry, I kinda forgot about this, but I made progress, yeah. Give me a moment and I am going to upload a revision shortly.

Morganamilo:
Don't worry about it. It's not like you're holding up development or anything, just wanted to check up.

install.go
```diff
@@ -542,3 +547,65 @@ func completeFileName(dir, name string) (string, error) {
...
+	return nil
+}
+
+// importKeys tries to import the list of keys specified in its argument.
+func importKeys(keys []string) bool {
+	args := []string{"--recv-keys", strings.Join(keys[:], " ")}
+	cmd := exec.Command(config.GpgBin, args...)
+	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
+
+	fmt.Printf("%s Importing key(s) with gpg...\n", arrow)
```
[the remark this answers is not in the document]
Morganamilo: This line is missing colour.

install.go
```diff
@@ -542,3 +547,65 @@ func completeFileName(dir, name string) (string, error) {
...
+	err := cmd.Run()
+
+	if err != nil {
+		fmt.Printf("%s Problem importing key(s)\n", arrow)
```
Morganamilo: This line is missing colour.

install.go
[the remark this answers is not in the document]
Morganamilo: Doing actions should start with ":: " not an arrow.

install.go
```diff
@@ -542,3 +547,65 @@ func completeFileName(dir, name string) (string, error) {
...
+			if err != nil {
+				problematic = append(problematic, key)
+				keys[key] = true
+			}
+		}
+		keyStr := "key"
+		if len(problematic) > 1 {
+			keyStr = "keys"
+		}
+		question := fmt.Sprintf("Problem detected with the following %s: %q. Import?", keyStr, strings.Join(problematic[:], ","))
```
Morganamilo: This is proving to be way too long of a line this should probably be done like:

```
GPG keys need importing:
    key1
    key2
    key3
Import? [Y/n]
```
Additionally it might be nice to say what pkgbuild wants to import what key although I am unsure how to format that currently.

This is a formatting issue though, you shouldn't need to worry about it until implementation is finished.

install.go
```diff
@@ -542,3 +547,65 @@ func completeFileName(dir, name string) (string, error) {
 
 	return "", nil
 }
+
+// checkGpgKeys iterates through the keys listed in the PKGBUILDs and if needed,
+// asks the user whether yay should try to import them.
+func checkGpgKeys(srcinfos map[string]*gopkg.PKGBUILD) error {
```
Morganamilo: Gpg is the program used, they are still PGP keys.

install.go
```diff
@@ -542,3 +547,65 @@ func completeFileName(dir, name string) (string, error) {
...
+	cmd := exec.Command(config.GpgBin, args...)
+	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
+
+	fmt.Printf("%s Importing key(s) with gpg...\n", arrow)
+	err := cmd.Run()
+
+	if err != nil {
+		fmt.Printf("%s Problem importing key(s)\n", arrow)
+	}
+	return err == nil
```
Morganamilo: Why not just return the error?

install.go
```diff
@@ -542,3 +547,65 @@ func completeFileName(dir, name string) (string, error) {
...
+	keys := map[string]bool{}
+	// args will be usde to check all the keys at once.
+	args := []string{"--list-keys", "--with-sig-check"}
+
+	for _, pkgbuild := range srcinfos {
+		for _, key := range pkgbuild.Validpgpkeys {
+			// Using ToUpper to make sure case won't be an issue.
+			keys[strings.ToUpper(key)] = false
+			// For this check, it is OK if there are repeated keys.
+			args = append(args, key)
```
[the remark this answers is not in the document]
Morganamilo: I forgot .toSlice() is only a stringSet thing due to the lack of generics. Point stands even if you have to do it manually.

Morganamilo:
>checkGpgKeys() seems to fail to say keys are missing if I try to install two packages witch use PGP keys but already have the keys from one of those packages.

Turns out `gpg --list-keys key1 key2 key3` will return 0 if any of the keys exist. My bad for assuming it would error if one of them didn't exist.

qrwteyrutiyoup:
Hello again, thanks for the feedback once more! I have incorporated most of the suggestions (if not all) in this new revision I am uploading shortly. As for `gpg --list-keys key1 key2 key3`, no worries, it was worth a try :) 

I ended up creating a file keys.go and associated tests in keys_test.go to make it easier to verify the expected behavior.

Morganamilo:
Looking good, I wont bother with a full review this time, my main complaint is formatting.

Mainly this bit:
image

and This:
```go
type pgpKeySet map[string][]string
```
Which you should change to:
```go
type pgpKeySet map[string][]*gopkg.PKGBUILD
```
instead of holding a string for the pkgbase just hold a pointer to the pkgbuild.

Take a look at `formatPkgBase()` that should be used to print package names, which takes a *PKGBUILD hence the above point.

Also I would knock the lines back an indent so that the key is indented once and the question is not indented. The packages should be a different colour than the key for contrast and use "\t" rather than spaces.

keys.go
```diff
@@ -0,0 +1,109 @@
...
+
+	// Let's check the keys individually, and then we can offer to import
+	// the problematic ones.
+	problematic := make(pgpKeySet)
+	args := append(gpgExtraArgs, "--list-keys")
+	for key := range keys {
+		cmd := exec.Command(config.GpgBin, append(args, key)...)
+		err := cmd.Run()
+		if err != nil {
+			problematic[key] = keys[key]
```
Morganamilo: Just use one pgpKeySet and `delete(keys, key)` if it's not problematic.

Morganamilo: Or actually just do the opposite and only add problematic keys in the first place.

qrwteyrutiyoup:
I changed the set to map strings to `[]*gopkg.PKGBUILD`, however I had to write something similar to `formatPkgBase`, to work with these `[]*gopkg.PKGBUILD` (maybe I should move it to `print.go`?) and added corresponding tests.

As for the colors, I picked yellow for the keys and green for the packages that require it; not sure these are good ones, so feel free to suggest something else.

Morganamilo:
Try to avoid yellow, it doesn't really work on light terminals.

Your `formatPkgbaseFromPkgbuilds()` dose not work properly. `formatPkgbase()` is designed to only list packages that are going to be installed not every package from the base. Yours lists every package in the base.

I would make `checkPgpKets()` like the rest of the functions around and be called like `(dc.Aur, srcinfos, dc.Bases, nill)`
which would allow you to use `formatPkgbase()`

for every package in dc.Aur, dc.Bases[pkgbase] should exist. So instead of looping through srcinfos, loop through the packages, get the srcinfo from  srcinfos[pkg.PackageBase] (which is safe to assume exists). That way you can have:
```go
type pgpKeySet map[string][]*rpc.Pkg
```
which is what I was thinking the first time around but confused it with gopkg.PKGBUILD.

That should then let you use the proper `formatPkgbase()`.

Little convoluted but it's all I've got.

Morganamilo:
It looks good now, I'll merge it after the next release which should be tomorrow.

Future todos for this would be:
- add `--gpg` flag to go along with `--git`, `--tar` `--makepkg` ect
- add `--gpgflags` similar to `--mflags`
- update docs for the new flags (usage and manpage)
- update shell completion for new flags

That doesn't have to be done this PR  although you're free to. I'm happy to do it myself if you want.

So happy to finally get this in master, it's a really nice feature thanks for your work.

Jguer:
I've been silent on this pull request as @Morganamilo has been really on top of it and our views on `yay` are very similar and I didn't want to create dissonance but I'm very excited for this pull and you've done great work @qrwteyrutiyoup and it needs to be recognized.

qrwteyrutiyoup:
@Morganamilo: thanks for the reviews and guidance. As for these items you listed as TODO, I can take a stab at them in a separate PR.

And thanks, @Jguer. It was nice getting this PR accepted, and I was able to get to know `yay` internals a little bit. Hopefully, I will be able to contribute more in the future. Keep up the great work, you people :)

Morganamilo:
@Jguer It looks like you forgot the version bump in this release, Are you gonna release a .506 quickly? Checking before I merge everything. Not that it matters too much, you can tag a commit from the past anyway.

Jguer:
The makefile is only used for yay-bin. `yay` uses normal `go build` so no rerelease is needed. I made the new `yay-bin` with the updated values

Morganamilo:
Alrighty then just wanted to make sure.

Issue #188, opened 2018-02-27
[RFC] Proposed Issue template


Jguer:
Added `\` so comments are displayed

#### Affected Version
\
\

#### Affected Command
\
\
\

#### Expected Results
\

#### Actual Results
\

Morganamilo:
- You should probably mention to try the -git version first before submitting an issue.
- Might want to put `yay -V` in backticks.
- Personally I always find the `Expected Results` `Actual Results` to be a bit annoying and preachy but that's probably just me. I see it used elsewhere a bunch so it can't be that bad.

Issue #185, opened 2018-02-27
Can't install package groups


notramo:
For example `yay -S mate` says that the package is missing from AUR. `sudo pacman -S mate` starts installing the MATE desktop.

Morganamilo:
@Jguer This used to work right? If so I probably broke it and can bisect.

Jguer:
This used to work. Probably broken by the new install patches

PR #189, opened 2018-02-27
Create issue template

Jguer:
#188

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,21 @@
+#### Affected Version
+
+<!-- Use yay -V to get installed version -->
+
+<!-- Example: `yay v3.373` -->
+
+#### Affected Command
+
+<!-- if applies -->
+
+<!-- Example: yay -Syu -->
+
+<!-- Example: yay -S yay-bin -->
+
+#### Expected Results
+
+<!-- Example: Package installed successfully -->
+
+#### Actual Results
+
+<!-- Example: Error is thrown -->
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,21 +1,19 @@
+<!-- The following template is just a suggestion -->
+
+<!-- Feel free to edit or discard it if you prefer a different format -->
+
 #### Affected Version
 
-<!-- Use yay -V to get installed version -->
+<!-- Please ensure you are using the latest yay-git package -->
+
+<!-- Use `yay -V` to get installed version -->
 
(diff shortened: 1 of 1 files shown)


Morganamilo:
I think what I said in #188 Is most of what I have to say on this.

I don't know how I feel about templates to be honest. They do force people to write better issues who otherwise wouldn't. But they get in the way of people who write good issues in the first place.

I tweaked the layout a little, too many comments maybe?

```markup
<!-- The following template is just a suggestion -->
<!-- Feel free to edit or discard it if you prefer a different format -->

#### Affected Version
<!-- Please ensure you are using the latest yay-git package -->
<!-- Use `yay -V` to get installed version -->
<!-- Example: `yay v3.373` -->

#### Expected Results
<!-- Example: Package installed successfully -->

#### Actual Results
<!-- Example: Error is thrown -->
<!-- Pasting terminal text is encouraged -->
<!-- If you think you know the cause feel free to explain>

###Steps to reproduce
<!-- Exact commands are useful -->
1.
2.
3.
```

Jguer:
I like the just as a suggestion but it also reminds people of needed information, I think removing the Expected results and Actual results would be a better idea. What I really want is to know if they've tried the git version first. 

For example #185 I used it was a problem in master because it had been touched around but I couldn't be certain by the issue.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,19 @@
...
+
+<!-- Example: `yay v3.373` -->
+
+#### Issue
+
+### Steps to reproduce
+
+<!-- Exact commands are useful -->
+
+1. 2. 3.
```
Morganamilo: Markup really doesn't like this, doing it on separate lines would be better.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,19 @@
...
+
+<!-- Please ensure you are using the latest yay-git package -->
+
+<!-- Use `yay -V` to get installed version -->
+
+<!-- Example: `yay v3.373` -->
+
+#### Issue
+
+### Steps to reproduce
```
Morganamilo: This is a different header size than the others.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,19 @@
+<!-- The following template is just a suggestion -->
+
+<!-- Feel free to edit or discard it if you prefer a different format -->
```
Morganamilo: Now that the template is not so in your face maybe we don't need this anymore?

Jguer:
For some reason my markdown got reformatted

PR #192, opened 2018-02-28
Correctly mark packages as dependencies

Morganamilo:
Previously Installing repo targets and repo dependencies of aur targets
was done in two steps.

doing `yay -S repo1 repo2 aur1 aur2` would lead to yay calling
`pacman -S repo1 repo2`. Then after we find all the dependencies of the
aur packages we call `pacman -S --asdeps <all repo dependencies>`. This
was an easy way to correctly mark dependencies. Since 005635b4 Both
these calls were merged into one command but dependency marking was
forgoten about.

Now correctly mark the dependencies through `pacman -D`

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


Morganamilo:
Although this works quite well, it can lead to some Pacman spam in the terminal:
```
libxdmcp: install reason has been set to 'installed as dependency'
polkit: install reason has been set to 'installed as dependency'
curl: install reason has been set to 'installed as dependency'
filesystem: install reason has been set to 'installed as dependency'
and so on for every dep pulled in.................
```

Even though it's done through one Pacman call, Pacman still wants to print each change on it's own line. I see a couple of things that could be done.
- Leave as is - Its kind of annoying but not really anything bad
- Hide the output - I think this is fine this info was never printed before so its not like we're removing anything
- Set the install reason via alpm - Probably annoying to implement but doable, I trust Pacman more though

Got any preference @Jguer? The middle option is probably the best one.

Jguer:
I would hide the output, it's not as if it's the first time yay hides pacman output. I would just be careful to watch out for non-zero output on that pacman command

Morganamilo:
it already does exit on error although there's no error message so that could be confusing. Could give an error message like "Failed to mark packages as dependencies: aborting" or maybe capture the stderr from Pacman. I would have to see what Pacman actually prints to stderr though see if its useful.

Morganamilo:
That should do it. If you think it looks good merge it.

Issue #181, opened 2018-02-27
Installing python-acme-git does not update VCS entry


Morganamilo:
the python2-acme repo has a lot of branches, 272 currently. The GitHub API will only list a certain amount of items per page. Because of this the master branch does not show up on the first page so Yay never finds the master branch to get the commit from.

You could traverse the pages with `?page=`  and increase the page size up to 100 with `?per_page` But I think a better solution is moving to a git `ls-remote` method described in #166.

Jguer:
That is a super edge case but I still think #166 is a win even if it didn't fix #181

Morganamilo:
I ended up brainstorming a couple of ideas in #166 which one do you mean? the `git ls-remote` is my current choice because it does not rely on the GitHub API which would fix #79 and allow us to expand --devel to any git repo even if it's not on GitHub.

Currently I'm working on #88 which I believe is working fine but I got stuck for ages as to why it wasn't working and it ended up being this #181. I am only working on #88 currently though. Not going to do anything with #166 until we have something more planned out.

Jguer:
Now that I see `git ls-remote https://github.com/Jguer/yay HEAD` I like it even more, because it's git general instead of github specific

Issue #166, opened 2018-02-20
[RFC] Use time based method to check devel packages


Morganamilo:
Instead of storing the latest commit we would instead use the build time of the package and check it against the time of the latest commit. The build time is stored inside the pacman package so no need for a yay_vcs file.

I could see this failing if a new commit is pushed during build.
1. Start build
2. New commit
3. Build finishes, build date is marked as now but a new commit came out before build date

Although if that happens you're either really unlucky or the repo is being updated so often a new commit will come in soon and trigger an update.

So it's not full proof but it does remove the need for a cache file and --gendb and would be an easy fix for #88 although it shouldn't be too hard to fix without migrating away from this feature.

We could also go the pacaur route and unconditional update every package ended in -git|-hg|-nightly and so on. If we keep the other devel method it might be worth adding this too as `--forcedevel` or something.

Jguer:
I like that idea, hadn't thought about it but it's very intelligent. 
I don't like the idea of unconditional update of every package, but the time based seems like a good option and the issue present is a minor thing.

Morganamilo:
Hopefully we can skip the github api and get the time of the latest commit directly via git which would open this up to supporting more pkgs. Still non git pkgs would still never update.

With the `--forcedevel` it wouldn't be too bad seems as it would only target pkgs with specific suffixes. Better still we can choose not to update the package after we run `pkgver()` and see if it's newer than the installed variant.

Morganamilo:
Looking around if we kept the hash method we can drop the github api and use git directly.

```
git ls-remote https://github.com/Jguer/yay HEAD
```
Will get the latest commit from the default branch.

```
git ls-remote https://github.com/Jguer/yay <branch>
```
Will do the same for whatever branch you choose.

Sadly I don't think it's possible to get the time of the hash.

hcorion:
:+1: I would appreciate a forcedevel option, as some of my -git packages still don't update when they do have updates (rpcs3-git, dolphin-emu-git)

Morganamilo:
I'm still thinking about how exactly we could do it. Pacaur just matched packages ending in `git/hg/nightly` and so on. A more full-proof way is to download all installed AUR packages, run the `pkgver()` function to check if there's an update but that would take a while.

PS if you're running the latest `yay-git`  then ``yay -S `yay -Qqs -- -git` `` Should update all `-git` packages.

notramo:
What about using the `--timedevel` option for this? It is more descriptive than `--forcedevel`
[...]

Issue #79, opened 2017-12-26
Git Rate Limiting with Devel


ibrokemypie:
Running ``yay -Syu`` with the devel flag leads to git rate limiting. Perhaps there could be a method for generating and adding an oath2 token to extend the rate limiting from 60 to 5000 using https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization, and store that in config

PR #196, opened 2018-03-02
Improve printing

### LINKED ISSUE #128: light terminal background support
Hi there, the current colors are pretty hard to read in a light-on-dark terminal. An alternative color setting would be appreciated. Cheers.

Morganamilo:
Refactoring and general improvements to printing and specific improvements for light themes.
For more information see commits.

Before light:
image
After light:
image
Before dark:
image
After dark:
image

Do forgive the awful fonts. The pictures were takes with totally unconfigured terminals.

This should fix #128. Stuff like colours is highly opinion based, feedback would be great. Especially from @Jguer @ijanos @samosaara @heyitswither seems as you took intrest in the issue. I really dislike light themes myself so I would appreciate opinions from people who actually use them.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -579,8 +579,8 @@ func numberMenu(pkgS []string, flags []string) (err error) {
 		aurQ.printSearch(numpq + 1)
 	}
 
-	fmt.Println(boldGreenFg(arrow) + boldGreenFg(" Packages to not upgrade (eg: 1 2 3, 1-3 or ^4)"))
-	fmt.Print(boldGreenFg(arrow + " "))
+	fmt.Println(bold(green(arrow) + green(" Packages to not upgrade (eg: 1 2 3, 1-3 or ^4)")))
+	fmt.Print(bold(green(arrow + " ")))
 	reader := bufio.NewReader(os.Stdin)
 	numberBuf, overflow, err := reader.ReadLine()
 	if err != nil || overflow {
--- a/cmd.go
+++ b/cmd.go
@@ -579,7 +579,7 @@ func numberMenu(pkgS []string, flags []string) (err error) {
 		aurQ.printSearch(numpq + 1)
 	}
 
-	fmt.Println(bold(green(arrow) + green(" Packages to not upgrade (eg: 1 2 3, 1-3 or ^4)")))
+	fmt.Println(bold(green(arrow + " Packages to install (eg: 1 2 3, 1-3 or ^4)")))
 	fmt.Print(bold(green(arrow + " ")))
 	reader := bufio.NewReader(os.Stdin)
 	numberBuf, overflow, err := reader.ReadLine()
--- a/config.go
+++ b/config.go
@@ -142,12 +142,12 @@ func editor() string {
 		}
 		fallthrough
 	default:
-		fmt.Println(boldRedFgBlackBg("Warning:"),
-			boldYellowFgBlackBg("$EDITOR"), whiteFgBlackBg("is not set"))
+		fmt.Println(bold(red(blackBg("Warning:"))),
+			bold(yellow(blackBg("$EDITOR"))), white(blackBg("is not set")))
 		fmt.Println("Please add $EDITOR or to your environment variables.")
 
 	editorLoop:
-		fmt.Print(greenFg("Edit PKGBUILD with:"))
+		fmt.Print(green("Edit PKGBUILD with:"))
 		var editorInput string
 		_, err := fmt.Scanln(&editorInput)
 		if err != nil {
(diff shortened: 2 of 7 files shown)


Jguer:
I don't like light themes but when I give workshops and such they are the best to display on a projector so they have their use :laughing: . 

In regards to the colours, I like the yellow display of package names more than violet, but I need to check on my own colorscheme, probably doesn't look as dreadful as the tango theme but still we should strive to make it look good on the default colorscheme.

Morganamilo:
Yellow looks pretty bad on light theme so I thought the easiest thing was just to full out replace it so I picked a colour that wasn't in use.

It does look a little bright though Maybe if I swap the colours on the package names and numbers so it's a slightly darker shade? Or swap maybe use the same colour that packages are in`-Ps`?

Jguer:
What about package names cyan and version numbers purple, I haven't tested it

Morganamilo:
Bold cyan looks pretty pad on white:
image

Non bold looks decent:
image

I think both work on dark:
image

image

Jguer:
screenshot from 2018-03-02 13-03-00

screenshot from 2018-03-02 13-01-52

Actually looks decent on the first scheme. Maybe the non-bold blue will be better .I can already feel the complaints about the other theme being more familiar and better but for me it looks quite decent. 

An exception will need to be added to hash colour to not generate the same colour as package name otherwise it can be a bit hard on the eyes (check 6 community/kernel-headers-musl 3.12.6_5-1 (743.6 KiB 3.2 MiB) )

Morganamilo:
Changed some more stuff. Went with cyan over blue for the packages because blue looks pretty bad on dark. I don;t think we need to a just `colour/hash()` because I'm using bold for the repo name and non bold for the package name.

Pushed the changes so you can test them.

Jguer:
Yes, forgot I had bold turned off in my shell so I can't really tell the difference. But I like this last iteration. Let's say 1 day for any other input and merge?

Jguer:
On an unrelated note, I would split into [votes] (popularity) or some easier way of reading it

Morganamilo:
I grouped them together because they're very closely related. Same with download size and install size. But if you prefer we can come up with something else.

I still wanna look over this in a bit make sure it's consistent. All package names show up as cyan ect. I'm also considering refactoring away the colour functions for direct use.

Something like
```
func cyan(in string) string {
	if alpmConf.Options&alpm.ConfColor > 0 {
		return "\x1b[36m" + in + "\x1b[0m"
	}

	return in
}

packageColour := cyan
```

Then use packageColour() whenever printing package names. Could make it easier to change this stuff in the future. Probably won't implement this though at least for now. Just thinking about it.

ijanos:
Hi there, so as a light theme user here is my opinion. 

The most important parts should always use the terminal foreground color, which is usually black or grey for the light theme users.  The most important things in the yay output are the package names so they should use the color I picked for the main foreground color in the terminal. This yellow on the screenshots are barely readable. 

I suggest looking at the pacaur color theme, I think they figured this thing out and it works just fine on a light terminal. See the attached an image:

pacaur

Thank you for asking for input and taking light color schemes into consideration.

Morganamilo:
Alright so I got some decent fonts now so the screenshots don't look horrible.

As you suggested no colour just bold:
image

And here's cyan:
image

I do think the black is better there cause it stands out more.

same two for dark:
image

image

I do prefer the cyan here on the dark side but bold does work well here too so for the sake of compatibility it may be best to just do bold.

Morganamilo:
Pushed some more changes. I encourage you to build this branch yourself if you're able @ijanos .

ijanos:
Tried it, awesome work! 

yay

Morganamilo:
looks like aur [installed] is set to green and bold while repo [instaled] is just green whoops. Need to fix that. Apart from that glad you like it, will probably finalise it here unless anyone else suggests some changes.

ijanos:
Haha, I didnt even notice :) thank you again for this work.

Morganamilo:
I say we merge this now then. I'm still open to feedback but we can deal with that in later PRs before release. Maybe the -git people will get an update and give us some feedback.

Jguer:
I'm in for merging now

Morganamilo:
Alright merged. This is was the last issue marked as a bug that I think is fixable this release.
#181 and #79 are basically the same thing and will be fixed with #166 
#146 I have ideas for but nothing concrete.

Once things settle and it seems there's no new critical bugs we should make a release. Then work on the last remaining bug issues. Both shouldn't be very hard but will require some sizeable changes to the code which is not what this release is about.

Also we need a vendor update too right?

Jguer:
TODO:

- Vendor Update
- Update PKGBUILD with make
- Wait a bit for any waves
- Release

Morganamilo:
Don't forget the -git package. Should go get yourself added as a co-maintainer.

samosaara:
Um I'm not sure, but can we use the terminal's color pallete? Like 0 though 7 where 7th color is the bg and 1st is fg?

Morganamilo:
We do use the terminal's colour pallet. The screenshots I've been sending have all been using the default colour schemes.

In reality with my theme Yay looks like this:
image

Morganamilo:
@Jguer Also just remembered maybe update the screenshots on the readme?

PR #208, opened 2018-03-06
Add support to skip pkgbuild downloads.

Morganamilo:
If a pkgbuild is already in cache and matches the version on
the aur skip the download.

The version we check comes from the .SRCINFO file on disk which is never
updated. (updates through pkgver() edit the pkgbuild but do not effect
the .SRCINFO). Therefore if the the version of the .SRCINFO matches the
AUR's version there must not be  an update.

In the case of the on disk version being newer than the AUR version we
can assume user interaction and they probably do not want it overwitten
so in that case also skip the download.

Gonna update the docs tomorrow just making this PR now so you can get a look at it. And boy does `--gendb` sure go fast now.

### DIFF
--- a/install.go
+++ b/install.go
@@ -164,7 +164,7 @@ func install(parser *arguments) error {
 			return fmt.Errorf("Aborting due to user")
 		}
 
-		err = dowloadPkgBuilds(dc.Aur, dc.Bases)
+		err = downloadPkgBuilds(dc.Aur, parser.targets, dc.Bases)
 		if err != nil {
 			return err
 		}
@@ -399,9 +399,9 @@ func parsesrcinfosGenerate(pkgs []*rpc.Pkg, srcinfos map[string]*gopkg.PKGBUILD,
 	return nil
 }
 
-func dowloadPkgBuilds(pkgs []*rpc.Pkg, bases map[string][]*rpc.Pkg) error {
+func downloadPkgBuilds(pkgs []*rpc.Pkg, targets stringSet, bases map[string][]*rpc.Pkg) error {
 	for k, pkg := range pkgs {
-		if !config.ReDownload {
+		if config.ReDownload == "no" || (config.ReDownload == "yes" && !targets.get(pkg.Name)) {
 			dir := config.BuildDir + pkg.PackageBase + "/.SRCINFO"
 			pkgbuild, err := gopkg.ParseSRCINFO(dir)
 
--- a/install.go
+++ b/install.go
@@ -399,20 +399,35 @@ func parsesrcinfosGenerate(pkgs []*rpc.Pkg, srcinfos map[string]*gopkg.PKGBUILD,
 	return nil
 }
 
-func dowloadPkgBuilds(pkgs []*rpc.Pkg, bases map[string][]*rpc.Pkg) (err error) {
+func dowloadPkgBuilds(pkgs []*rpc.Pkg, bases map[string][]*rpc.Pkg) error {
 	for k, pkg := range pkgs {
-		//todo make pretty
-		str := bold(cyan("::") + " Downloading (%d/%d): %s\n")
+		if !config.ReDownload {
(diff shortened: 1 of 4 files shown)


install.go
```diff
@@ -399,20 +399,35 @@ func parsesrcinfosGenerate(pkgs []*rpc.Pkg, srcinfos map[string]*gopkg.PKGBUILD,
...
-		str := bold(cyan("::") + " Downloading (%d/%d): %s\n")
+		if !config.ReDownload {
+			dir := config.BuildDir + pkg.PackageBase + "/.SRCINFO"
+			pkgbuild, err := gopkg.ParseSRCINFO(dir)
+
+			if err == nil {
+				version, err := gopkg.NewCompleteVersion(pkg.Version)
+				if err == nil {
+					if !version.Newer(pkgbuild.Version()) {
+						str := bold(cyan("::") + " PKGBUILD up to date, Skipping (%d/%d): %s\n")
```
Jguer: Maybe just add 'Use --redownload to force download of PKGBUILD'

Morganamilo: You mean add it to the "PKGBUILD up to date, Skipping" string? I think it's kinda long enough already and I tend to dislike being over informative. Especially then the user is likely to see this string a ton of times.

It's detailed in the man page and usage so the user should be able to figure it out themselves.

Morganamilo:
I want to extend this before it's merged :
- Yes - set if you pass the flag twice
- no
- Targets - only re download targets, not dependencies

Also I purpose we use "yes"/"no" over true/false for all config options so that they can easily be extended if we need to in the future.

I'm kinda not sure how to implement this though with the way our config works.
passing `--redownload` once should set `redownload=targets` pasing it twice should set `redownload=yes`.

This is simple to implement but the one problem I see is if `redownload=targets` via the config then passing `--redownload` once would  bump it up from `targets` to `yes` when normally it would require it to be passed twice.

I don't like this behaviour it seems kinda unexpected and confusing and I don't really know how to fix it without larger changed to the config system.

Any thoughts?

Jguer:
It took me over 2 reads to understand the desired behaviour and I may be  slow sometimes but I believe it shouldn't work like that, it's overcomplicated. True and false is unambiguous for now. 
I think we can take this to an issue discussion on reforming the config system but for now leave it as is.

Morganamilo:
I don't think It's that complicated, passing  `--redownload` once will only re download the targets you pass to `-S`. Passing `--redownload` twice will redownload all the dependencies too. I don't think it's very important to have here but I really do want this for `--rebuild` which I plan to add at some point.

Take the following situation. I have some AU package installed that is now broken because the system libs have updated so this AUR package needs to be recompiled. Because I already installed it once, it and all its dependencies are in cache. Doing `yay -S package` will just reinstall the version in cache.

What I want to do is `yay -S package --rebuild` but I don't have the make dependencies installed. So even though I have the makedeps in cache yay is also going to try to rebuild all the makedeps when I just want to rebuild the one package.

Morganamilo:
also what would your opinion be on having these flags as run time options only and not stored in the config?

Jguer:
Thinking about it, wouldn't `--redownload` and `--redownload-all` make more sense, seems more intuitive.

Morganamilo:
Sure that would work. I'm a really big fan of Pacman's 'pass once to do x pass twice to do x but slightly differently'. But yeah I'll do it as you said with two separate flags.

Would you prefer it be part of the config or only a run time flag?

Morganamilo:
I could implement the `--save` option we talked about before and then there would be no need to run time only flags.

Jguer:
Fine by me, that would justify the bump to v4 (as if it wasn't more than justified already). I think if it doesn't bother you to implement it, we can run it as a trial in master

Morganamilo:
Alright I'll keep this as is and merge it then and add `--save` in a new PR

PR #213, opened 2018-03-07
Add GitBin and tweak defaults

Morganamilo:
Use cacheHome for builddir instead of hardcoding ~/.cache
Use the command names in config.*Bin options.

For example PacmanBin is changed to just "pacman" this means yay will
call the pacman binary in PATH. If the user wants to use a different
binary they can still specify a full path in the config.


Morganamilo:
@Jguer  I'm happy to merge this, I just want the okay for the changes to defaults.

Jguer:
I'm okay with this as well :smile:

PR #220, opened 2018-03-08
Add flags for all missing config options

Morganamilo:
With this The user should never have to manually edit the yay config
file. All options can be set directly through yay using
`yay --<option> <value> --save`


Jguer:
Ok, after this merge, I'll update completions and release V4

Morganamilo:
Did you start on this? I can take a look at the zsh completion if you like.

Jguer:
Fine by me, I'll update fish completions first

PR #225, opened 2018-03-08
Use goroutinuies for devel updates

Morganamilo:
The update process was already partly parallel: repo, aur and devel
updates where each given their own goroutine.

This gives two further layers of parallelization. Firstly the
`needsUpdate()` function is now run in parallel for each package checked.
One package may have many vcs sources so `needsUpdate()` also checks all
of its sources in parallel.

This gives an aproxamte 3x speedup for `yay -Su --devel` timing from
when the command starts to when the number menu apears.

unfortunately git://bitbucket.org/volumesoffun/polyvox.git never
resolves on my machine for some reason so it always hits the 5 second
timout period.

It then moves on to http:/bitbucket.org/volumesoffun/polyvox.git/ which
does resolve as expected. I have not looked into it but I fear this
applies to all gitbucket repos. Luckly they are few and far between.

### DIFF
--- a/query.go
+++ b/query.go
@@ -338,21 +338,18 @@ func aurInfo(names []string) ([]rpc.Pkg, error) {
 	outOfDate := make([]string, 0, len(names))
 
 	makeRequest := func(n, max int) {
+		defer wg.Done()
 		tempInfo, requestErr := rpc.Info(names[n:max])
 		if err != nil {
-			wg.Done()
 			return
 		}
 		if requestErr != nil {
-			//return info, err
 			err = requestErr
-			wg.Done()
 			return
 		}
 		mux.Lock()
 		info = append(info, tempInfo...)
 		mux.Unlock()
-		wg.Done()
 	}
 
 	for n := 0; n < len(names); n += config.RequestSplitN {
--- a/upgrade.go
+++ b/upgrade.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"sort"
+	"sync"
 	"unicode"
 
 	alpm "github.com/jguer/go-alpm"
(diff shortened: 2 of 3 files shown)


Morganamilo:
I'm not sure how much of a speed up this will give others, It's going to depend on how many vcs packages you have installed. For me most of the speed up probably comes from [vim-youcompleteme-git](https://aur.archlinux.org/packages/vim-youcompleteme-git/) which was a whopping 16 vcs sources:
```bash
source=('git+https://github.com/Valloric/YouCompleteMe.git' #ycm
        'git+https://github.com/ross/requests-futures.git'  #ycm
        'git+https://github.com/Valloric/ycmd.git'          #ycm
        'git+https://github.com/kennethreitz/requests.git'  #ycmd
        'git+https://github.com/bottlepy/bottle.git'        #ycmd
        'git+https://github.com/slezica/python-frozendict.git'    #ycmd
        'git+https://github.com/PythonCharmers/python-future.git' #ycmd
        'git+https://github.com/vheon/JediHTTP.git'               #ycmd
        'git+https://github.com/davidhalter/jedi.git'             #jediHTTP
        'git+https://github.com/Pylons/waitress.git'              #ycmd,jediHTTP
        'git+https://github.com/nsf/gocode.git'                   #ycmd
        'git+https://github.com/Manishearth/godef.git'            #ycmd
        'git+https://github.com/nosami/OmniSharpServer.git'       #ycmd
        'git+https://github.com/icsharpcode/NRefactory.git'       #OmniSharpServer
        'git+https://github.com/jbevain/cecil.git'                #OmniSharpServer
        'git+https://github.com/jwilm/racerd.git'                 #ycmd
        )
```

Morganamilo:
I changed up the implementation a bit. Gives about a 0.25x speedup on my end and this speed up will only affect packages with many vcs souces. Anyway a speedup is a speedup.

I already requested a review but I normally use WaitGroups over channels so I would appreciate it if you can check if the channel logic is correct, thanks.

Jguer:
I'm all in for speedups. Tomorrow I have some time to dedicate to `yay`

Jguer:
Functionality wise all seems fine. Survived `--gendb` and `-Syu --devel` on my machine. I believe hasUpdate() can be made more readable with the sync package, but just a question of preference. I can try and brush it up a bit after the merge

Morganamilo:
Well originally I did use the sync package but I switched it to channels for a little extra speedup as explained in 954b297. I would prefer WaitGroups but I don't think it can be done using them.

for me the first commit saw a x3 speedup and the second give an additional x0.25 speedup. That's for the whole update fetching not just devel updates. Would you mind benchmarking it on your end? I have 27 sources in my vcs.json. Obviously the bigger that number the bigger the speedup,

PR #230, opened 2018-03-10
Add info for all files and directories that used

Morganamilo:


aswild:
Looks good to me, thanks!

Always nice to document as much as possible in the man page :)

PR #239, opened 2018-03-12
Ensure targets are never listed as make depends

Morganamilo:
when doing `yay -S foo bar` where `bar` is a makedep of `foo` in rare
cases bar would be listed as [aur make] even thought it was explicitly
requested.

### DIFF
--- a/dependencies.go
+++ b/dependencies.go
@@ -131,6 +131,10 @@ func getDepCatagories(pkgs []string, dt *depTree) (*depCatagories, error) {
 		})
 	}
 
+	for _, pkg := range pkgs {
+		dc.MakeOnly.remove(pkg)
+	}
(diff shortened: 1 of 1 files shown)


Jguer:
I'm beginning to feel comfortable with this rev. Do you any more impending patch? Otherwise I'm done with bash completions and some formatting today and release tomorrow

Morganamilo:
I don't think so, the last few rapid fire PRs have just been small things I noticed and fixed then and there. No need to wait on me, make the release whenever you want.

PS have you thought about using annotated tags? More meta date, more descriptive and signable.

Jguer:
I think we can have a release day tomorrow.

Annotated tags seem nice for the next release.

Issue #87, opened 2018-01-04
RFC: Detect dependency upgrades and rebuild needed AUR & devel packages too


maximbaz:
Recently I experienced the following: I have an AUR package A installed, that depends on a package B from the official repositories. Package B was updated, so I upgraded it with yay. However package B provides now a different *.so file, so with upgrading B the package A started crashing in runtime.

The fix was as simple as rebuilding package A, but it took me some time to figure out this.

My proposal is the following: I would like yay to track dependencies of AUR and devel packages, and when their dependencies update, suggest to rebuild these packages too. What do you think about it?

z3ntu:
Also the ruby version upgrade from 2.4 to 2.5 made a bunch of AUR package rebuilds needed.

AladW:
https://github.com/trizen/trizen/issues/86#issuecomment-360409395

Morganamilo:
Forgot this was still open, closing in favour of #243.

PR #243, opened 2018-03-14
 Add --rebuild flag

Morganamilo:
Similar to the --redownload flag, when specifed targets will be rebuilt
even if an up to date version is cached. --rebuildall can be used to
ensure uninstalled dependencies are rebuilt as well.

Additionally, unlike --redownload there is also --rebuildtree. This
causes a rebuild and reinstall of a package and all of it's dependencies
recursivley. This is designed for when a libary updae, breaks an
installed AUR package due to a partial upgrade. polybar is a common
example

--rebuild allows you to easily skip the cache and rebuild against a newer
libary version. --rebuildtree is a more nuclear option where you can
rebuild the whole dependency tree.

This is my purposed fix for #87 It does not really do what the issue asks because that would be a project in itself. These flags should provide an easier way to mange such things manually though.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -397,6 +397,14 @@ func handleConfig(option, value string) bool {
 		config.ReDownload = "all"
 	case "noredownload":
 		config.ReDownload = "no"
+	case "rebuild":
+		config.ReBuild = "yes"
+	case "rebuildall":
+		config.ReBuild = "all"
+	case "rebuildtree":
+		config.ReBuild = "tree"
+	case "norebuild":
+		config.ReBuild = "no"
 	case "mflags":
 		config.MFlags = value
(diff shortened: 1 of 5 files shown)


Jguer:
I like, haven't reviewed the code but cmus-git is always breaking due to the lack of this

Morganamilo:
One thing I'm not too sure on is the name. I went with `--rebuildchain` for some reason, although thinking about it most of the internal code uses the word tree. So for sake of consistency I'm thinking of changing it to `--rebuildtree`. Do you have a preference?

Jguer:
`--rebuildtree` sounds more accurate

Morganamilo:
Alright, changed.

PR #246, opened 2018-03-15
Experimental Concurrent Source downloading

Jguer:
To-fix: Clobbered display messages

Display messages are a mess from makepkg. It would be possible to hide all output and just present which goroutines are still active, but some detail is left out

### DIFF
--- a/install.go
+++ b/install.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"strconv"
 	"strings"
+	"sync"
 
 	alpm "github.com/jguer/go-alpm"
 	rpc "github.com/mikkeloscar/aur"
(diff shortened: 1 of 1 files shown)


Morganamilo:
Have you benchmarked this at all? Would like to see some real world results.

Jguer:
Have only slightly benchmarked on my $(yay -Qemq) seems to be faster, I still want to mature this a bit and get some real numbers behind it in diverse situations before marking it pull ready.

Morganamilo:
So I did some testing.

ran `yay -Qemq | time ./yay --mflags --skipchecksums -S -` (the checksums for one of the packages i have installed is currently broken) and ctr+c'd as soon as the source download finished.

All the pkgbuilds were already cached but none of the pkgbuild sources were.

That comes to 52 different pkgbases, `ttf-noto ` being the largest.

Rebased this branch against master and tested against master:
    This branch: 7:42.05
    Master: 10:37.07

Not a bad speedup, there needs to be some improvement on the erroring though. Right now if one thing fails it prints the error at that point in time and the rest of the go routines continue. If one call to makepkg errors the rest should exit. I'm not sure how makepkg downloads files though, it may download everything to a temp directory then move it when it finishes or it may download it in the final location from the start. If we interrupt a half downloaded package will it mess things up?

If we're able to implement this in a safe way and with some nicer printing this could probably be extended to the pkgbuild download too.

Morganamilo:
Running `ay -Qemq | time ./yay --mflags --skipchecksums -S -` then doing ctr+c lots of makepkg instances get left running in the background, stuck forever with no output to the terminal. Weirdly the same does not happen if not running Yay through time.

Also if you do `pkill yay` while its running, all the makepkg instances keep running, keep displaying to the terminal and exit when finished. I've realised the happens to all exec calls.

Jguer:
Yes, I think this can be extended for PKGBUILD. Away from workstation during the weekend, but I'll keep working on this from Monday.

Jguer:
Closing this pull request until more satisfactory results are achieved

AladW:
Does Go have some GNU parallel equivalent? Parallel computing is complicated, so if you really wanted this using some existing effort is a good idea.

In my experience how much it speeds up things heavily depends on the user's machine. If you're on some low powered VPS you'll hardly notice anything, but if you're on some 32 core desktop it's a dramatic improvement.

Jguer:
There's no problem in synchronization and what we are leveraging is mostly the bandwidth of the user and sometimes single thread extraction. My problem with this is still on displaying the progress to the user in a straightforward way and defining when concurrent download and PKGBUILD sourcing should be used (more than 4 packages, more than 8 packages?)

PR #256, opened 2018-03-17
Updates on PGP keys

qrwteyrutiyoup:
1 - Fix tests after #254 
2 - Add --gpg and --gpgflags alongside with respective entries in the manual and usage
3 - Update completion files for these flags

### DIFF
--- a/bash-completion
+++ b/bash-completion
@@ -68,7 +68,7 @@ _yay() {
   s u')
   common=('arch cachedir color config confirm dbpath debug gpgdir help hookdir logfile
            noconfirm noprogressbar noscriptlet quiet save mflags buildir editor
-           makepkg pacman tar git config requestsplitn sudoloop nosudoloop
+           makepkg pacman tar git gpg gpgflags config requestsplitn sudoloop nosudoloop
            redownload noredownload redownloadall root verbose' 'b d h q r v')
   core=('database files help query remove sync upgrade version' 'D F Q R S U V h')
 
--- a/cmd.go
+++ b/cmd.go
@@ -46,6 +46,7 @@ Permanent configuration options:
     --pacman   <file>    pacman command to use
     --tar      <file>    bsdtar command to use
     --git      <file>    git command to use
+    --gpg      <file>    gpg command to use
     --config   <file>    pacman.conf file to use
 
     --requestsplitn <n>  Max amount of packages to query per AUR request
@@ -63,9 +64,10 @@ Permanent configuration options:
     --noredownload       Skip pkgbuild download if in cache and up to date
     --rebuild            Always build target packages
     --rebuildall         Always build all AUR packages
-    --rebuildtree       Always build all AUR packages even if installed
+    --rebuildtree        Always build all AUR packages even if installed
     --norebuild          Skip package build if in cache and up to date
     --mflags <flags>     Pass arguments to makepkg
+    --gpgflags <flags>   Pass arguments to gpg
     --sudoloop           Loop sudo calls in the background to avoid timeout
     --nosudoloop         Do not loop sudo calls in the background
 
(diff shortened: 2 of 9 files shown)


Morganamilo:
Not home and on mobile so I cant do a full review, but one thing it looks like you need to add `gpgflags` to `hasparam()` in parser.go.

qrwteyrutiyoup:
Oops, indeed, need to update hasParam, thanks.

Morganamilo:
>One thing I've noticed is because we count missing keys as 'a non 0 return from gpg'. If the user sets --gpgflags to something invalid, it will count keys as invalid when really it's a gpg error.

>What's more annoying is that gpg seems to return the same error code for missing keys and invalid flags. I don't know of a way to handle this right now.

This still applies, it appears I made the comment on the commit and not the PR so it disappeared  when you rebased. It's not a big deal but something to point out.

I'm actually on a real PC this time and not mobile. So I have had a proper look. Thanks for the test fix, I forgot to check them. Everything else apart from the above point seems in order, I'll merge it now.

qrwteyrutiyoup:
> This still applies, it appears I made the comment on the commit and not the PR so it disappeared when you rebased. It's not a big deal but something to point out.

Indeed, `gpg` does not make it easy to know what exactly failed, if the retrieving of keys itself or simply incorrect/ivalid parameters. 

How about adding something like "Please double check the extra arguments provided to gpg" to the error message, when `gpgflags` is set to something custom? Might at least give a better clue, in case of an error.

Morganamilo:
I'm not a fan of warnings like that, if the user changes gpgflags I expect them to know what they're doing. Having to see a message every time just in case somebody does something wrong feels annoying and a little patronising.

Personally I would just mention it in the manpage where there's room for detailed information and the user is free to check at any time.

Morganamilo:
Also thinking about it. The first `gpg` call will fail causing Yay to think it needs the key. But the second call where it imports the key is displayed to the user. So gpg should display a decent error message there and Yay will exit.

qrwteyrutiyoup:
Maybe we could use something like `gpg <gpgflags> --list-keys` as we enter `checkPgpKeys` as some sort of sanity check, before doing the actual checking/importing of keys. If it fails in there, it would be likely due to bad flags. 

`gpg` actually returns a string like `gpg: invalid option "--foo"` in this case, so we could use that message in `yay`'s error message.

Morganamilo:
A pre-check could work, although I still don't really think it's really needed.

Parsing the error output I would rather not do, there's a ton of things that can go wrong other than invalid args, we can't check for them all. Exit status' were made for this but sadly gpg only ever returns 0, 1 or 2.

qrwteyrutiyoup:
Ok, so I won't do anything here for now; it looks good the way it is currently IMO:
```
./yay -S --gpgflags "--foo" cower
[...]
==> Import?  [Y/n] 
:: Importing keys with gpg...
gpg: invalid option "--foo"
==> Error: Problem importing keys
```

Morganamilo:
I am right in assuming the fish completion is still not updated right?

qrwteyrutiyoup:
Right. I thought I had done it as well, but just checked out and I updated bash and zsh only.

Morganamilo:
I just assumed you didn't do it because you don't run fish. I don't run fish either so that's why I normally let @Jguer do it. Not that its that hard to copy and paste a line and change the flag name.

I don't mind if you do it or not, just wanted to make sure this doesn't get forgotten about.

Jguer:
Sweet, I run `fish` full time so I can test it out. Anymore place missing completions/documentation entries before release?

qrwteyrutiyoup:
Just spotted a rogue `/n’` in the beginning of the man page. Other than that and this completion, I am not aware of something else :)

Morganamilo:
Strange, how did I manage that? Through sed maybe?

I'll change it back to `'\" t` then.

Morganamilo:
Anyway apart from that. I think the only new stuff we have is: `--gpg` `-gpgflags` `--rebuild*` and `-Qu`. I think the only thing missing is `--rebuild*`.

Issue #262, opened 2018-03-20
Upgrade inconsistency


samosaara:
I really like the new install algorithm, but I think this new strategy of minimizing user interaction generated some inconsistencies when upgrading the system, with the new strategy I just assumed after I choose the packages I wanted to install it would just confirm everything with me to then do it's job completely unattended,  but when `-Syu`'ing It's kinda boring to select everything, do the pacman part, wait for it to finished an then come back to do the aur packages prompts. Sure, I do can `--no-confirm` but then I can't select packages or review some of the pkg builds.

Morganamilo:
You seem to have messed up a little there....

samosaara:
Yeah miss-clicked and submitted the issue before typing it, but one of the solutions I can propose would be doing the pkg build's reviewing prompt thingie beforehand to confirm everything and then actually do all the updating BUT! with `pacman -S --noconfirm [selected pacman packages to update]`

PR #264, opened 2018-03-21
Move repo install to after numbermenus

Morganamilo:
Implements #262


Morganamilo:
@samosaara What do you think about this?

samosaara:
Flawless. Thanks!

PR #265, opened 2018-03-21
Refactor and improve the upgrade menu 

### LINKED ISSUE #218: Package appears in aur and devel during -Suy --devel
When a devel package is update on the AUR page and there are new commits the package appears in the `aur` and `devel` categories.

I think it's probably best to remove the entry from `aur` and keep the `devel` entry as the devel is going to be equal or newer. 

Applies to current release (3.440):
image

And latest master (3.440.r20.gd3f0053):
image

Morganamilo:
Fixes #218 

The goroutine code is simpler now so hopefully I didn't make a dumb mistake but their still easy to mess up so I'd like a second opinion before merging.

See commits for a more detailed description.

### DIFF
--- a/upgrade.go
+++ b/upgrade.go
@@ -28,9 +28,18 @@ func (u upSlice) Len() int      { return len(u) }
 func (u upSlice) Swap(i, j int) { u[i], u[j] = u[j], u[i] }
 
 func (u upSlice) Less(i, j int) bool {
-	iRunes := []rune(u[i].Repository)
-	jRunes := []rune(u[j].Repository)
+	if u[i].Repository != u[j].Repository {
+		iRunes := []rune(u[i].Repository)
+		jRunes := []rune(u[j].Repository)
+		return lessRunes(iRunes, jRunes)
+	} else {
+		iRunes := []rune(u[i].Name)
+		jRunes := []rune(u[j].Name)
+		return lessRunes(iRunes, jRunes)
+	}
+}
 
+func lessRunes(iRunes, jRunes []rune) bool {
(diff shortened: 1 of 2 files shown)


upgrade.go
```diff
@@ -83,52 +84,62 @@ func getVersionDiff(oldVersion, newversion string) (left, right string) {
...
-	errC := make(chan error)
+	var wg sync.WaitGroup
+	var develUp upSlice
+
+	var repoErr error
+	var aurErr error
+	var develErr error
 
 	fmt.Println(bold(cyan("::") + " Searching databases for updates..."))
+	wg.Add(1)
```
Jguer: Just a small detail, here maybe wg.Add(2) and immediately launch the 2 go routines and then print Searching databases for updates and Searching AUR for updates.

Jguer:
Don't have many updates to test it on now, from a code perspective it seem good, I can try and update using this version for another day, but I don't see any issue with it

Morganamilo:
Added a new commit 95bbc0e that fixes #146. Using the method I described in #146.

upgrade.go
Jguer: Was just to leave the main go routine a bit more entertained, I don't object to leaving it as is

Jguer:
Worked fine, on a mix of repo,aur and devel packages

Morganamilo:
As mentioned in 98ea801 This should fix the replaces problem in #146. I couldn't really test it at the time because replaces don't come up that often.

image

It finally came up though so I can confirm it's working.

PR #269, opened 2018-03-22
 More conflict checking

Morganamilo:
This commit extends the conflict checking a lot, it adds support for:
- Conflicting with provides as well as actual package names
- Reverse conflicts
- Inner conflicts

Both normal conflicts and inner conflicts are run in parallel.
Messages are now printing when checking conflicts.

This also fixes packages sometimes being listed as conflicting with
themselves.

The inner conflict is a little verbose and could be toned down a little
but I am insure exactly how to tone it down.

There's more code here than I would like, mainly because of the annoyingness of dealing with both `rpc.Pkg` and `alpm.Package` types at the same time.

I did some testing:
- `yay -Qq | yay -S -` Lists no conflicts
- `yay -S ros-lunar-desktop-full` Lists no conflicts
- `yay -S superproductivity superproductivity-git` lists inner conflicts
- `yay -S git git-git` lists inner conflicts:
- `yay -S yay yay-git yay-bin` lists inner conflicts
- `yay -S superproductivity` with `superproductivity-git` installed lists a conflict and vice versa
- `yay -S wine wine-staging` does not list inner conflicts:
   - This is due to some weirdness with wine never making it into the dc in the first place, probably because wine-staging already provides wine so it doesn't bother? Needs looking into but not a bug with this pr.


Jguer:
Tested it, Inner conflicts don't seem too verbose

Morganamilo:
I was thinking of cases like this being to verbose:
image

`yay` is checked for inner conflicts and it shows it conflicts with `yay-git` and `yay-bin`
`yay-bin` is checked for inner conflicts and it shows it conflicts with `yay-git` and `yay`
`yay-git` is checked for inner conflicts and it shows it conflicts with `yay` and `yay-bin`
So makes sense but a bit over the top I think.

the inner function add is redefined a lot too, I was thinking of making a new file, util.go, to dump all the random functions like min() and max().

Jguer:
I like the new file plan, our package files have been swelling a bit. I used to have something like that and I know it applies to packages [https://blog.golang.org/package-names](https://blog.golang.org/package-names) made me destroy it.

>  Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them. 

Maybe since we have a few more comparison functions, a `comparison.go`?

Morganamilo:
I was thinking of just having a file `utils.go` but still be part of the main package so I don't know if that applies.

Jguer:
Yes, technically it doesn't apply, just if we could make a better code organization we should try or we'll be dumping functions into `utils.go` in no time. 

That being said `pacman` also has it's own, if there is no where else to put it I'll :see_no_evil:

Morganamilo:
I thought `utils` because we really don't have that many misc functions, we could always split utils out in future if it gets too big.

On the topic of refactoring we have quite a lot of files so I would like to see them split into directories.

Something like:
```
/repo/
    README.md
    .gitignore
    ect...
    completions/
        fish
        zsh
        bash
     main/yay
        *.go
     doc/
        yay.8
     vendor/
        vendorstuff
```

Jguer:
I like that though, I'll take care of that since I also have to update the PKGBUILDs. You can create the `utils.go`

Jguer:
Let's stabilize  for a release @Morganamilo ?

Morganamilo:
One last code change apart from refactoring that I want to do is move gendb out of install to its own function. It always feels like it gets in the way in install and you have to be careful when tweaking install as to not break gendb.

Having it in its own function should make it harder to break.

Jguer:
I agree with that refactor, I think after that we can start preparing a release.

PR #272, opened 2018-03-22
Update version for close release

Jguer:
Completions and man page moved. PKGBUILD for `yay-git` already fixed  
Signed-off-by: Jguer <me@jguer.space>


Morganamilo:
What about moving the *.go files into a dir called yay or main?

Jguer:
`main` I think creates problems executable naming wise (at least it did ~go1.7).    
`yay` turns into `go get github.com/jguer/yay/yay` and  
 if memory serves  and `go get github.com/jguer/yay` returns not a go package. 

I don't think it's necessary to move the go files, not much win to be had.

(ps: I think the vendor directory would have to be moved inside the yay folder for `dep` to work as intended, and for yay/yay to use the vendored packages)

Morganamilo:
Alright fair enough then, I really hate the GOPATH stuff.

Jguer:
I know, I hate it too and everything related to automated testing with alpm, vendoring and structure expansion makes me foam from the mouth :laughing:

Issue #268, opened 2018-03-21
yay -Yc removes optionally depended packages and does not remove them recursively


ratorx:
#### Affected Version

yay v4.505 (from the AUR)
yay v3.373 (build from source)

#### Issue
yay -Yc attempts to uninstall packages that are optionally depended on by another package. Also, dependencies are not removed recursively (i.e. `yay -Yc` tries to remove the set of packages that would be removed by `pacman -R $(pacman -Qttdq)` rather than the more useful (in my opinion) `pacman -Rns $(pacman -Qtdq)`)

Is this a deliberate choice?

In addition to this, it doesn't warn when removing optional packages like `pacman -Rns $(pacman -Qttdq)` does.

#### Steps to reproduce

1. Install any package which has optional depends and then the optional dependency (E.g. `neovim` and `python-neovim`). In particular, install using `pacman -S neovim` and `pacman -S --asdeps python-neovim`.
2. Use `yay -Yc`

#### Possible fixes
* Use pacman commands to remove dependencies (Make `yay -Yc` run `pacman -Rns $(pacman -Qtdq)` instead of a call to libalpm).
* Use `alpm_pkg_compute_requiredby` as well as `alpm_pkg_compute_optionalfor` from libalpm when checking each candidate package for dependencies.

Morganamilo:
This is a thing I noticed a while ago but hadn't really thought to do anything about it because I hardly use this feature so have focused on other things.

Now there's an open issue that ups the priority of fixing it, I'll look into it soon.

ratorx:
If there are more important issues to fix, I can have a go at this one. The alpm fix seems pretty simple to implement, although will require a commit in the go-alpm repo as well. Just wanted to make sure it wasn't deliberate before making a pull request.

Morganamilo:
I'm assuming it was just done that way because it was easier to implement. Even if it was deliberate, feedback is always welcome and we're open to change. I would rather it keep optional dependencies too.

If you're gonna implement this yourself I have a couple things I'd like to see it do:
- As mentioned don't remove optional deps.
- Do remove optional deps if `c` is passed twice
- Instead of relying on  `-Rs` to recursively remove deps, figure out the deps recursively inside of yay. This allows yay to print everything It's going to remove.
- Print the total remove size

ratorx:
Should yay print the total remove size before passing the arguments to pacman? This seems slightly redundant as pacman already prints the total removed size when `pacman -R` is run.

[...]

PR #275, opened 2018-03-22
Recursively remove dependencies when using yay -Yc

ratorx:
Fix https://github.com/Jguer/yay/issues/268

* Recursively removes unneeded dependencies
* Preserves optional dependencies by default (overridden with -cc)
* Does not list package removals, just passes them into pacman and relies on pacman confirmation dialog

### DIFF
--- a/query.go
+++ b/query.go
@@ -299,13 +299,7 @@ func hangingPackages(removeOptional bool) (hanging []string, err error) {
 		}
 
 		pkg.Provides().ForEach(func(dep alpm.Depend) error {
-			if deps, ok := provides[dep.Name]; ok {
-				deps.set(pkg.Name())
-			} else {
-				ss := make(stringSet)
-				ss.set(pkg.Name())
-				provides[dep.Name] = ss
-			}
+			addMapStringSet(provides, dep.Name, pkg.Name())
 			return nil
 		})
 		return nil
--- a/query.go
+++ b/query.go
@@ -274,26 +274,89 @@ func packageSlices(toCheck []string) (aur []string, repo []string, err error) {
 
 // HangingPackages returns a list of packages installed as deps
 // and unneeded by the system
-func hangingPackages() (hanging []string, err error) {
+// removeOptional decides whether optional dependencies are counted or not
+func hangingPackages(removeOptional bool) (hanging []string, err error) {
 	localDb, err := alpmHandle.LocalDb()
 	if err != nil {
 		return
 	}
 
-	f := func(pkg alpm.Package) error {
-		if pkg.Reason() != alpm.PkgReasonDepend {
+	// safePackages represents every package in the system in one of 3 states
+	// State = 0 - Remove package from the system
+	// State = 1 - Keep package in the system; need to iterate over dependencies
+	// State = 2 - Keep package and have iterated over dependencies
+	safePackages := make(map[string]uint8)
+	// provides stores a mapping from the provides name back to the original package name
+	// Assumption - multiple installed packages don't provide the same dependency
+	provides := make(map[string]string)
+	packages := localDb.PkgCache()
+
+	// Mark explicit dependencies and enumerate the provides list
+	setupResources := func(pkg alpm.Package) error {
+		if pkg.Reason() == alpm.PkgReasonExplicit {
+			safePackages[pkg.Name()] = 1
+		} else {
+			safePackages[pkg.Name()] = 0
+		}
+
+		pkg.Provides().ForEach(func(dep alpm.Depend) error {
+			provides[dep.Name] = pkg.Name()
+			return nil
+		})
+		return nil
+	}
+	packages.ForEach(setupResources)
+
+	iterateAgain := true
+	processDependencies := func(pkg alpm.Package) error {
+		if state, _ := safePackages[pkg.Name()]; state == 0 || state == 2 {
 			return nil
 		}
-		requiredby := pkg.ComputeRequiredBy()
-		if len(requiredby) == 0 {
-			hanging = append(hanging, pkg.Name())
-			fmt.Println(pkg.Name() + ": " + magenta(human(pkg.ISize())))
 
+		safePackages[pkg.Name()] = 2
+
+		// Update state for dependencies
+		markDependencies := func(dep alpm.Depend) error {
+			// Don't assume a dependency is installed
+			state, ok := safePackages[dep.Name]
+			if !ok {
+				// Check if dep is a provides rather than actual package name
+				if p, ok2 := provides[dep.Name]; ok2 {
+					iterateAgain = true
+					safePackages[p] = 1
+				}
+
+				return nil
+			}
+
+			if state == 0 {
+				iterateAgain = true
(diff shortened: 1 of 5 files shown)


Jguer:
Any easy packages to install and uninstall to test functionality?

Morganamilo:
I'm trying with git, has a lot of optional stuff.

Morganamilo:
Something doesn't quite work here. Yay should persevere optional dependencies for stuff we are keeping. But if we are removing `foo` I want the un-needed  optional deps for `foo` uninstalled also.

For example I have everything installed as a dependency, not one package on my system is listed as explicit. Yet doing `yay -c` It wants to keep the opt depends for git installed, even though git is going to be removed by `yay -c`

Easy way to reproduce:
```
mkdir ~/db
yay -Sy -b ~/db
yay -S -b ~/db --asdeps tk perl-libwww perl-term-readkey perl-mime-tools perl-net-smtp-ssl perl-authen-sasl perl-mediawiki-api perl-datetime-format-iso8601 perl-lwp-protocol-https perl-cgi python2 subversion gnome-keyring git
yay -c
```

Also I would rebase out the merge commits, they're just clutter really.

ratorx:
I had a look into it. It seems to be a problem with `pacman -Rs` rather than any code in yay.

I tested this by doing:
```
pacman -S --asdeps git perl-term-readkey # Just one of the optional dependencies to be minimal
```

When `pacman -Rs git`, only direct dependencies are removed (in my case `perl-error`)
However, this leaves `perl-term-readkey` unneeded dependencies behind (found by both `yay -c` and `pacman -Qdt`).

A potential fix would be to implement a recursive walk in yay (what I had originally), instead of relying on `pacman -Rs`.

I found another problem however - dependency cycles are not resolved.

As an example, if I do `pacman -S --asdeps neovim python-neovim` and then try to remove unnecessary packages (using either `pacman -Rs $(pacman -Qtdq)` or `yay -c`), neither of them picks these packages up. I would have to think more about how to fix this, but if you have any ideas, let me know.

Morganamilo:
Doing it recessive is probably the best way to deal with it. Not sure on the exact implementation though.

As for circular stuff. Maybe keep a cache of your current tree, then as you work your way up if you find something that's in that cache assume it's not needed. No idea if it will work, just an idea.

Morganamilo:
As for implementation my current thought is: assume everything is to be removed, iterate over explicit packages and mark everything in there to be kept, remove what is not to be kept.

Also as well as keeping optional deps, I think we should keep makedeps for AUR packages.

ratorx:
That seems very doable. It should deal with circular and optional dependencies as well.

Can you point me towards any existing code that gets makedeps from the AUR? My current understanding is that it can't be accessed through libalpm?

Morganamilo:
You can do it in the latest git for alpm :(

Infact I was using the git version while developing go-alpm and got confused when the functions didn't work, hence:
image

You'll have to use `aurInfo()` to get the makedeps.

Morganamilo:
Any reason you closed this?

ratorx:
It was accidental, I undid some of the merge commits and rebased, but I'll leave it closed for now. I'll reopen when I have a fixed version.

Morganamilo:
Alright, if you're starting over I would recommend `git pull --rebase` over `git pull`.

ratorx:
Will do. I've got the basic thing recursion package marking working, just need to add support for AUR makedeps.

ratorx:
I had a look into getting makedeps for AUR packages. Considering that using aurinfo involves making web requests and adds a fair bit of complexity to determining hanging packages, do you think its worth it to determine and keep makedeps?

The alternative would be to wait for stable libalpm to be updated so that the bindings in go-alpm work and then integrate that.

Morganamilo:
I don't really mind either way.

I wouldn't say using `aurInfo()` adds that much complexity though, the function is pretty easy to use. The internet access is the more annoying thing, you wouldn't expect Yay to need that for removing packages.

Pacman v5.0.2 was released `Sat Jun 3 22:17:02 2017 ` so who knows when the next update will be.

ratorx:
I've redone it now, and tested it a bit - it seems to work!

Morganamilo:
Looking at the vendor update it looks like you edited the file manually. Instead you should run `dep ensure -update` to update the deps.

Morganamilo:
Also run `make test` and to lint `go fmt`.

ratorx:
Ahh sorry, didn't realise. Done both now.

Morganamilo:
Tested the code a little and it seems like it works alright.

This is false though:
```
	// Assumption - multiple installed packages don't provide the same dependency
```

So this would probably fail on something like java.

ratorx:
Ahh alright, I can work around that assumption, just wasn't sure whether I needed to when I wrote the code.

query.go
```diff
@@ -300,7 +299,13 @@ func hangingPackages(removeOptional bool) (hanging []string, err error) {
 		}
 
 		pkg.Provides().ForEach(func(dep alpm.Depend) error {
-			provides[dep.Name] = pkg.Name()
```
Morganamilo: There's a function called `addMapStringSet` that does all this for you ;)

feel free to create `addMapStringString` as well or find a generic way.

ratorx: I was not aware of `addMapStringSet` and I'll rewrite my code.

I'm not too sure what you mean by `addMapStringString`, or which part of my code you think I should use it in.

Morganamilo: Yeah I wasn't thinking when I said that. Was thinking about `map[stiring]string` but obviously you can set that directly.

Jguer:
Let's leave this patch for the next release cycle as it's still undergoing changes, you're the closest on it @Morganamilo , is that ok with you?

Morganamilo:
That was my plan already. Would like my two PRs merged though.

Morganamilo:
I gave this a test again and it seems to work fully now so nice job. I still need to make sure though and give the code a proper read.

Morganamilo:
Also is there any reason you're not using ComputRequiredBy/ComputreOptionalFor anymore?.

ratorx:
In order to deal with cyclic dependencies, I switched strategy, so I now need to get the depends and optdepends for a given package, rather than the packages which depend/optionally depend on it.

Morganamilo:
I guess we had opposite idea's then. You want to mark the explicit packages, find the stuff it depends on and loop through the package list to find them. My idea was to mark the explicit packages, iterate the package list calling the Compute*() functions on them and see which ones are required by the explicit packages.

As a bonus the Compute*() functions take versioned depends into account which is nice to support even if it's not really needed.

Morganamilo:
This worked when I did some limited testing last time. New release is out and a new window is open so I'll give this some testing now and merge it. In the meantime conflicts need to be resolved.

Morganamilo:
Just reminding you that you need to resolve these conflicts. If you don't know how GitHub has their own tutorial for it. https://help.github.com/articles/resolving-a-merge-conflict-on-github/

Morganamilo:
Could you squash/rebase this? Makes things neater but not that important.

PR #277, opened 2018-03-22
 Separate gendb from install

Morganamilo:
And also added printing when a vcs source is added.


Morganamilo:
This probably also some problems that I have not tested but I would have guessed exists. Like if `foo` has a dependency `bar` but `bar` is not installed. The old gendb, being part of install() would resolve uninstalled deps. So it would pull in `bar` and add its vcs sources to the db even though the package is not installed.

Having it on its own and not using the deptree ensures it queries installed aur packages only.

PR #278, opened 2018-03-22
Add missing err check

Morganamilo:
Weirdly enough the way I found this was by messing around with gcc-go. It threw this error on me:
```
./conflicts.go:298:6: error: ‘err’ declared and not used
./conflicts.go:298:6: error: ‘err’ declared and not used
  var err error
      ^
```
I think the standard go compiler should have caught this. Compiler bug?

Also while testing:
 - gccgpo creates a yay binary of size 1.5MB (probably because it links dynamically to alpm.so)
 - go creates a yay binary of size 4.7MB
 - gccgo has an install size of 149MB and download size of 22MB
 - gccgo has an install size of 330MB and download size of 67MB

Just thought that was intresting.

### DIFF
--- a/conflicts.go
+++ b/conflicts.go
@@ -314,6 +314,11 @@ func checkForAllConflicts(dc *depCatagories) error {
 	}()
 
 	wg.Wait()
+
+	if err != nil {
+		return err
+	}
+
 	if len(innerConflicts) != 0 {
 		fmt.Println(
 			red("\nInner conflicts found:"))


Jguer:
That is very interesting, gcc-go  is less optimized than go, I'll try it out and analyse the binary to see what's stripped

Jguer:
@Morganamilo  blog.jguer.space/post/yaygcc/ I tried to analyse a bit the difference. Feel a bit silly for not remembering the go runtime immediately

Jguer:
> I though gcc was meant to be the faster one, also I read it has better performance when calling c code which we do a lot of.

True, I also thought about that, but at least from my shitty benchmarks that is not verifiable (maybe resolving a complex dependency chain gcc-go would have an advantage)

> You said running dynamically linked programs is always slower so could it just be that.

[https://stackoverflow.com/questions/4667882/is-a-statically-linked-executable-faster-than-a-dynamically-linked-executable](https://stackoverflow.com/questions/4667882/is-a-statically-linked-executable-faster-than-a-dynamically-linked-executable) has a nice, albeit layman, view on it.

> Also I realised if I dynamically link the go run time with go the yay binary is then 904kb. I doubt you'd want to make go a runtime-dependency though. save 4MB gain a 330MB dependency.

:rofl: then it would become the AUR manager by go developers for go developers only

PR #280, opened 2018-03-23
 Improve makefile

Morganamilo:
Created .go/src which symlinks to vendor/. Running `make` now builds
yay using the packages in vendor. This can still be overiden by running
`make GOPATH=$HOME/go`

Even though the makefile calls git to get the version, that can be
overiden using `make MAJORVERSION=5 MINORVERSION=400`, skipping the call to git. This
allows building yay/yay-bin using the makefile

The yay binary is now always yay instead of taking the name of the
parent directory.

Added vet and fmt checks to `make test` for more checking.

Reordered some stuff, made sure the dependency order is correct. Any
command such as `make package` can be ran at any point, even after a
`make clean` instead of having to execute the correct make commands in
order.

Changed install form using `go install` to a traditional install. The
pkgbuilds can now use `make DESTDIR=$PKGDIR install`. Added uninstall to match.

One thing this commit does not do is have proper recepies which depend
on files instead of phonies. The reason for this is that it does not
play that well with go's build system. Go is smart enough to only
recompile needed files anyway.

here's how the PKGBUILD would look after this is merged:
```sh
# Maintainer: Jguer <joaogg3@gmail.com>
pkgname=yay
pkgver=5.577
pkgrel=1
pkgdesc="Yet another yogurt. Pacman wrapper and AUR helper written in go."
arch=('i686' 'x86_64' 'armv7h' 'armv6h' 'aarch64')
url="https://github.com/Jguer/yay"
license=('GPL')
options=('!strip' '!emptydirs')
depends=(
  'sudo'
)
makedepends=(
  'git'
  'go'
)
conflicts=('yay-bin' 'yay-git')
source=("https://github.com/Morganamilo/yay/archive/v${pkgver}.tar.gz")
sha1sums=('7cd481a391b7ade75e8c8b61dbed6c7f0016fbfd')

build() {
  cd "$srcdir/$pkgname-$pkgver"
  make VERSION=$pkgver
}

package() {
  cd "$srcdir/$pkgname-$pkgver"
  make VERSION=$pkgver DESTDIR="$pkgdir" install
}
```


Jguer:
Worked great, updating `yay-git` to use it

Jguer:
`yay-git` updated. Works beautifully.

Morganamilo:
I was planning on leaving out the `VERSION=` for the -git PKGBUILD and just letting Yay figure it out itself. But after looking at it now`yay -V` gives `yay v4.505.r82.g69f4475` which matches `yay Q yay-git` which I think is a plus.

Jguer:
Yes, it's working wonderfully  _famous last words_

Issue #281, opened 2018-03-23
Zsh completion broken


Morganamilo:
Any time I try to hit tab with `yay git` I get:
```
_yay:206: parse error near `)'
```

This was probably caused by #272.

Jguer:
linters and zsh sometimes...
I'll go check it

PR #283, opened 2018-03-23
Fix zsh completions

Jguer:
Signed-off-by: Jguer <me@jguer.space>


Jguer:
Broken by passing it through `fish_indent` :innocent:

Jguer:
closes #281

PR #284, opened 2018-03-23
Do not depend on the Internet for keys_test.go

qrwteyrutiyoup:
We now mock a PGP key server to provide the keys used during the
testing, making it so that we do not need the Internet anymore for
the testing of the PGP key import feature.


Morganamilo:
This is how it should have been the first time, but I kinda forgot gpg uses the internet. Honestly I did not look into the tests that much last time.

Giving them a proper look though, why is `config.BuildDir` set? we're just testing keys right not building anything?

Also we might have other tests in future that make use of test data (or we might have some sh scripts) so I would probably move the keys to `tests/keys`. Or could you just generate the keys through gpg?

And that web server you start, I don't know how it behaves exactly but if it's possible it should reject connections from outside of the program or outside of the device.

qrwteyrutiyoup:
> This is how it should have been the first time, but I kinda forgot gpg uses the internet. Honestly I did not look into the tests that much last time.

Indeed, I realized once it had been merged. Apologies :)

> Giving them a proper look though, why is config.BuildDir set? we're just testing keys right not building anything?

Since 158b80c5, `checkPgpKeys` parses `.SRCINFO` directly, and the file is expected to be at build dir/package base/, so the test creates a dummy build dir to place dummy SRCINFO files to be parsed.

> Also we might have other tests in future that make use of test data (or we might have some sh scripts) so I would probably move the keys to tests/keys. Or could you just generate the keys through gpg?

How about testdata/keys/? `testdata` is special-cased by `go tool`, as well as directories starting with `_` and `.`, so it seemed a good option.

> And that web server you start, I don't know how it behaves exactly but if it's possible it should reject connections from outside of the program or outside of the device.

Sure thing. I will make it listen at 127.0.0.1.

Morganamilo:
I believe `checkPgpKeys()` got changed again to not parse .SRCINFOS.

And sure if its some go magic use testdata.

qrwteyrutiyoup:
Ah, right you updated the tests in 42fdd01d42627bb0bfaceeff6b865965efc3266c. config.BuildDir is not needed since then.

Morganamilo:
Yeah, you updated the tests for the srcinfo stuff then updated them again for something else, I used the magic of git to undo the srcinfo stuff so I didn't look at them too much.

Morganamilo:
Tests run about 2x faster for me now which is nice. This all ready? It looks it, but I don't get emailed for rebases/amends so I can't tell.

qrwteyrutiyoup:
Yup, it's updated already. And my bad, I thought it would email the reviewers to inform of the update. Next time I will write a comment to let you know of changes.

Morganamilo:
Don't worry about it, it looked done, just wanted to be sure.

Morganamilo:
enjoy the +4k lines to the repo lol.

PR #286, opened 2018-03-23
Suppress pacman error printing

Morganamilo:
add22f59577ce2314fb9655dd0f2e4db9d61e8f1 added error checks to all the
passToPacman commands. This makes `yay -Q nonexistantpackage` return
non 0 as it should. Annoyingly it also made yay print `exit status = n`
which is the error string from passToPacman calls. This error doesn't
add much and is quite annoying, expecially when calling pacman commands
like `-Q` or `-Si` where yay should be kind of 'hidden' and print just
like pacman does.

Now set the error string to "" for pacman commands and don't print an
error if it == "" (avoids empty line printed).

Also behave more like pacman when using `yay -Qu`.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -465,7 +465,11 @@ func passToPacman(args *arguments) error {
 
 	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
 	err := cmd.Run()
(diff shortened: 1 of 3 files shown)


Jguer:
Ok, let's focus on stabilizing for release and get back to development after

PR #294, opened 2018-03-26
Update fish completion regarding GPG options

qrwteyrutiyoup:
Added completion for --gpg and --gpgflags.
This was forgotten in #256.


Morganamilo:
Just reading through the fish config for the first time. Line 81 is a duplicate of 55 right?

qrwteyrutiyoup:
Looks like that. I ran fish in here to test and it showed `--config                             (Alternate config file)`, though. I think we can remove line 85 (or the other one).

Morganamilo:
Line 81 is wrong anyway, there is no config command, it's a config file. I assume it's just a blind copy and paste error.

I think the line 55 is from the actual pacman fish completion seems as `--config` is a pacman flag after all.

PR #296, opened 2018-03-27
update shell completion

Morganamilo:


Jguer:
If there is nothing pressing until the end of this afternoon french time, I think we can go ahead with a release.

PR #297, opened 2018-03-28
Use git clone for pkgbuild downloading

Morganamilo:
## Use git clone for pkgbuild downloading

This is a restart of #130 given that PR has not gone anywhere and @simon04 has still not answered any of my questions related to the PR such as:

>Also how does this handle migrating. If i have a package that I installed using the tarball method then switch to the git method will it handle it correctly?

I decided it was better to start over. This PR also uses some of the code from #130.

This PR in its current state is mainly for discussion on how to implement things. Currently all this PR does is replace the download method during `yay -S` to use `git`. This allows easy benchmarks, just checkout between the two versions and test for yourself.

## Testing
My testing was done against `ros-lunar-desktop-full` which on my machine wants to download 251 Aur packages.

Time taken to download all the packages assuming none exist in cache:
- Current tarball method: `3:26`
- git clone method: `1:43`

If all packages in cache and up to date then Yay will not try to redownload them but if using `--redownloadall` then the results are as follows:
- Current tarball method: `3:26` (The same download is done, the time does not change)
- git clone Method: `1:10` (it only needs to pull instead of clone)

This shows git to actually be faster than the tarball downloads.

## Features

The features I plan to implement before merging are:
- [x] Config setting for clone/tarball
- [x] Use git clone in -G for AUR
~~Use git clone in -G for ABS~~
- [x] Handle cached packages that come from tarball or git clone no matter the current config

Ideas that may be implemented:
 - [ ] `@` versioning: When installing AUR packages. For example `yay -S yay@1e01eaf` would build the package from that revision of the PKGBUILD.
 - [ ] `<>=` versioning: Pacman already supports using these during `-S` to ensure the package you want fits in the range you specify. What I purpose is slightly different. Yay will find the first latest commit that matches the version you specify. So `yay -S yay<4` will install the first PKGBUILD with a pkgver of less than 4.

## Discussion

This PR is mostly here for discussion on how to implement certain features.
### Migrating between tarball and git
My current idea of this is to only respect the config option if the cache does not already exist. If it does exist check for the .git directory. If that exists use the git method no matter the config. If it does not use tarball no matter the config.

### Updating the cache
When we have a cached package and the PKGBUILD updated how should we go about pulling changes? The tarball method is simple and just unpacks the tar into whats already there. Overwriting files that exist in the tar and in cache but leaving untracked files alone.

Doing a simple git pull could cause complications  if there are any user made changes. Currently this PR resets to head then pulls which will undo any user made changes. #130 uses a similar method where they fetch then reset to upsteam master. I do not know id there are any real differences between the two.

Alternatively we could just do a pull by itself. If there are merge conflicts it would be the user's job to resolve them.

### Pacman wrap
using `<>=` behaviour different to Pacman might be frowned upon.
maybe combine it with `@`
`yay -S yay@7655bd9`
`yay -S yay@=3.505`
`yay -S yay@<4`

### Optimisation
This implementation is very basic, there's a couple of possible tricks that could be done to speed things up or reduce disk space but I have not tested them.

`git gc` Should probably be run after every pull.

`--singlebranch` could be used to only clone master, although most AUR packages probably only have master, I am unsure if the AUR would even allow you to push a different branch.

`--depth 1` Is an obvious one but would not allow the user to then checkout specific commits if they wished to. Also possible features such as `@` and `<>=` versioning would not work with this. It might be possible to run `--depth 1` then run `--unshallow` when needed.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -253,10 +253,16 @@ func handleConfig(option, value string) bool {
 		config.AnswerUpgrade = value
 	case "noanswerupgrade":
 		config.AnswerUpgrade = ""
+	case "gitclone":
+		config.GitClone = true
+	case "nogitclone":
+		config.GitClone = false
 	case "gpgflags":
 		config.GpgFlags = value
 	case "mflags":
 		config.MFlags = value
+	case "gitflags":
+		config.GitFlags = value
 	case "builddir":
 		config.BuildDir = value
 	case "editor":
--- a/cmd.go
+++ b/cmd.go
@@ -563,3 +563,15 @@ func passToMakepkg(dir string, args ...string) (err error) {
 	}
 	return
 }
+
+func passToGit(dir string, _args ...string) (err error) {
+	gitflags := strings.Fields(config.GitFlags)
+	args := []string{"-C", dir}
+	args = append(args, gitflags...)
+	args = append(args, _args...)
+
+	cmd := exec.Command(config.GitBin, args...)
+	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
+	err = cmd.Run()
+	return
+}
--- a/download.go
+++ b/download.go
@@ -6,8 +6,27 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 )
+
+// Decide what download method to use:
+// Use the config option when the destination does not already exits
+// If .git exists in the destination uer git
+// Otherwise use a tarrball
+func shouldUseGit(path string) bool {
+	_, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		return config.GitClone
+	}
+
+	_, err = os.Stat(filepath.Join(path, ".git"))
+	if os.IsNotExist(err) {
+		return false
+	}
+
+	return true
+}
 
 func downloadFile(path string, url string) (err error) {
 	// Create the file
@@ -27,6 +46,37 @@ func downloadFile(path string, url string) (err error) {
 	// Writer the body to file
 	_, err = io.Copy(out, resp.Body)
 	return err
+}
+
+func gitDownload(url string, path string, name string) error {
+	_, err := os.Stat(filepath.Join(path, name, ".git"))
+	if os.IsNotExist(err) {
+		err = passToGit(path, "clone", url, name)
+		if err != nil {
+			return fmt.Errorf("error cloning %s", name)
+		}
+
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("error reading %s", filepath.Join(path, name, ".git"))
+	}
+
+	err = passToGit(filepath.Join(path, name), "fetch")
+	if err != nil {
+		return fmt.Errorf("error fetching %s", name)
+	}
+
+	err = passToGit(filepath.Join(path, name), "reset", "--hard", "HEAD")
+	if err != nil {
+		return fmt.Errorf("error reseting %s", name)
+	}
+
+	err = passToGit(filepath.Join(path, name), "merge", "--no-edit", "--ff")
+	if err != nil {
+		return fmt.Errorf("error merging %s", name)
+	}
+
+	return nil
 }
 
 // DownloadAndUnpack downloads url tgz and extracts to path.
@@ -129,8 +179,18 @@ func getPkgbuildsfromAUR(pkgs []string, dir string) (err error) {
 	}
 
 	for _, pkg := range aq {
-		downloadAndUnpack(baseURL+aq[0].URLPath, dir, false)
-		fmt.Println(bold(green(arrow)), bold(green("Downloaded")), bold(magenta(pkg.Name)), bold(green("from AUR")))
+		var err error
+		if shouldUseGit(filepath.Join(dir, pkg.PackageBase)) {
+			err = gitDownload(baseURL+"/"+pkg.PackageBase+".git", dir, pkg.PackageBase)
+		} else {
+			err = downloadAndUnpack(baseURL+aq[0].URLPath, dir, false)
+		}
+
+		if err != nil {
+			fmt.Println(err)
+		} else {
+			fmt.Println(bold(green(arrow)), bold(green("Downloaded")), bold(magenta(pkg.Name)), bold(green("from AUR")))
+		}
 	}
 
 	return
--- a/config.go
+++ b/config.go
@@ -42,6 +42,7 @@ type Configuration struct {
 	GpgFlags      string `json:"gpgflags"`
 	MFlags        string `json:"mflags"`
 	SortBy        string `json:"sortby"`
+	GitFlags      string `json:"gitflags"`
 	RequestSplitN int    `json:"requestsplitn"`
 	SearchMode    int    `json:"-"`
 	SortMode      int    `json:"sortmode"`
@@ -50,6 +51,7 @@ type Configuration struct {
 	NoConfirm     bool   `json:"-"`
 	Devel         bool   `json:"devel"`
 	CleanAfter    bool   `json:"cleanAfter"`
+	GitClone      bool   `json:"gitclone"`
 }
 
 var version = "3.373"
@@ -138,6 +140,7 @@ func defaultSettings(config *Configuration) {
 	config.PacmanConf = "/etc/pacman.conf"
 	config.GpgFlags = ""
 	config.MFlags = ""
+	config.GitFlags = ""
 	config.SortMode = BottomUp
 	config.SortBy = "votes"
 	config.SudoLoop = false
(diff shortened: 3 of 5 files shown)


AladW:
>Doing a simple git pull could cause complications if there are any user made changes. Currently this PR resets to head then pulls which will undo any user made changes. #130 uses a similar method where they fetch then reset to upsteam master. I do not know id there are any real differences between the two.

I used to `git stash` instead of `git reset` so the user could at least restore his changes from the command line. This caused some mysterious issues I couldn't figure out so I stuck to `git fetch` and `git merge` in the end. The main drawback imo with `git pull` is that you can't tell from the exit codes if something was actually updated or not, so it's harder to take actions accordingly. (same for `git fetch`)

Interesting idea on `<>= versioning` btw

Morganamilo:
I agree merge is probably the way to go. If a PKGBUILD has been edited, the changes should probably be kept between updates. They can always clean build and grab a fresh copy if they want to discard them.

Stashing seems like a good idea too. Too bad it caused some black magic errors.

And yeah optimisation does seem a little pointless compared to the built package sizes.

Morganamilo:
@AladW P.S. With the way the table is ordered now, what's the weight of optional entries? I believe it's considered worse than green even though I think there's nothing wrong with user choice. Maybe consider the default value instead?

optional, default yes = green
optional, default no = yellow

Jguer:
I'll try and benchmark as well, but operating only on the existing knowledge in this thread I'm beginning to flip flop on this: 

- Is there a reason to keep tar support since we already have git as an almost super dependency and git still seems faster than native code? You agreed with me at the beginning on keeping tar support @Morganamilo , any thing else I may be missing right now?

Other stuff though:
- I'll leave concurrent download of PKGBUILDs after this is merged and focus on source download for now.
- I like the versioning ideas

Morganamilo:
My mentality is always give users a choice. Don't ever just take things away without good reason *cough* desktop icons. I'm happy for git to become the default setting though.

AladW:
I think having tar as a backup and git default is the best option. Simply if git breaks (which is does in [silly ways](https://github.com/AladW/aurutils/issues/332)) you can keep using tar.

>With the way the table is ordered now, what's the weight of optional entries? I believe it's considered worse than green even though I think there's nothing wrong with user choice. Maybe consider the default value instead?
>optional, default yes = green
>optional, default no = yellow

I've proposed [a while back](https://wiki.archlinux.org/index.php/Talk:AUR_helpers#.22AUR_repo_diff.22) to use a green entry if the helper supports `git clone` and `git diff` and yellow if only `git clone`. What the default is wouldn't matter in that case.

Morganamilo:
I would have thought the point of git clone was just to use the git endpoint over the tarball one. Even though git provides diff functionality in a very convenient way diffs themselves are not a git thing and could easily be implemented in the tarball method.

P.S. thanks for rewording the wiki I'm not so good with words.

AladW:
Yes I have similar doubts, hence I didn't make any changes on that regard. 

In any event, to me the sorting is `N/A = green > yellow >> red`, so I don't see a row with red entries going above one with yellow ones. Perhaps this should be further clarified in the description.

Morganamilo:
>`N/A = green > yellow >> red`

I agree with this but having optional with a good default as green. Or have optional as yellow but have it act as green in the sorting.

>After all, a yellow entry either means that something of note isn't enabled by default

That sounds like you agree with my first point.

>like helpers that build without inspection

Honestly I think helpers that are insecure by default, even if there is a flag to secure it, should be red. Similarly the opposite should be true. If they are secure by default but have a flag to operate in insecure mode then it should be green.

While I could let it slide on git clone or whatever, being insecure is the worst offence.



Users are probably more likely to use the default settings than anything else, especially when they start using a helper.

Jguer:
User choice sounds good, as long as the fastest is the default I'm happy.

Morganamilo:
Alright I'll work on this following what I said in the original description as well as have git be the default.

Version stuff was worth mentioning but out of scope for this PR.

AladW:
>I agree with this but having optional with a good default as green. Or have optional as yellow but have it act as green in the sorting.

Oh, that's how it has always been. At least aurutils has always supported both tar and git downloads but git is the default. Yaourt also does, but it disabled them by default so it gets a yellow one. So it already matches what you proposed in https://github.com/Jguer/yay/pull/297#issuecomment-376981224.

>Honestly I think helpers that are insecure by default, even if there is a flag to secure it, should be red. 

Well, the description of the secure column (much like the pacman wrap column) says "by default". I don't think you'll get much opposition if you wanted to discuss those helpers to get a red entry too though.

Morganamilo:
Oh I didn't realise it was already like this. Aurutils just said yes instead of optional so I assumed it was git only. Maybe that should be explained better.

AladW:
How about https://wiki.archlinux.org/index.php?title=AUR_helpers&diff=515201&oldid=515185

Morganamilo:
I think it should be extended slightly to explain that "Yes" does not mean the opposite feature is unavailable. e.g. Yes to git clone does not imply no to tarball.

AladW:
I'm not sure how to word this or if it makes sense to add this to the general note, but perhaps we can reword the description of the "git clone" column where it's most relevant. (I was never fond of the arbitrary "uses git because tar is deprecated" since there's no harm in supporting tar and upstream hasn't mentioned they would actually remove this functionality.)

For now I've just removed the reference to tar and focused on git. https://wiki.archlinux.org/index.php?title=AUR_helpers&diff=515208&oldid=515201

Morganamilo:
I wasn't sure how to word it myself, why I just gave an example instead. I think it's good as it is now, nice job.

Morganamilo:
Git clone for abs stuff is a little more annoying but still doable. I'm not sure weather it's worth the effort to implement it. Could just use tarball only for abs stuff.

asp should be used for most abs stuff anyway, Yay's implementation is rather basic. I've though about extending `-G` to support stuff like dependency resolving. If that is done maybe aps should be used as a dependency.

Morganamilo:
Doing something like `git clone https://git.archlinux.org/svntogit/packages.git --single-branch --branch packages/git git` will clone an ABS package but the directory structure does not match the normal `-G` way. The pkgbuild ends up being located in `git/repos/extra-x86_64` which of course varies depending on the repo.

It's possible to do the clone then move and delete files to get the standard layout, but that then removes the main benefits of cloning from git. Pulling wouldn't work, diffs wouldn't work unless you compare against the old pkgbuild's location.

For this reason I think it is best to no support git on ABS packages, leave it up to asp.

download.go
```diff
@@ -29,6 +48,32 @@ func downloadFile(path string, url string) (err error) {
...
+func gitDownload(url string, path string, name string) error {
+	_, err := os.Stat(filepath.Join(path, name, ".git"))
+	if os.IsNotExist(err) {
+		err = passToGit(path, "clone", url, name)
+		if err != nil {
+			return fmt.Errorf("error cloning %s", name)
+		}
+
+		return nil
+	} else if err != nil {
```
qrwteyrutiyoup: No need for this `else` (and the next one), since the blocks end with a `return`.

cmd.go
```diff
@@ -526,3 +526,15 @@ func passToMakepkg(dir string, args ...string) (err error) {
 	}
 	return
 }
+
+func passToGit(dir string, _args ...string) (err error) {
```
qrwteyrutiyoup: Maybe pick a different name for this `_args` variable? I skimmed quickly through the code and didn't find other variables named similarly (starting with an underline).

Morganamilo: I would rather not have to have two separate variables at all like in the other passTo* functions but it didn't work like that. If you've got a better method or name feel free to suggest one.

Morganamilo:
I think this is about ready, you can debate weather or not we should show the git output but that's about it. You wanna give this a final test yourself @Jguer before merging?

Jguer:
Tested, looking good, apparently I murdered the merge a bit though

Morganamilo:
Well I have no idea what you did but it looks like you merged. Close this?

AlexWayfer:
You can now change `No` at `Git clone` column for `yay` there, I think: https://wiki.archlinux.org/index.php/AUR_helpers

Morganamilo:
Nope gotta wait for a new release first.

AlexWayfer:
> Nope gotta wait for a new release first.

Yeah, sorry.

Morganamilo:
Speaking of, the release is now. @AladW If you would be so kind.

AladW:
[Done](https://wiki.archlinux.org/index.php?title=AUR_helpers&diff=518852&oldid=517897)

Morganamilo:
Thanks!

Jguer:
All `yay` packages updated, bumped major version as a marker for major change (git clone introduction)

AladW:
Somewhat offtopic, but I've added a discussion to lessen restriction on what an "active" helper is, as well as move out some entries from the main comparison table. Feel free to make comments. 

https://wiki.archlinux.org/index.php/Talk:AUR_helpers#Effectiveness_of_the_.22inactive.22_table

PR #301, opened 2018-03-29
Save provides seperatly from pkgnames

### LINKED ISSUE #279: Better dependency resolving
#### Affected Version
yay v4.505

#### Issue
For example you want to install foo-bin package which has dependency bar. But you want to install bar-bin dependency instead of compile bar from sources. Far as I know here isn't option to choose should yay use regular packages, bin or git. You might think you can just do <code>yay -S bar-bin foo-bin</code> to install binaries, but you can't do that because yay doesn't check that bar-bin provides bar. So yay will end up installing bar-bin bar and foo.(of course that doesn't work because bar-bin conflicts with bar)

So it would be nice to have option what would let you choose should yay use bin or git packages if there is any available. And fix to the problem what I mentioned earlier.

#### Steps to reproduce

Here is real world example:
1. <code>yay -S mingw-w64-binutils-bin mingw-w64-headers-bin mingw-w64-crt-bin mingw-w64-winpthreads-bin mingw-w64-gcc-bin</code>

Morganamilo:
Now if `foo` wants to install `bar`

if you do `yay -S foo bar-git` Yay will detect bar is satisfied and not
try to install `bar` as well as `bar-git`.

fixes #279

### DIFF
--- a/dependencies.go
+++ b/dependencies.go
@@ -15,6 +15,7 @@ type depTree struct {
 	Aur       map[string]*rpc.Pkg
 	Missing   stringSet
 	Groups    stringSet
+	Provides  map[string]string
 }
 
 type depCatagories struct {
@@ -31,6 +32,7 @@ func makeDepTree() *depTree {
 		make(map[string]*rpc.Pkg),
 		make(stringSet),
 		make(stringSet),
+		make(map[string]string),
 	}
 
 	return &dt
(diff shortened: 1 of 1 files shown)


Morganamilo:
All the dep stuff is getting a little bloated with all the extra checks bolted on. I think I'm going to look through and simplify it at some point.

Issue #305, opened 2018-03-31
[Feature] Do a full system upgrade (-Syu) if no flags are given


Trollwut:
Hi folks!

Man, this AUR helper looks fine! I love the "packages not to update" on an update and some more features.

When running `yay` without any flags, I get this output:

    $ yay
    no operation specified (use -h for help)

Any plans on just doing a `yay -Syu` of no flags are given?

Morganamilo:
Are you a user of `pac` by any chance? I was and I did quite enjoy that feature. I thought about adding it when I first started using Yay then just kinda forgot about it. It shouldn't be hard to add if people want this.

Trollwut:
Yes I am :) I did some color fixing for outdated packages for that project

`yay` would be the first helper that would drive me away from abandoned `pacaur` :)

Morganamilo:
Thanks, I did try to clone a lot of the features, batch install mainly, so even though Yay is it's own thing I do want it to be a better version of Pacaur. If you know how to program and have worked on AUR stuff before you could always go implement this yourself ;)

Trollwut:
Never used Go language and I'm not the best programmer, but maybe I'll try it if I have a bit freetime

DL6AKU:
Tried for a fix. First time in Go, please be gentle ;)

Jguer:
This is a nice idea, I'll start using it myself, although it might not be the favourite for some users. I don't there will be complaints but let's see if any user prefers this an option until the next release. 

PS: thinking about it a config option to set the default OP for when no OP is passed also sounds nice (and defaulting to -Syu)

PR #308, opened 2018-03-31
Default action (running yay without any args) is now to -Syu.

DL6AKU:
Fixes / implements #305

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -526,34 +526,33 @@ func (parser *arguments) parseCommandLine() (err error) {
 	usedNext := false
 
 	if len(args) < 1 {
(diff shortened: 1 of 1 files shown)


Morganamilo:
Thanks for the PR, looks about perfect as is.

It looks like you have some weird miss match between your git credentials and GitHub account though, just thought I'd point it out. Makes it look like some one else authored and you're just committing. (unless that is what is happening)

DL6AKU:
> It looks like you have some weird miss match between your git credentials and GitHub account though, just thought I'd point it out. Makes it look like some one else authored and you're just committing. (unless that is what is happening)

Yes, I need to fix this. Thanks.

PR #329, opened 2018-04-05
Highlight diff between old and new versions better

AlexWayfer:
Split by dots, pluses, dashes, etc., not only `Version` and `Pkgrel`.

Don't make new version different bold.

Inspired by [`pikaur`](https://github.com/actionless/pikaur).

-----

Before:

image

After:

image

### DIFF
--- a/upgrade.go
+++ b/upgrade.go
@@ -55,11 +55,11 @@ func (u upSlice) Less(i, j int) bool {
 
 	if found != nil {
 		return less
-	} else {
-		iRunes := []rune(u[i].Repository)
-		jRunes := []rune(u[j].Repository)
-		return lessRunes(iRunes, jRunes)
 	}
+
+	iRunes := []rune(u[i].Repository)
+	jRunes := []rune(u[j].Repository)
+	return lessRunes(iRunes, jRunes)
 
 }
 
--- a/upgrade.go
+++ b/upgrade.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"unicode"
 
 	alpm "github.com/jguer/go-alpm"
 	rpc "github.com/mikkeloscar/aur"
@@ -74,13 +75,50 @@ func getVersionDiff(oldVersion, newversion string) (left, right string) {
 	}
 
 	if errOld == nil && errNew == nil {
-		if old.Version == new.Version {
-			left = string(old.Version) + "-" + red(string(old.Pkgrel))
-			right = string(new.Version) + "-" + green(string(new.Pkgrel))
-		} else {
-			left = red(string(old.Version)) + "-" + string(old.Pkgrel)
-			right = bold(green(string(new.Version))) + "-" + string(new.Pkgrel)
+		oldVersion := old.String()
+		newVersion := new.String()
+
+		if oldVersion == newVersion {
+			return oldVersion, newVersion
 		}
+
+		diffPosition := 0
+
+		checkWords := func(str string, index int, words ...string) bool {
+			for _, word := range words {
+				wordLength := len(word)
+				nextIndex := index + 1
+				if (index < len(str)-wordLength) &&
+					(str[nextIndex:(nextIndex+wordLength)] == word) {
+					return true
+				}
+			}
+			return false
+		}
+
+		for index, char := range oldVersion {
+			charIsSpecial := !(unicode.IsLetter(char) || unicode.IsNumber(char))
+
+			if (index >= len(newVersion)) || (char != rune(newVersion[index])) {
+				if charIsSpecial {
+					diffPosition = index
+				}
+				break
+			}
+
+			if charIsSpecial ||
+				(((index == len(oldVersion)-1) || (index == len(newVersion)-1)) &&
+					((len(oldVersion) != len(newVersion)) ||
+						(oldVersion[index] == newVersion[index]))) ||
+				checkWords(oldVersion, index, "rc", "pre", "alpha", "beta") {
+				diffPosition = index + 1
+			}
+		}
+
+		samePart := oldVersion[0:diffPosition]
+
+		left = samePart + red(oldVersion[diffPosition:len(oldVersion)])
+		right = samePart + green(newVersion[diffPosition:len(newVersion)])
 	}
 
 	return
--- a/print.go
+++ b/print.go
@@ -362,60 +362,52 @@ func formatTime(i int) string {
 	return fmt.Sprintf("%d/%02d/%02d", t.Year(), int(t.Month()), t.Day())
 }
 
-func red(in string) string {
+const (
+	redCode     = "\x1b[31m"
+	greenCode   = "\x1b[32m"
+	yellowCode  = "\x1b[33m"
+	blueCode    = "\x1b[34m"
+	magentaCode = "\x1b[35m"
+	cyanCode    = "\x1b[36m"
+	boldCode    = "\x1b[1m"
+
+	resetCode = "\x1b[0m"
+)
+
+func stylize(startCode, in string) string {
 	if useColor {
-		return "\x1b[31m" + in + "\x1b[0m"
+		return startCode + in + resetCode
 	}
 
 	return in
+}
+
+func red(in string) string {
+	return stylize(redCode, in)
 }
 
 func green(in string) string {
-	if useColor {
-		return "\x1b[32m" + in + "\x1b[0m"
-	}
-
-	return in
+	return stylize(greenCode, in)
 }
 
 func yellow(in string) string {
-	if useColor {
-		return "\x1b[33m" + in + "\x1b[0m"
-	}
-
-	return in
+	return stylize(yellowCode, in)
 }
 
 func blue(in string) string {
-	if useColor {
-		return "\x1b[34m" + in + "\x1b[0m"
-	}
-
-	return in
+	return stylize(blueCode, in)
 }
 
 func cyan(in string) string {
-	if useColor {
-		return "\x1b[36m" + in + "\x1b[0m"
-	}
-
-	return in
+	return stylize(cyanCode, in)
 }
 
 func magenta(in string) string {
-	if useColor {
-		return "\x1b[35m" + in + "\x1b[0m"
-	}
-
-	return in
+	return stylize(magentaCode, in)
 }
 
 func bold(in string) string {
-	if useColor {
-		return "\x1b[1m" + in + "\x1b[0m"
-	}
-
-	return in
+	return stylize(boldCode, in)
 }
 
 // Colours text using a hashing algorithm. The same text will always produce the
--- a/upgrade_test.go
+++ b/upgrade_test.go
@@ -1,0 +1,56 @@
+package main
+
+import "testing"
+
+func TestGetVersionDiff(t *testing.T) {
+	useColor = true
+
+	type versionPair struct {
+		Old string
+		New string
+	}
+
+	in := []versionPair{
+		{"1-1", "1-1"},
+		{"1-1", "2-1"},
+		{"2-1", "1-1"},
+		{"1-1", "1-2"},
+		{"1-2", "1-1"},
+		{"1.2.3-1", "1.2.4-1"},
+		{"1.8rc1+6+g0f377f94-1", "1.8rc1+1+g7e949283-1"},
+		{"1.8rc1+6+g0f377f94-1", "1.8rc2+1+g7e949283-1"},
+		{"1.8rc2", "1.9rc1"},
+		{"2.99.917+812+g75795523-1", "2.99.917+823+gd9bf46e4-1"},
+		{"1.2.9-1", "1.2.10-1"},
+		{"1.2.10-1", "1.2.9-1"},
+		{"1.2-1", "1.2.1-1"},
+		{"1.2.1-1", "1.2-1"},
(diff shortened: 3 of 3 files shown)


AlexWayfer:
It's my first experience with Go lang. :sweat_smile: 

Feel free to comment and criticize the code.

Morganamilo:
It certainly looks nice, but whats that long regex string meant to do exactly? There's probably a simpler way to do this a simpler way through strings.FieldsFunc or something.

The current implementation looks a little inconsistent. glib spits between 1.8 and rc1 but open ssh does not split between 6 and p. Also re2 is split for some reason even though their all numbers.

It's definitely a nice feature though once all the kinks have been worn out.

AlexWayfer:
> but open ssh does not split between 6 and p.

It does: https://repl.it/repls/StickyGrandHarddrives

There is just `6` also changed to `7`.

> Also re2 is split for some reason even though their all numbers.

For year separation, there is comment in the code: https://github.com/Jguer/yay/pull/329/files#diff-135dc004c4305f5b7fe8d80752bd129aR67

It's like major version, I think. Easier to see the changes for human.

-----

> but whats that long regex string meant to do exactly?

It's not so long for me. :sweat_smile: 

> There's probably a simpler way to do this a simpler way through strings.FieldsFunc or something.

Maybe. But it's harder to separate year in version with `strings.FieldsFunc`, I think.

AlexWayfer:
Replaced `[\.\-\+]?` with `([\.\-\+]|$)` in Regexp for greater accuracy.

Jguer:
I notice some weird separations on versions but my biggest issue is the inclusion of the `regexp` package for only this small small operation

```
du -h yay yay-master 
4.8M    yay
4.7M    yay-master
```
I won't go into a full crusade against bloating the binary with packages and making everything in house ( #103 ) but this doesn't seem to justify it. I agree with @Morganamilo on if we change this to use the `strings` package.

About the code, thank you for trying your hand at go for `yay`, I've reviewed and it is well written, I would caution the use of indexes after the regex without cheking the length of the result first 
` if oldVersion[inds[0]:inds[1]] != newVersion[inds[0]:inds[1]]  `

Also, I would check the return for `nil`, just in case make it not crash but continue.
 > FindAllStringIndex is the 'All' version of FindStringIndex; it returns a slice of all successive matches of the expression, as defined by the 'All' description in the package comment. A return value of nil indicates no match.

Morganamilo:
I see what you mean with ssh and re2 but that just leads me to think `8rc1` should be considered one block.

Also I don't see much use in the year separation, also looking at the code it just it just looks for any 4 digits right? So it could falsely trigger on any number?

Also this seems to fail to separate on `~`:
image

I think it would just be simpler to split on anything not alphanumeric.

AlexWayfer:
> I notice some weird separations on versions but my biggest issue is the inclusion of the regexp package for only this small small operation

It's including just once, and it may be useful later. And there is 28 KB, not 100:

```
$ du /usr/bin/yay ./yay
4712	/usr/bin/yay
4884	./yay
```

And it's only + 3.65%.

> I agree with @Morganamilo on if we change this to use the strings package.

Okay. :disappointed: 

> About the code, thank you for trying your hand at go for yay, I've reviewed and it is well written

Thank you.

> I would caution the use of indexes after the regex without cheking the length of the result first
if oldVersion[inds[0]:inds[1]] != newVersion[inds[0]:inds[1]]

Why?

> Also, I would check the return for nil, just in case make it not crash but continue.

I think I don't need for this if you're asking me to remove `regexp` package. But OK, I'll be more careful with `nil` in Go.

-----

> I see what you mean with ssh and re2 but that just leads me to think 8rc1 should be considered one block.

What? No! `1.8rc1` — "the first release candidate of version 1.8".

> Also I don't see much use in the year separation

Maybe it's superfluous, yes.

> looking at the code it just it just looks for any 4 digits right?

Yes.

> So it could falsely trigger on any number?

Yes, but I think (hope) there will no another case with more 4 digits besides years.

> Also this seems to fail to separate on ~:

Thank you.

> I think it would just be simpler to split on anything not alphanumeric.

Maybe.

AlexWayfer:
> I agree with @Morganamilo on if we change this to use the strings package.

How I'll join version parts?

[Example](https://repl.it/repls/StickyGrandHarddrives): 

```go
package main

import(
  "fmt"
  "strings"
  "unicode"
)

func main() {
  splitVersion := func(char rune) bool {
    return !unicode.IsLetter(char) && !unicode.IsNumber(char)
  }
	
  fmt.Printf("%v", strings.FieldsFunc("1.8rc2+1+g7e949283-1", splitVersion))
}
```

Output: `[1 8rc2 1 g7e949283 1] `

Where will I get the dropped dividers (`.`, `+`, `-`)?

AlexWayfer:
> For implementation I was thinking something like.

Thank you, I made something like this.

> I can see rc1 being a single block but 8rc1 shouldn't be its own block that's for sure. You could split when a number turns into a letter but not vice versa. I don't know if that will look weird in other circumstances though. For example git hashes which should be one block.

I think it's very dangerous, especially for `git` hashes, yes. I made with `"rc"` parsing. This condition can be extended with other combinations (`beta`, `alpha`, `pre`, etc.) in the future.

-----

OK, I literally made the function for these keywords.

Morganamilo:
That "or one version ends" thing is important because what if the new version is shorter than the old one?
image

There's also this case:
image

the `4+` shouldn't be highlighted right?

Apart from that it appears to work perfectly.

AlexWayfer:
Thank you, @Morganamilo!

1. I moved `break` before "`diffPosition` increasing" for cases with `1.2.3-1` → `1.2.33-1` (`"-" != "3"`, so it breaks).
2. I've added check for one of version ending into "`diffPosition` increasing" for don't highlight last part.
3. I've added check `index >= len(newVersion)` before `char != rune(newVersion[index])` (that was moved to top) for cases like `avahi` from you, but downgrade.

Morganamilo:
I wrote a few tests for some of the cases, currently a few of them fail but shouldn't be too hard to fix. Feel free to extend the test cases:

```go
func TestGetVersionDiff(t *testing.T) {
	useColor = true

	type versionPair struct {
		Old string
		New string
	}

	in := []versionPair{
		{"1-1", "1-1"},
		{"1-1", "2-1"},
		{"1-1", "1-2"},
		{"1.2.3-1", "1.2.4-1"},
		{"1.8rc1+6+g0f377f94-1", "1.8rc2+1+g7e949283-1"},
		{"2.99.917+812+g75795523-1", "2.99.917+823+gd9bf46e4-1"},
		{"1.2.9-1", "1.2.10-1"},
		{"1.2.10-1", "1.2.9-1"},
		{"1.2-1", "1.2.1-1"},
		{"0.7-4", "0.7+4+gd8d8c67-1"},
	}

	out := []versionPair{
		{"1-1", "1-1"},
		{red("1-1"), green("2-1")},
		{"1-" + red("1"), "1-" + green("2")},
		{"1.2." + red("3-1"), "1.2." + green("4-1")},
		{"1.8" + red("rc1+6+g0f377f94-1"), "1.8" + green("rc2+1+g7e949283-1")},
		{"2.99.917+" + red("812+g75795523-1"), "2.99.917+" + green("823+gd9bf46e4-1")},
		{"1.2." + red("9-1"), "1.2." + green("10-1")},
		{"1.2." + red("10-1"), "1.2." + green("9-1")},
		{"1.2" + red("-1"), "1.2" + green(".2-1")},
		{"0.7" + red("-4"), "0.7" + green("+4+gd8d8c67-1")},
	}

	for i, pair := range in {
		o, n := getVersionDiff(pair.Old, pair.New)

		if o != out[i].Old || n != out[i].New {
			t.Errorf("Test %d failed for update: (%s => %s) expected (%s => %s) got (%s => %s)", i+1, in[i].Old, in[i].New, out[i].Old, out[i].New, o, n)
		}
	}
}
```

upgrade.go
```diff
@@ -74,13 +75,43 @@ func getVersionDiff(oldVersion, newversion string) (left, right string) {
 	}
 
 	if errOld == nil && errNew == nil {
-		if old.Version == new.Version {
-			left = string(old.Version) + "-" + red(string(old.Pkgrel))
-			right = string(new.Version) + "-" + green(string(new.Pkgrel))
-		} else {
-			left = red(string(old.Version)) + "-" + string(old.Pkgrel)
-			right = bold(green(string(new.Version))) + "-" + string(new.Pkgrel)
+		oldVersion := string(old.Version) + "-" + string(old.Pkgrel)
```
Morganamilo: The version strings are passed in, you then convert them to a CompleteVersion and back to a string. Won't this always equal the same thing that was passed in? If there is some reason to do this you should be using old.String()

upgrade.go
```diff
@@ -74,13 +75,43 @@ func getVersionDiff(oldVersion, newversion string) (left, right string) {
...
+			}
+
+			if (
+				!(unicode.IsLetter(char) || unicode.IsNumber(char)) ||
+					(
+						(
+							(index == len(oldVersion) - 1) ||
+							(index == len(newVersion) - 1)) &&
+						(len(oldVersion) != len(newVersion))) ||
+					checkWord(oldVersion, index, "rc") ||
```
Morganamilo: Why not just make checkWord variadic? Would be much cleaner to do `checkWord(oldversion, index, "rc", "pre", "alpha", "beta")`

Morganamilo:
Just add that function to upgrade_test.go and run `make test`. You need to declare the package and add the testing import, just look at the other _test.go files and copy it.

AlexWayfer:
Thank you, but I mean it'll be good to add this manual for other people.

Morganamilo:
This is part of the go tool chain and not really specific to this project, you could find documentation for this somewhere on the go website.

The only thing special we have is `make test` which just sets the gopath to the packaged vendors, and runs some tests using the go toolchain. 

if you wanted to run go test directly you can either set the go path yourself to use the packaged dependencies `GOPATH=$PWD/.go go test` or you can use go get to fetch the dependencies `go get -d github.com/Jguer/yay && cd ~/go/src/github.com/jguer/yay && go test`. But like I said that's standard go stuff and is documented elsewhere.

AlexWayfer:
@Morganamilo, OK! I just offered.

I somehow ran tests and passed them, thank you.

Also I was forced to make some style changes for `gofmt` in other file (865fe7c).

And then I got something like this:

```
Test 1 failed for update: ("1-1" => "1-1") expected ("1-1" => "1-1") got ("1-1\x1b[31m\x1b[0m" => "1-1\x1b[32m\x1b[0m")
```

(empty colored strings)

So, I've added a check for `len(in) > 0` in color functions, but it's kind of stupid, so I've created common function `stylize("red", in)` with all checks inside (d714c63).

upgrade.go
AlexWayfer: Done, thank you!

upgrade.go
AlexWayfer: Done, thank you!

Morganamilo:
For the first test cause you could have just changed the expected result to be an empty coloured string. Or in `getVersionDiff()` just have a check at the start:
```
if old == new {
    return old, new
}
```

If you're going to go with the `stylize()` stuff though I'm fine with that. You should define them as consts though instead of creating a new map every function call.

Morganamilo:
You know everything at compile time just pass the variable:
```go
func red(in string) string {
    if useColor {	
        return stylize(redCode, in)
    }
}
```

AlexWayfer:
OK, I got it. Thank you. Made.

Morganamilo:
It looks good now I think, sadly I've finally ran an update so that makes testing a little harder. It does pass the test cases though so should be fine. To be nit picky you could make `resetCode` a constant then have `startCode + in + resetCode` but that's all I really think of looking at the code.

Thanks for the PR, the version diffs look really nice now. I'll merge this once @Jguer makes a release.

AlexWayfer:
> sadly I've finally ran an update so that makes testing a little harder

There is my update:

image

> To be nit picky you could make resetCode a constant then have startCode + in + resetCode

OK, done, thank you.

> Thanks for the PR, the version diffs look really nice now. 

Thank you, I did my best. :grin:

AlexWayfer:
Yay! :tada: 

Thank you.

Zesko:
The highlight of diff version does not show me when using `yay v11.3.0`

Has the feature been removed?

AlexWayfer:
> The highlight of diff version does not show me when using `yay v11.3.0`
> 
> Has the feature been removed?

Em… no, as I see:

image

Maybe the problem in your terminal application or shell?

Zesko:
Before that I used `yay -Syu` that does not show it.

 but yay with two options `--upgrademenu --combinedupgrade` shows it.

See: #1827

AlexWayfer:
I guess yay's highlight doesn't support `VerbosePkgLists` pacman's option, but can be wrong.

Zesko:
Yes, because `VerbosePkgLists` text format is designed by pacman, it does not follow "default" highlight of Yay. 
but theoretically, Yay could customize it when using the same output of pacman.

AlexWayfer:
Yes, so I think you could remain #1827 open (and rename and detail it?).

Zesko:
I opened it, but this issue #1827 shows the wrong label "Bug", because I did not know `--combinedupgrade` before. 

I can not change the label "Bug" to "feature".

PR #339, opened 2018-04-08
Document fcf0ee34d17e20dc7cb99e8b6547f71c5ab522e4

Morganamilo:


doc/yay.8
```diff
@@ -34,6 +36,9 @@ Perform yay specific print operations\&.
 Downloads PKGBUILD from ABS or AUR\&.
 .RE
 .PP
+If no arguments are provided 'yay -Syu' will be preformed\&.
```
Jguer: *performed

Morganamilo: Isn't that what it says?

Jguer: preform
verb
past tense: preformed; past participle: preformed

    form (something) beforehand.
    "a preformed pool"

perform
verb
past tense: performed; past participle: performed

    1.
    carry out, accomplish, or fulfil (an action, task, or function).

Jguer: I don't really mind typos specially in code, but I know it will always become a pull request down the line

Morganamilo: Oh right they looked so similar I couldn't see the difference. Thanks.

Morganamilo: I do hate typos but the longer I work on a feature the more I tend to 'just get it done'. This wasn't a typo though I just didn't know the difference between the words.

Jguer: No problem, pulling out dictionary references still seemed like the easiest way to show it was in the spelling.

Thinking of maybe doing a release today, have any critical pull request to make?

Morganamilo: Critical? No. Shell completion is a thing though, but the --help is already done so it's basically copy and paste.

--sortby should have completion for the supported types, just like how --color will cycle between always, never and auto.

PR #342, opened 2018-04-09
Localizing Yay take two

Morganamilo:
This is a working example of localization. This mostly focuses on the framework, no actual translation has been done.

The only string that has been translated is the warning to not run Yay as root. Even then it's not a real translation, just a dud for testing.

It should be noted that Yay looks for .mo files in its install prefix so running `make && ./yay` will not work. Instead you should install Yay `make PREFIX=/usr/local install` then use the installed binary.

image

TODO:
- [ ] Get a fully working template that includes all strings
- [x] Ignore things that should not be translated such as `http.Get()`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ _cgo_defun.c
 _cgo_gotypes.go
 _cgo_export.*
 
+*.mo
+*.pot
+
 *.exe
 *.test
 *.prof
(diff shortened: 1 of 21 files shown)


Jguer:
https://github.com/Jguer/yay/blob/localize-template/locale/pt_PT.po

I think these should be all the strings except the ones escaped with literal marks.

LothaireCliquennois:
Well...
What will happen with that PR ?

Jguer:
reworked and merged in https://github.com/Jguer/yay/commit/732f1a5412cd196433b960164c03e9a63d59c4bd

Issue #347, opened 2018-04-11
yay -Yc offers to remove recently updated packages


DX37:
Yay 5.657 sometimes offers me to delete some critical packages, when I run `yay -Yc`.

For example, on my Raspberry Pi 3B with Arch Linux, when I'm updating `linux-raspberrypi`, which obviously contains kernel, after it running the command will offer me to delete that package with kernel.

Morganamilo:
can you confirm its marked as explicit? `yay -Qi linux-raspberrypi`

DX37:
@Morganamilo you talking about it?
`Install Reason  : Installed as a dependency for another package`

DX37:
I also installed some gvfs packages on PC Arch Linux by myself, but yay offered me do delete them too, until I removed them and installed again.

Morganamilo:
Looks like some stuff is marked wrong: 
```
pacman -Qqg base base-devel | sudo pacman -D --asexp linux-raspberrypi linux-raspberrypi-headers raspberrypi-bootloader raspberrypi-bootloader-x raspberrypi-firmware -
```

should mark a bunch of stuff as explicitly installed and catch most things if they are wrong. If you notice anything else that should be explicit mark them too.

DX37:
@Morganamilo Thanks for help. I'll close issue now.

Jguer:
I just marked everything I had on `-Yc` as explicit since I usually clean unneeded dependencies.
 `yay -D --asexplicit $(pacman -Qtdq)`

I really need to transition `yay` to container based testing to avoid this in the future. It's going on my TODO list for when I have a bit more time.

Morganamilo:
I have dabbled a bit in this myself. I was trying to get some sort of testing working. I have a pkgbuild that sets up a local AUR website with a predictable database. This is good for testing resolving Dependencies via the rpc but not much else.

Ideally We should be able to set up a bunch of dud packages that basically do nothing apart from have the correct meta data for testing. Combine this with a copy of the current Pacman database and we have a full never changing set of packages. Then we simply have all the packages in cache so Yay does not need to fetch them.

This gives us free reign to control everything and has no need for an internet connection which allows us to have control of everything. Know the exact state of what should be installed, what should be explicit, ect.

Then use mikkeloscar's Arch docker with Travis and we have automated testing.

DX37:
> I really need to transition yay to container based testing to avoid this in the future. It's going on my TODO list for when I have a bit more time.

[...]

PR #348, opened 2018-04-11
Don't mark repo upgrades as deps during sysupgrade

Morganamilo:
Before `yay -Syu` called `pacman -Sy <pkgs to upgrade>`
We then later switched to it calling `pacman -Syu` this lead to yay
seeing no targets to when it was upgrading a bunch of packages it
assumed they must be deps. Correct this by adding repo packages to the
targets list.

Also ensure we dont mark packages as dependencies if they are already
installed. For example we install `foo` which requires `bar>5` but we
only have `bar=4` installed. In this case installing `foo` will pull bar
in as a dependency but it should not be marked as such because it
already exists.

### DIFF
--- a/config.go
+++ b/config.go
@@ -187,7 +187,7 @@ func editor() (string, []string) {
 		fallthrough
 	default:
 		fmt.Println()
-		fmt.Println(bold(red(arrow) + " Warning:"), cyan("$EDITOR"), "is not set")
+		fmt.Println(bold(red(arrow)+" Warning:"), cyan("$EDITOR"), "is not set")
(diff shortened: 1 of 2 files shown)


Morganamilo:
I found this thanks to #347, This is a pretty bad bug which sucks but at least It's found now. You should test this a bit yourself then I think we should make a new release right away.

Also while looking into this problem. The second point about already installed packages. This bug seems to be present in Pacaur, Yaourt and Trizen. Pikaur failed to build for me so I don't know if it will be effected.

Jguer:
I have no way of testing this right now but I'm going to make a release either way since this is a very big regression

PR #366, opened 2018-04-23
Update man page and tweak usage

Morganamilo:


Morganamilo:
@Jguer  I've been away for a couple of days so I've kind of forgot what exactly needs doing. This is the only thing I can think of, after this are we ready for release?

Jguer:
I caught a completion file bug the other day, so I'm holding it off a bit while `yay-git` is in daily use to see if something else pops up, my plan would be to tag by Thursday at most

Jguer:
After that I think we can start working seriously on localisation

Morganamilo:
Yeah there's just a bunch of prep to do really.

- Sort out how exactly we want to print errors
- reorder all the strings to be in a nice printf format
- localise the continueTask() and number menu stuff

AladW:
Just wanted to say that you guys are doing an admirable job with the release engineering. aurutils is the exact opposite - I just commit to master at random hoping things don't break too badly and release every other year.

Morganamilo:
Thanks, means a lot coming from TU. I always imagined it as the opposite, we just pick random release dates and you're waiting for everything to be perfect.

Issue #364, opened 2018-04-19
[Feature] Compress package list


tomswartz07:
#### Affected Version
`yay v5.675`

#### Issue
Out of the box, when running `-Syu`, the output is very screen intensive. 
Is it possible to add an option to compress the list of upgraded packages, similar to pacaur?

This way, the upgrades are visible on one screen's worth of text and scrolling through the output is not necessary, if there are more updates than lines of terminal.

#### Steps to reproduce
```sh
yay -Syu
 99 core/btrfs-progs                                        4.16-1 -> 4.16-2
 98 core/ca-certificates-mozilla                             3.36-1 -> 3.36.1-1
 97 core/libldap                                          2.4.45-4 -> 2.4.46-1
.....
```

Ideally, a config option could enable output like:
```sh
yay -Syu

Packages (98) core:
              attica-5.45.0-1  babl-0.1.46-1  btrfs-progs-4.16-2  bubblewrap-0.2.1-1  
              extra:
              virtualbox-5.2.10-1  virtualbox-guest-iso-5.2.10-1 ......
```

Morganamilo:
We already have a pacaur style install list after the upgrade menu. Simply hiding the menu when `--answerupgrade` is set should do it.

tomswartz07:
@Morganamilo ah, good catch- I did not know about that feature. 

I noticed a few minor issues with that in the manual (typos, etc), I'll toss up a PR to clarify that, if you want?

Morganamilo:
Yes of course, contributors are always welcome.

PR #367, opened 2018-04-23
Fix typos

tomswartz07:
- Clarify issues found in #364
- Fix a few typos

### DIFF
--- a/doc/yay.8
+++ b/doc/yay.8
@@ -104,7 +104,7 @@ Print current yay configuration\&.
 .PP
 \fB\-n \-\-numberupgrades\fR
 .RS 4
-Print number of packages that need to be updated\&. Note this does not preform
+Print number of packages that need to be updated\&. Note this does not perform
 a database refresh\&. Run \fByay -Sy\fR Before this for an up to date result\&.
 .RE
 .PP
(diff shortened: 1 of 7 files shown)


doc/yay.8
```diff
@@ -201,22 +201,22 @@ Sort AUR results by a specific field during search\&.
...
 the same when parsed\&.
 .RE
 .PP
 \fB\-\-answerupgrade\fR <Repo|^Repo|None|...>
 .RS 4
-Set a predetermined answer for the upgrade menu question\&. This answer
-will be used instead of reading from standard input but will be treaded exactly
-the same\&.
+Set a predetermined answer for the upgrade menu question\&. Selects which packages
+groups or repos to select for updades\&. This answer will be used instead of
```
Morganamilo: This answers what packages to not upgrade. Also you can't enter groups unless you mean `1-3 `which I would call ranges to not confuse with groups like `base` and `base-devel`.

Morganamilo:
There's quite a bit here, thanks for fixing all this. Glad these are all minor typos and there was nothing horribly wrong.

doc/yay.8
tomswartz07: Ah, apologies- got the logic back-to-front. Updated commit incoming.

tomswartz07:
Glad to help- the program `codespell` helped quite a bit as well :)

doc/yay.8
Morganamilo: There's still the bit about groups, I would either leave it out or change it to package ranges.

Morganamilo:
I've used aspell in the past, didn't think to check the completion files though. codespell seems really neat, thanks for making me aware of this.

doc/yay.8
Morganamilo: Looks good now thanks.

PR #371, opened 2018-04-27
Use megacheck to lint project

Jguer:
Fixes
```
keys_test.go:76:24: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (SA1012)
keys_test.go:141:24: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (SA1012)
parser.go:223:3: this value of arg is never used (SA4006)
parser_test.go:22:6: identical expressions on the left and right side of the '!=' operator (SA4000)
print.go:343:33: this value of err is never used (SA4006)
download.go:24:2: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (S1008)
parser.go:622:29: should use make(intRanges, 0) instead (S1019)
parser.go:623:29: should use make(intRanges, 0) instead (S1019)
query.go:356:6: should write state := safePackages[pkg.Name()] instead of state, _ := safePackages[pkg.Name()] (S1005)
upgrade.go:196:35: should use make([]alpm.Package, 0) instead (S1019)
upgrade.go:197:29: should use make([]string, 0) instead (S1019)
keys_test.go:39:6: func newSplitPkg is unused (U1000)
vcs.go:208:6: func inStore is unused (U1000)
```

@Morganamilo  `parser.go:223:3: this value of arg is never used (SA4006)` didn't go deep into this as I mostly close my eyes to parser issues (and laziness on this account may also contribute)  but could this be a problem?

### DIFF
--- a/keys_test.go
+++ b/keys_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -34,10 +35,6 @@ func init() {
 
 func newPkg(basename string) *rpc.Pkg {
 	return &rpc.Pkg{Name: basename, PackageBase: basename}
-}
-
-func newSplitPkg(basename, name string) *rpc.Pkg {
-	return &rpc.Pkg{Name: name, PackageBase: basename}
 }
 
 func getPgpKey(key string) string {
@@ -73,7 +70,7 @@ func TestImportKeys(t *testing.T) {
 	config.GpgFlags = fmt.Sprintf("--homedir %s --keyserver 127.0.0.1", keyringDir)
 
 	server := startPgpKeyServer()
-	defer server.Shutdown(nil)
+	defer server.Shutdown(context.TODO())
 
 	casetests := []struct {
 		keys      []string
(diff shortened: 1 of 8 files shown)


Morganamilo:
The value is never used because it is part of the naked return. If arg was truly unused it would be impossible to do `yay --dbpath ~/testdb -S foo` which I do quite often.

I've never heard of megacheck but maybe this should be reported to them.

Morganamilo:
Actually thinking about it the second arg will overwrite the first one. So I guess the lintier is right after all.

keys_test.go
```diff
@@ -73,7 +70,7 @@ func TestImportKeys(t *testing.T) {
 	config.GpgFlags = fmt.Sprintf("--homedir %s --keyserver 127.0.0.1", keyringDir)
 
 	server := startPgpKeyServer()
-	defer server.Shutdown(nil)
+	defer server.Shutdown(context.TODO())
```
Morganamilo: Seems like it should have been `server` passed here all along. @qrwteyrutiyoup any reason it was nil?

qrwteyrutiyoup: Sorry, I am currently with (very) limited Internet access. No good reason it was nil; I must have missed it. Thanks for the fix!

PR #383, opened 2018-05-04
Use ISO 8601 date format (yyyy-mm-dd)

simon04:


Morganamilo:
I like it a lot more this way, thanks.

Issue #384, opened 2018-05-05
Request: show Arch package news stuff when updating


samvdkris:
I saw a comment on Reddit about someone who made an alias to run `curl -s https://www.archlinux.org/feeds/news/ | xmllint --xpath //item/title\ \|\ //item/pubDate /dev/stdin | sed -r -e "s:<title>([^<]*?)</title><pubDate>([^<]*?)</pubDate>:\2\t\1\n:g"` before running `yay -Syu` to see if there are any packages that need manual intervention or have any problems. I thought it would be cool if yay automatically did this when updating. I'd make a pull request, but I know absolutely nothing about Go and even making a script that just executed `ls` failed for some reason. If anyone could add this (or another implementation of something similar), that would be really cool.

Also, I just wanted to say this AUR helper is really cool. Thanks to everyone who contributed to it! :)

heyitswither:
Probably best to add this as an option, maybe `--news`

samvdkris:
Yeah, I guess some people would just get annoyed. Or maybe just make it togglable in `~/.conf/yay/config.json` (already exists for other settings).

heyitswither:
Since you brought up the config file, it might be cool to add "hooks," commands that will be run on certain events. So instead of hard-coding this feed command, users can put whatever they want to run on updates.

Morganamilo:
On mobile wrong button.

I do want to get some comment and news fetching into yay, probably as an extension to -P.

No plant to implement right now but I like the idea.

PeachIceTea:
-P seems perfect for that and --news would work as well. However -Pn is already taken for a different function, maybe -Pw would do?

Morganamilo:
I've thought about -W/--web also, this might deserve its own namespace.

PeachIceTea:
It does seem like something that people might wanna access more quickly. Also it would be nice because the short and the long form would match. It is however a bit ambiguous.

Morganamilo:
What makes it ambiguous?

PeachIceTea:
I just think that it is not as obvious as others, e.g. --sync or --upgrade. --web isn't immediately clear in what it does.

ashleybone:
An alternative to adding this feature is using pacmatic, a pacman wrapper in the community repo that does this (it also prints mailing list chatter).  pacmatic has an environmental variable, pacman_program, that lets you specify an alternative program to wrap.  I've been using it with yay for a while.

Morganamilo:
[...]

PR #385, opened 2018-05-06
Get arch package news using --print --news

PeachIceTea:
Implements #384 

Not sure about using -w but -n is already taken.


Morganamilo:
That was way simpler to implement than I expected. As I mentioned I'm away and on mobile right now so I'm unable do review it.

Thanks for the PR, I'll take a look when I can, unless Jguer gets to it first.

PeachIceTea:
#384 was asking for the news to be printed during install. Would be easy enough to print the messages of the last 30 days and have a setting to disable it, but there hasn't been any feedback on that yet.

PR #386, opened 2018-05-06
Add the feed to sync 

PeachIceTea:
#384 also asked for the feed to be printed during upgrades. The sync feed is limited to printing the last 30 days by default but it can be adjusted by users as well as completely disabled.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -320,7 +320,7 @@ func handlePrint() (err error) {
 	case cmdArgs.existsArg("u", "upgrades"):
 		err = printUpdateList(cmdArgs)
 	case cmdArgs.existsArg("w", "news"):
-		err = printNewsFeed()
+		err = printNewsFeed(false)
 	case cmdArgs.existsArg("c", "complete"):
 		switch {
 		case cmdArgs.existsArg("f", "fish"):
--- a/config.go
+++ b/config.go
@@ -48,12 +48,14 @@ type Configuration struct {
 	RequestSplitN int    `json:"requestsplitn"`
 	SearchMode    int    `json:"-"`
 	SortMode      int    `json:"sortmode"`
+	FeedLimit     int    `json:"feedlimit"`
 	SudoLoop      bool   `json:"sudoloop"`
 	TimeUpdate    bool   `json:"timeupdate"`
 	NoConfirm     bool   `json:"-"`
 	Devel         bool   `json:"devel"`
 	CleanAfter    bool   `json:"cleanAfter"`
 	GitClone      bool   `json:"gitclone"`
+	SyncFeed      bool   `json:"syncfeed"`
 }
 
 var version = "5.688"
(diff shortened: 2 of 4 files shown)


Morganamilo:
Instead of the last thirty days how about just showing the news since the last update?

Morganamilo:
Also it seems you forgot that `-w` is `--downloadonly`.

PeachIceTea:
Is the timestamp of the last successful install saved somewhere already? 

Also, I don't think the overlap with -w is an issue. --download only is a sync options, --news is a print option.

Morganamilo:
>Is the timestamp of the last successful install saved somewhere already?

The install time of all packages is tracked, just iterate over them all and find the newest one.

>Also, I don't think the overlap with -w is an issue. --download only is a sync options, --news is a print option.

I skimmed it on mobile and I thought this PR made it so adding -w during a -S command also prints the news. My bad.

To be honest I'm not that keep on printing the news during sync, It's very easy to just do `yuay -Pw && yay`.

Morganamilo:
>The install time of all packages is tracked, just iterate over them all and find the newest one.

Thinking over this more, maybe instead of the last package install we could get the time of the last refresh (`-y`). Also to be safe add something like a week onto that. That way even if you do refresh/install you can still easily see the news for a week.

Another idea is to always show the last news entry no matter what. That way you can check the news see its still on that same topic and know there's nothing new.

PeachIceTea:
I wonder if we would want to assume that people always go through yay (which is how I think we would track the last -y unless we have a way of pacman telling us as well).  Its why I chose to use the 30 days in the first place.

On the other hand I suppose it does no real harm if people use pacman sometimes since they would simply get more news rather than missing some.

> Another idea is to always show the last news entry no matter what. 

I think that wouldn't be a good idea because people would get used to seeing one item listed and might miss if that changes. The plan is still to only print the title of the item, right?

Morganamilo:
>I wonder if we would want to assume that people always go through yay

Never assume that, the package install times are tracked so I thought the refresh times might be too. No idea though, I have not looked into it. Either way it's not a big deal I'm happy to go off of install times.

>The plan is still to only print the title of the item, right?

I guess that would probably work best, then if they want to read it they can check it manually.

Personally I don't want this to be attached to sync I would rather use `yay -Pw` as mentioned. But if others want it and there's a config option I don't care too much.

Jguer:
> I guess that would probably work best, then if they want to read it they can check it manually.
Personally I don't want this to be attached to sync I would rather use yay -Pw as mentioned. But if others want it and there's a config option I don't care too much.

I don't even have to write. This ^ and off by default

Morganamilo:
@PeachIceTea You continuing with this? Agree/disagree with the stuff discussed? I don't mind making an alternative PR if you're no longer interested.

Jguer:
Closing this for inactivity

PR #387, opened 2018-05-07
 Print the descriptions of news item

Morganamilo:
Print the full descriptions of each news item.

The description is formatted as html, basic parsing is done to display
it properly. -q/--quiet can be used to display title only.

Respect bottomup for -Pw

//cc @PeachIceTea 

@Jguer You asked for the news struct to be collapsed on the last PR, here it is extended again so that the Item struct could have its own `Print()` function.

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -7,16 +7,8 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"html"
 )
-
-var htmlEscapeSequences = map[string]rune{
-	"quot": '"',
-	"apos": '\'',
-	"amp":  '&',
-	"lt":   '<',
-	"gt":   '>',
-	"nbsp": '\u008a',
-}
 
 // A basic set implementation for strings.
 // This is used a lot so it deserves its own type.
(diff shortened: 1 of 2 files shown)


PeachIceTea:
I didn't notice the bottomup option. Good catch.

However printing them all is kind of a mess. Maybe we should make it interactive? Only printing the title and then allowing people to select which they want to read?

Jguer:
Now that methods are implemented it makes sense. Don't you want to move news related code to `news.go` though?

Jguer:
Fair enough for me

PeachIceTea:
I think there is a benefit to just seeing a short list of the news headlines. Maybe we can introduce a flag like -Pwa to list all news with descriptions and have the interactivity by default?

PeachIceTea:
Although thinking about it again. If we display the headlines during install then it would be fine if --news printed the description as well.

Although -Pwa to print all items might be nice still.

Morganamilo:
>Although -Pwa to print all items might be nice still.

Like I mentioned I was thinking of passing news twice for this `-Pww` (I am reserving -a for --aur if that ever happens)

Right now it prints everything by default and adding `-q` will print all headlines.

I'm going to merge this as is because your open PR has code for figuring out which headlines to print. So rather than re write that you can just merge this in.

PR #396, opened 2018-05-08
Print warnings about newer local packages then remote versions

AlexWayfer:
Inspired by [`pakku`](https://github.com/kitsunyan/pakku)

Before:

image

After:

image

(I'm not sure about format, colors and condition, but this info is very helpful).

### DIFF
--- a/upgrade.go
+++ b/upgrade.go
@@ -138,6 +138,8 @@ func upList(warnings *aurWarnings) (aurUp upSlice, repoUp upSlice, err error) {
 	var aurErr error
 	var develErr error
 
+	pkgdata := make(map[string]*rpc.Pkg)
+
 	fmt.Println(bold(cyan("::") + bold(" Searching databases for updates...")))
 	wg.Add(1)
 	go func() {
@@ -148,7 +150,7 @@ func upList(warnings *aurWarnings) (aurUp upSlice, repoUp upSlice, err error) {
 	fmt.Println(bold(cyan("::") + bold(" Searching AUR for updates...")))
 	wg.Add(1)
 	go func() {
-		aurUp, aurErr = upAUR(remote, remoteNames, warnings)
+		aurUp, aurErr = upAUR(remote, remoteNames, pkgdata, warnings)
 		wg.Done()
 	}()
 
@@ -162,6 +164,8 @@ func upList(warnings *aurWarnings) (aurUp upSlice, repoUp upSlice, err error) {
 	}
 
 	wg.Wait()
+
+	printLocalNewerThanAUR(remote, pkgdata)
 
 	errs := make([]string, 0)
 	for _, e := range []error{repoErr, aurErr, develErr} {
@@ -228,9 +232,7 @@ func upDevel(remote []alpm.Package) (toUpgrade upSlice, err error) {
 
 	for _, pkg := range toUpdate {
 		if pkg.ShouldIgnore() {
-			left, right := getVersionDiff(pkg.Version(), "latest-commit")
-			fmt.Print(yellow(bold(smallArrow)))
-			fmt.Printf(" Ignoring package upgrade %s (%s => %s)\n", cyan(pkg.Name()), left, right)
+			printIgnoringPackage(pkg, "latest-commit")
 		} else {
 			toUpgrade = append(toUpgrade, upgrade{pkg.Name(), "devel", pkg.Version(), "latest-commit"})
 		}
@@ -242,14 +244,16 @@ func upDevel(remote []alpm.Package) (toUpgrade upSlice, err error) {
 
 // upAUR gathers foreign packages and checks if they have new versions.
 // Output: Upgrade type package list.
-func upAUR(remote []alpm.Package, remoteNames []string, warnings *aurWarnings) (upSlice, error) {
+func upAUR(
+	remote []alpm.Package, remoteNames []string,
+	pkgdata map[string]*rpc.Pkg, warnings *aurWarnings) (upSlice, error) {
+
 	toUpgrade := make(upSlice, 0)
 	_pkgdata, err := aurInfo(remoteNames, warnings)
 	if err != nil {
 		return nil, err
 	}
 
-	pkgdata := make(map[string]*rpc.Pkg)
 	for _, pkg := range _pkgdata {
 		pkgdata[pkg.Name] = pkg
 	}
@@ -263,9 +267,7 @@ func upAUR(remote []alpm.Package, remoteNames []string, warnings *aurWarnings) (
 		if (config.TimeUpdate && (int64(aurPkg.LastModified) > pkg.BuildDate().Unix())) ||
 			(alpm.VerCmp(pkg.Version(), aurPkg.Version) < 0) {
 			if pkg.ShouldIgnore() {
-				left, right := getVersionDiff(pkg.Version(), aurPkg.Version)
-				fmt.Print(yellow(bold(smallArrow)))
-				fmt.Printf(" Ignoring package upgrade: %s (%s => %s)\n", cyan(pkg.Name()), left, right)
+				printIgnoringPackage(pkg, aurPkg.Version)
 			} else {
 				toUpgrade = append(toUpgrade, upgrade{aurPkg.Name, "aur", pkg.Version(), aurPkg.Version})
 			}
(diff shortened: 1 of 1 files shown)


Morganamilo:
You should probably use smallArrow over the warning text, Also it's worth noting -git packages always appear newer than the version in the AUR, and It's quite annoying:
image

AlexWayfer:
OK. I'll test and change it on the computer with non-git packages (on my current there are only git packages). Tomorrow, I guess.

AlexWayfer:
> Also it's worth noting -git packages always appear newer than the version in the AUR, and It's quite annoying

How can I check this? Just by name suffix?

Morganamilo:
That's all you can do really. See `isDevelName()`.

AlexWayfer:
> See isDevelName().

Thank you. I think I should ignore packages with all these suffixes from this function: `svn`, `nightly`, etc.

AlexWayfer:
> You should probably use smallArrow over the warning text

Used (yellow and bold).

> `isDevelName()`

Used.

Morganamilo:
The pkgname should be coloured the same as the ignore upgrade text. Also now we have a mix of messaged that start with text then say the package name and we have some messages that do the opposite. One of them should be swapped around.

AlexWayfer:
> I think color of "Out Of Date AUR Packages" is better.

Made:
image

Morganamilo:
>I think color of "Out Of Date AUR Packages" is better.

They're both the same colour.

AlexWayfer:
> They're both the same colour.

Nope: `cyan` in "Out of date" and `magenta` in number of upgrade.

Oh, sorry, I misread "ignore upgrade text".

OK, I'll try to do this.

(I can't figure out how to ignore multiple packages).

Morganamilo:
```
--ignore <package>
           Directs pacman to ignore upgrades of package even if there is one available. Multiple
           packages can be specified by separating them with a comma.
```

AlexWayfer:
Thank you. Early I got:

```
./yay -S --help

...

--ignore <pkg>   ignore a package upgrade (can be used more than once)
```

I thought the option can be used more than once. :man_shrugging: 

Separation with space didn't work.

Morganamilo:
Funnily enough `--cachedir` can be used more than once, just like how you were trying to use ignore. Why? I have no idea?

>--ignore <pkg>   ignore a package upgrade (can be used more than once)

This does seem misleading. The message comes from Pacman though, maybe check if there's already something open about it?

Morganamilo:
>Should I make this messages in separate cycle, or mix messages about ignoring and local newer is fine?

I don't really mind really, you could have it do newer warnings then ignore warnings but still in the same block. But that would probably require some refactoring.

AlexWayfer:
> What I mean is. The ignore package text starts with some text then says the package name. While the local is newer starts with the package name then has some text. Its inconsistent and one should be swapped.

Done. I also changed `=>` to `->` in ignore packages, how it's in local newer packages and upgrade packages.

> The message comes from Pacman though

Yeah, that's true.

> maybe check if there's already something open about it?

I didn't found.

> I don't really mind really, you could have it do newer warnings then ignore warnings but still in the same block. But that would probably require some refactoring.

OK, I'll try. For now it looks like this:

image

AlexWayfer:
> I don't really mind really, you could have it do newer warnings then ignore warnings but still in the same block. But that would probably require some refactoring.

I think it's done now:

image

AlexWayfer:
(Probably, we can print this lists as upgrade table, but I'm not sure that this is good idea)

Morganamilo:
>Done. I also changed => to -> in ignore packages, how it's in local newer packages and upgrade packages.

Funny thing is I changed it from `->` to `=>` because `=>` is what Pacman uses. We do use `->` in the menu and honestly I find `->` to look better anyway so I don't mind changing it back.

I'm happy with how it is now though (apart from the syntax errorr ;)).

AlexWayfer:
> because => is what Pacman uses

Where?!

> We do use -> in the menu and honestly I find -> to look better anyway so I don't mind changing it back.

OK… I can type any thing. I just want to see the same thing between sections.

> apart from the syntax errorr ;)

Oh… Sorry. I still not used to Go, and plugin with `gofmt` doesn't work in this case.

Morganamilo:
>Where?!

Here:
image

Morganamilo:
image


Looks like you didn't change all of the `->`. Whatever you pick make sure to change them all. Maybe extract them all into a function.

AlexWayfer:
> Here:

There is another format at all (`warning:`).

> Looks like you didn't change all of the ->. Whatever you pick make sure to change them all. Maybe extract them all into a function.

Thank you. Changed.

But I got mix of "repo ignored", "newer aur", "aur ignored":

image

:disappointed: 

I wanted to take off printing "newer" before "repo ignored", but these parts of code in `go` (routines?) and I'm stuck.

---

Oh, I can print this after all. Made this:

image

I think it's better than before.

AlexWayfer:
> There is another format at all (warning:).

Oh. Or we can change format from `-> Text: package (versions)` to `warning: package: text (versions)` in both blocks (about ignoring and local newer), and make this as in `pacman`.

As you wish.

Morganamilo:
Nah, I like the arrow format more than warning text.

When it comes to `-> package: text (versions)` vs `-> text: package (versions)` I did say above that I don't mind, as long as it's all consistent.

Morganamilo:
You still haven't pushed the changes of the last image.

AlexWayfer:
> You still haven't pushed the changes of the last image.

Are you sure?

I moved print-function call to the end of AUR qurery: https://github.com/Jguer/yay/pull/396/files#diff-135dc004c4305f5b7fe8d80752bd129aR271

Morganamilo:
I haven't read the code yet so I don't know how it works but but this is what I'm getting:
image

Morganamilo:
I looked at your code. Basically you only wait for the AUR update checks to finish, you forgot about the devel stuff.

Try this:
```patch
From 047dea3421b4bdd6b7918a05b31dc80d850ce1f8 Mon Sep 17 00:00:00 2001
From: morganamilo <morganamilo@gmail.com>
Date: Fri, 11 May 2018 21:02:36 +0100
Subject: [PATCH] Print local newer than text after ignore text

Signed-off-by: morganamilo <morganamilo@gmail.com>
---
 upgrade.go | 15 ++++++++-------
 1 file changed, 8 insertions(+), 7 deletions(-)

diff --git a/upgrade.go b/upgrade.go
index 6b6f445..c1555b4 100644
--- a/upgrade.go
+++ b/upgrade.go
@@ -138,6 +138,8 @@ func upList(warnings *aurWarnings) (aurUp upSlice, repoUp upSlice, err error) {
 	var aurErr error
 	var develErr error
 
+	var pkgdata = make(map[string]*rpc.Pkg)
+
 	fmt.Println(bold(cyan("::") + bold(" Searching databases for updates...")))
 	wg.Add(1)
 	go func() {
@@ -148,7 +150,7 @@ func upList(warnings *aurWarnings) (aurUp upSlice, repoUp upSlice, err error) {
 	fmt.Println(bold(cyan("::") + bold(" Searching AUR for updates...")))
 	wg.Add(1)
 	go func() {
-		aurUp, aurErr = upAUR(remote, remoteNames, warnings)
+		aurUp, aurErr = upAUR(remote, remoteNames, pkgdata, warnings)
 		wg.Done()
 	}()
 
@@ -163,6 +165,8 @@ func upList(warnings *aurWarnings) (aurUp upSlice, repoUp upSlice, err error) {
 
 	wg.Wait()
 
+	printLocalNewerThanAUR(remote, pkgdata)
+
 	errs := make([]string, 0)
 	for _, e := range []error{repoErr, aurErr, develErr} {
 		if e != nil {
@@ -240,15 +244,14 @@ func upDevel(remote []alpm.Package) (toUpgrade upSlice, err error) {
 
 // upAUR gathers foreign packages and checks if they have new versions.
 // Output: Upgrade type package list.
-func upAUR(remote []alpm.Package, remoteNames []string, warnings *aurWarnings) (upSlice, error) {
+func upAUR(remote []alpm.Package, remoteNames []string, pkgdata map[string]*rpc.Pkg, warnings *aurWarnings) (upSlice, error) {
 	toUpgrade := make(upSlice, 0)
-	_pkgdata, err := aurInfo(remoteNames, warnings)
+	info, err := aurInfo(remoteNames, warnings)
 	if err != nil {
 		return nil, err
 	}
 
-	pkgdata := make(map[string]*rpc.Pkg)
-	for _, pkg := range _pkgdata {
+	for _, pkg := range info {
 		pkgdata[pkg.Name] = pkg
 	}
 
@@ -268,8 +271,6 @@ func upAUR(remote []alpm.Package, remoteNames []string, warnings *aurWarnings) (
 		}
 	}
 
-	printLocalNewerThanAUR(remote, pkgdata)
-
 	return toUpgrade, nil
 }
 
-- 
2.17.0
```

AlexWayfer:
> Basically you only wait for the AUR update checks to finish, you forgot about the devel stuff.

That's true — I don't have installed devel stuff.

> Try this:

Thank you, made like this.

Jguer:
One down,  just another and I think we can call this stable.

PR #397, opened 2018-05-08
Add EditorConfig file

AlexWayfer:
More info [here](http://editorconfig.org/)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+indent_style = tab
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true


.editorconfig
```diff
@@ -0,0 +1,8 @@
+root = true
+
+[*]
+indent_style = tab
```
AlexWayfer: There is also can be `indent_size = N` (for alignment, [article about this](https://dmitryfrank.com/articles/indent_with_tabs_align_with_spaces)), but I don't know the preferred size (and I didn't see alignment of code in this project at all).

Jguer:
I write using spaces and indent 4. But even then, this is half useless because all of our code is supposed to be formatted with gofmt, what we should force is gofmt before committing

AlexWayfer:
OK… I have some troubles with my environment: Atom Editor + EditorConfig plugin (which is responsible for indentation, newlines and trailing spaces) + disabled Whitespace plugin (core default plugin for these things). I switched from Workspace to EditorConfig because every project has its own style and convention about previously mentioned things. But `gofmt` is something new for me (I never seen such automatic thing before).

I'll think about it. You can close this PR if you want.

> I write using spaces and indent 4.

But code has tabs… and we can not specify indentation size, or even style. And config will have information about trailing spaces, newlines, etc.

Jguer:
https://atom.io/packages/go-plus seems nice to integrate gofmt into the workflow.

AlexWayfer:
> https://atom.io/packages/go-plus seems nice to integrate gofmt into the workflow.

Yeah, seems working, thank you.

PR #407, opened 2018-05-11
Tweak readme, add install and contrib sections

### LINKED ISSUE #406: Installation instructions
I know that this might be kinda controversial in arch community
but why not just having dumb as fuck installation instructions in the readme?

noobs would welcome that, and non-noobs appreciate not having to go to aur page, looking for the exact pkgbuild url, and then go to arch wiki to go dig in to how exactly one installs package manually because the last time they did it was like year ago...

- `sudo pacman -S wget`
- `wget https://aur.archlinux.org/cgit/aur.git/snapshot/yay.tar.gz`
- `tar -xzf yay.tar.gz`
- `cd yay`
- `makepkg -si`

Morganamilo:
Closes #406 

There was also some mention of adding some contributing instructions somewhere but It's buried in the comments of an issue/pr somewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Yet another Yogurt - An AUR Helper written in Go
 There's a point in everyone's life when you feel the need to write an AUR helper because there are only about 20 of them.
 So say hi to 20+1.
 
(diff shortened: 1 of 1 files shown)


Jguer:
Seems nice, I would just move `## Examples of custom operations` before Installation but otherwise it's pretty nice

PR #418, opened 2018-05-14
Fixes #417

### LINKED ISSUE #417: -G does not work for directories with certain characters
#### Affected Version
v6.727.r39.g250d6c6

#### Issue
Trying to use `-G` with ABS packages in a path containing characters such as `(` or whitespace does not work. Looking at the code it's because the tar command:

https://github.com/Jguer/yay/blob/250d6c61f7f563f1e1b5701d62c34b3d9619c56e/download.go#L97-L99

Calls the shell and passes the arguments that way. This causes these characters to be interpreted by the shell. We could either use the tar command directly or do it pragmatically through go like how we unpack `packages.gz`.

I don't really want to touch this right now though because there's all sorts of weirdness in that command I don't really understand.

Jguer:
- Ports `-G` to current state of imports
- Moves ABS logic away from a simple download and unpack utility
- Use `$CacheHome` to store downloaded archives and delete it in the end
- Uniform messages in `-G`
- Extra checks to avoid similar problems or directory clobbering

### DIFF
--- a/download.go
+++ b/download.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 )
 
 // Decide what download method to use:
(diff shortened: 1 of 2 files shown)


Jguer:
@Morganamilo I tested it against `/tmp/ \(eke\ e/` and it's working. I would have preferred to use something else instead of syscalling `mv` but I was getting `fs` errors from golang by using `os.Rename`.

I believe it removes all "cryptocode" from `-G`. 
I also tested it with `yay -G gimp catch2 boost shfmt neovim-symlinks` for a composite action and it also works. I would merge it immediately but since you brought it up, I leave it for your review

Morganamilo:
Seems to work, worth merging as a quick fix but I would like to see the transition to https://golang.org/pkg/archive/tar/ + https://golang.org/pkg/compress/gzip/.

I'll take a look to see if I can get rid of `mv` in the mean time if you don't plan to implement the above point anytime soon. (I remember it was nofix in the projects page)

Jguer:
It was a `nofix` for the amount of code needed for such a "simple" output.  To transition to the native libraries I would like to do a performance+binary size analysis first.

PR #420, opened 2018-05-16
Only show new news when calling -Pw

Morganamilo:
New news is defined as news that is newer than the install date of the
newest package on the users system.

All news can be shown using -Pww

### DIFF
--- a/print.go
+++ b/print.go
@@ -425,14 +425,20 @@ type Item struct {
 	Creator     string `xml:"dc:creator"`
 }
 
-func (item Item) Print() error {
+func (item Item) Print(installTime time.Time) {
+	var fd string
 	date, err := time.Parse(time.RFC1123Z, item.PubDate)
 
 	if err != nil {
-		return err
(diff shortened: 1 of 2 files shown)


Morganamilo:
#386 Is taking a while and I would like to get a new release out soon. This is what I wanted the news printing to be like. This should finalise how the news works and its syntax for the release. Adding it to the install messages can always be added at a later point.

I did not add the week buffer that I mentioned before, I don't know if there's much of a need for it. Another thing I'm thinking about is using the package build time over the install time. To get packages with a newer build time you would have to `-Sy` and install. At this point you've probably checked the news.

Jguer:
True, I also want to close this cycle as soon as possible, so I'm moving #386 to `v7x`  (and probably, depending on the amount of work from rework localization to `v7xx`)

Morganamilo:
Localisation is not that much work I think. I just want most of the features finished so the translations don't have to be updated again and again.

PR #423, opened 2018-05-17
Document news and use build date over install date

Morganamilo:
Consider this the final tweak to news this cycle.


Jguer:
Fair enough, going to finish fish completions tomorrow and I believe we can package tomorrow

Morganamilo:
Looks like #385 was nice enough to update the completions. Although after #420 `-q/--quiet` is a valid option now so maybe we should add that too.

Jguer:
I had time critical stuff to deliver yesterday, today I have more time to release and deal with any fallout

PR #401, opened 2018-05-10
Dependency system rewrite

Morganamilo:
The current dependency system has gotten a bit bloated. After it was originally written versioned deps, db/name support and other little things were added on. This ended up with a lot oh hacked together code to work with the existing system and of course bugs.

This rewrite aims to be mostly the same, with a few improvements and much cleaner code. Currently it supports mostly everything that it did before.

### Improvements
Hopefully the cleaner code and more thought out version checking will fix stuff like https://github.com/Jguer/yay/issues/319#issuecomment-386725934

Fixes #368 

Slightly less over the top conflict messages

Better missing package text:
image

AUR provide solver #18: 
image

TODO:

- [x] Finish up some stuff in the code
- [ ] Comments
- [ ] Tests
- [ ] Decide on how the AUR warnings should work
- [ ] handle mixed aur and repo provides?
- [ ] Better handling of makedeps? #398 

Fixes #422 #416 #368 #18

PR #425, opened 2018-05-20
all: remove unused parameters and results

mvdan:
Found with mvdan.cc/unparam.


Jguer:
I won't take this commit right now as #401 is already touching a lot of the affected functions and it was written to:

> The current dependency system has gotten a bit bloated. After it was originally written versioned deps, db/name support and other little things were added on. This ended up with a lot of hacked together code to work with the existing system and of course bugs.

PR #447, opened 2018-05-31
Implement diffs for pkgbuild viewing.

### LINKED ISSUE #131: [Feature] Diff view of PKGBUILD on upgrade
I have begun to use yay as my primary AUR utility since pacaur is now unmaintained. 
This tool has been a great replacement!

One thing that I do miss is the ability to see the difference of a PKGBUILD file on an upgrade.

As suggested, I normally check out the PKGBUILD file to ensure that the download sources are good and to see that everything is generally in order. Currently, yay allows you to open an editor on installation/upgrade to review or modify the PKGBUILD file before installation. This is great for installations. On upgrades, I found it helpful for the diff between the new and old PKGBUILD to be shown so I could quickly see what was changed and not have to review the entire PKGBUILD file.

I could see this added two different ways, but open to discussion/thought

1. Default behavior that on upgrade it asks if you want to view the diff output (y/N). If N is chosen, the installation starts as it does now. If Y is chosen, the diff output is shown and an additional prompt is shown asking to edit the file.

2. Have this as an additional parameter that can be passed when an upgrade is called (--diff) or something like that.

With number 1, I know it would add an additional step for people that want to see the diff output, but for other users it would not add any additional steps. Also, I think the people looking at the diff output would think this was an acceptable trade off.

With number 2, The people that cared for this option could either use the parameter when calling yay or easily make an alias.

Thank you for your time and work that you do for the community.

Morganamilo:
diff viewing can be toggled via --[no]showdiffs. When enabled diffs will
be shown for packages between the current HEAD and upstream's HEAD.
Packages downloaded via tarballs will just have their pkgbuilds printed.

git diff is used to show diffs. Therefore the pager for diffs can be
shown via the PAGER and GIT_PAGER enviroment variables.

Fixes #131 after all this time.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -304,6 +304,10 @@ func handleConfig(option, value string) bool {
 		config.PGPFetch = true
 	case "nopgpfetch":
 		config.PGPFetch = false
+	case "showdiffs":
+		config.ShowDiffs = true
+	case "noshowdiffs":
+		config.ShowDiffs = false
 	case "a", "aur":
 		mode = ModeAUR
 	case "repo":
@@ -603,7 +607,6 @@ func passToMakepkgCapture(dir string, args ...string) (string, string, error) {
 	args = append(args, mflags...)
 
 	cmd := exec.Command(config.MakepkgBin, args...)
-	cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
 	cmd.Dir = dir
 	cmd.Stdout = &outbuf
 	cmd.Stderr = &errbuf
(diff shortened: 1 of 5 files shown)


Morganamilo:
This follows the default `git diff` + `less` behaviour. So small diffs are printed directly, while larger diffs stay in less interactively. This annoys me a little. What do you think about that? Worth coming up with a solution for? @Jguer

Jguer:
I'll check it out. Would you prefer to always go to less or that they should be always printed directly?

Morganamilo:
I would like it to always go to less, but then piping manually is more annoying and less wont display colour without the `-R` flag.

Morganamilo:
Git uses `-F/--quit-if-one-screen by default` for less. If you export your own less flags `LESS=RSX` then that overrides this and less will always show. I'll consider this problem solved.

There's another problem with empty diffs but that should be an easy fix.

Morganamilo:
I changed the UX a little, I'd say I'm happy with it. Setting `LESS=RX` makes it perfect. Maybe that should go into the FAQ.

Jguer:
@Morganamilo after I check the diff of an upgrade package how would I go about editing that PKGBUILD?

Morganamilo:
You'd have to open the file in your editor yourself or use --noshowdiffs. This is why I had the diffs always show the full file paths. I don't know how important editing is to you. I always saw it as viewing but you also happened to be in an editor. 

I thought It would be tedious to have edit and diff prompts at the same time. Maybe you can think of something that works.

Jguer:
Requires more work:
We could menu it, `NONE (default) , DIFF, EDIT`

Less work:
If you don't care for the diff then you also don't care to edit. So only ask to `EDIT` if you `DIFFED` it

For me the less work option works

PR #475, opened 2018-06-09
Fix failing tests

alanjjenkins:
There were several calls to fmt.Errorf in setPaths where the returned error was not
being used. This was indicated by ```make test``` as shown here:

```
make test
gofmt -l *.go
go vet
./main.go:16: result of fmt.Errorf call not used
./main.go:21: result of fmt.Errorf call not used
./main.go:25: result of fmt.Errorf call not used
./main.go:30: result of fmt.Errorf call not used
./main.go:35: result of fmt.Errorf call not used
./main.go:39: result of fmt.Errorf call not used
make: *** [Makefile:43: test] Error 2
```

With these changes the tests now all pass with no errors.

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,30 +13,30 @@ import (
 func setPaths() error {
 	if _configHome, set := os.LookupEnv("XDG_CONFIG_HOME"); set {
 		if _configHome == "" {
-			fmt.Errorf("XDG_CONFIG_HOME set but empty")
+			return fmt.Errorf("XDG_CONFIG_HOME set but empty")
 		}
 		configHome = filepath.Join(_configHome, "yay")
 	} else if _configHome, set := os.LookupEnv("HOME"); set {
 		if _configHome == "" {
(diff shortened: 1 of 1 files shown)


Morganamilo:
One simple mistake that got copy pasted 10 times, thanks for noticing.

Although in this case the errors will still fall through. You should keep the `return nil` at the end and replace all of thr `err = ` with returns.

main.go
```diff
@@ -10,33 +10,33 @@ import (
 	alpm "github.com/jguer/go-alpm"
 )
 
-func setPaths() error {
+func setPaths() (err error) {
```
Morganamilo: There's no need to define err here, we don't use it nor do we use a naked return. Apart from that nitpick its good.

alanjjenkins: Fixed :smile:

Morganamilo:
All's good now, thanks.

PR #476, opened 2018-06-09
Run yay as root but build as nobody

alanjjenkins:
Hey @Jguer / @Morganamilo,

I have been attempting to get yay to run as root and then switch to the nobody user to build packages but have hit a small speed bump and was wondering if I could borrow your eyes?

I believe I am almost there as it seems to be able to build the initial package you specify as nobody but it the package has unsatisfied AUR dependencies these seem to still be being built as root and I am unsure why at the moment.

Could you take a look? (I am still looking but not seeing it at the moment).

Thanks,

Alan Jenkins

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -4,10 +4,16 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
+	"os/user"
+	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


cmd.go
```diff
@@ -616,11 +676,63 @@ func passToMakepkg(dir string, args ...string) (err error) {
...
-	cmd.Dir = dir
+
+	// Run makepkg as nobody if running yay as root
+	if 0 == os.Geteuid() {
+		nobodyUser, err := user.Lookup("nobody")
+		if err != nil {
+			return fmt.Errorf("Unable to find the UID / GID of user \"nobody\" to execute makepkg: %v", err)
+		}
+
+		nobodyUid, err := strconv.Atoi(nobodyUser.Uid)
```
Jguer: Should be nobodyUID

cmd.go
```diff
@@ -616,11 +676,63 @@ func passToMakepkg(dir string, args ...string) (err error) {
...
+		nobodyGid, err := strconv.Atoi(nobodyUser.Gid)
+		if err != nil {
+			return fmt.Errorf("Unable to find the GID of user \"nobody\" to a string: %v", err)
+		}
+
+		cmd.SysProcAttr = &syscall.SysProcAttr{}
+		cmd.SysProcAttr.Credential = &syscall.Credential{Uid: uint32(nobodyUid), Gid: uint32(nobodyGid)}
+
+		// create temporary directory for nobody to build in
+		tempDir, err := ioutil.TempDir("", "yay")
```
Jguer: Needless and possibly harmful overload of tempdir.
Should be tempDir, err = ioutil.TempDir("", "yay")

Morganamilo:
I've stayed away from even attempting this because of how complicated it could be to implement. I've thought about it a little and the past and there's always a bunch of gotchas that prevent me from even wanting to try this.

Here's some of the gotchas you will need to take care of:

- How will build dir be respected? Sure you can always build in `/tmp` (and I assume this is what youre doing). What If I have a package to big to build in `/tmp`? There's `/var/tmp` which you might be able to use instead.

- How are you going to handle config reading? makepkg allows overwridding the default config in `~/.config/pacman/makepkg.conf` (`~` being `/root` of course). Also remember makepkg's config file is bash. I can write `PKGDEST=~/pkg` in it, at which point `~` should still resolve to `/root`

- The same point applies for git's config when cloning and showing diffs.

- How are you going to handle PKGDEST? The only thing I can think of is manually overriding it then using `--packagelist` to manually move the packages where they should go.

- Yay's build dir is just where the pkgbuild is cloned to. You can actually override where packages are built using `BUILDDIR`

- I've mentioned `BUILDDIR` and `PKGDEST`. don't forget about `SRCPKGDEST LOGDEST GNUPGHOME`

Pikaur's systemd approach seems to work for them. I have not tried it so I don't know if all the above points are accounted for but it does seem to work for the most basic setup (no custom configs / environment variables).

My recommended approach for this is set up a dedicated AUR user and create your own wrapper that just just does:
```sh
#!/usr/bin/bash
su aur -c "yay $1"
```

Morganamilo:
>Unfortunately the wrapper you suggest wouldn't work for the use case that I have (Running yay inside of Puppet) due to the package needing to be created as a normal user, then installed using root when Puppet has to run as root (it is an annoying issue).

Could you link to some documentation on puppet? It's not a thing I've heard of before.

If you give your custom aur user sudoless password then that wrapper can call sudo to install the packages just fine.

I don't mind looking into this pr just to see where it goes. You might want to look into http://0pointer.net/blog/dynamic-users-with-systemd.html before getting too far into it though.

AladW:
If you use that systemd thing, please make it optional and not mandatory like pikaur does. Having a helper explicitely rely on systemd whereas pacman does not is a terrible scenario in my book.

Morganamilo:
Well it would only be required when running as root. I don't know if pikaur needs it when running as a normal user.

AladW:
Last I checked, it does, but I could be wrong

Jguer:
Closing this for inactivity

PR #479, opened 2018-06-10
Fixed trivial typo

reisub0:
packags -> packages


Morganamilo:
For some reason codespell doesn't detect this. Oh well thanks for fixing it.

PR #480, opened 2018-06-10
Rework editing and diff showing

Morganamilo:
Clean build needs to happen before downloading pkgbuilds so that they
can be deletd before downloading.

Editing and diff viewing needs to happen after downloading the
pkgbuilds.

Prevously we asked to clean and edit at the same time. Then clean,
download pkgbuilds and open the editor.

This poeses a problem for diff viewing and editing. It's likley that the
user will see the diff and use that to decide if they want to edit the
pkgbuild. Using the current method, the user will be asked to view diffs
and edit before actually seeing any diffs.

Instead split cleaning diff showing and editing to three seperate menus
in the following order:
 1. show clean menu
 2. clean
 3. download pkgbuilds
 4. show diff menu
 5. show diffs
 6. show edit menu
 7. edit pkgbuilds

Also each menu is seperatly enableable. By default only the diff menu is
shows. If the user wishes to clean build, edit pkgbuilds or disable
diffs then the user can use the --[no]{clean,diff,edit}menu flags. This
replaces the --[no]showdiffs flags.

@Jguer tell me how you like this approach.

### DIFF
--- a/config.go
+++ b/config.go
@@ -65,6 +65,9 @@ type Configuration struct {
 	Provides      bool   `json:"provides"`
 	PGPFetch      bool   `json:"pgpfetch"`
 	ShowDiffs     bool   `json:"showdifs"`
+	CleanMenu     bool   `json:"cleanmenu"`
+	DiffMenu      bool   `json:"diffmenu"`
+	EditMenu      bool   `json:"editmenu"`
 }
 
 var version = "5.688"
@@ -168,7 +171,9 @@ func defaultSettings(config *Configuration) {
 	config.AnswerUpgrade = ""
 	config.GitClone = true
 	config.Provides = true
-	config.ShowDiffs = true
+	config.CleanMenu = false
+	config.DiffMenu = true
+	config.EditMenu = false
 }
 
 // Editor returns the preferred system editor.
--- a/cmd.go
+++ b/cmd.go
@@ -313,10 +313,18 @@ func handleConfig(option, value string) bool {
 		config.PGPFetch = true
 	case "nopgpfetch":
 		config.PGPFetch = false
-	case "showdiffs":
-		config.ShowDiffs = true
-	case "noshowdiffs":
-		config.ShowDiffs = false
+	case "cleanmenu":
+		config.CleanMenu = true
+	case "nocleanmenu":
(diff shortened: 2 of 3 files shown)


Jguer:
From my limited testing it seems good. I would default to cleanbuild and edit menu on by default for continuity sake.

Morganamilo:
Sure cleaning, but what about editing? I've always seen editing as a glorified view option. Diffs are meant to replace this for the most part. Those who explicitly want to edit pkgbuilds will find a way to enable that.

Jguer:
Cleaning definitely, I can compromise on that. Ok, fair enough by me

Morganamilo:
I'm ready for a new release now. And by now I mean a couple of days testing.

config.go
```diff
@@ -65,6 +65,9 @@ type Configuration struct {
 	Provides      bool   `json:"provides"`
 	PGPFetch      bool   `json:"pgpfetch"`
 	ShowDiffs     bool   `json:"showdifs"`
```
StephenBrown2: This item is now unused, can it be removed?

Morganamilo: I removed it from the defaultSettings function and all the documentation but forgot to remove it from the actual struct.

If you want to remove it yourself feel free.

glubsy:
I am confused. yay doesn't show any menu to edit PKGBUILD before starting compilation anymore? I can't seem to edit anything, only see diffs and then off it goes. :/

Morganamilo:
`yay --editmenu --save`

glubsy:
You are fast! Thank you, 
And I'm an idiot. I also checked my config file in ~/.config/yay/config.json and somehow "editmenu" was set to false.

Morganamilo:
Editmenu is now off by default as diffmenu is kind of a replacement. Personally I don't see much of a need for both. But you're free to enable both, or enable it and turn off diffs.

blargle-ragequit:
I don't see why editmenu isn't the default.  Very often I reinstall or retry installing packages solely because I need to change the PKGBUILD and have zero need for diff.

Morganamilo:
The last comment explained exactly why. You're free to enable it yourself.

PR #486, opened 2018-06-11
Add flags for diff and upgrade to match the rest and document it all

Morganamilo:


Morganamilo:
@Jguer I think all that's left now is shell completion. Where did you get them from originally by the way? Might be a good idea to pull updated shell completion for pacman 5.1.

Jguer:
I pulled them from the pacman project and in some places I found nice ~lines of code~ inspiration in apacman. I believe it is time to spruce them up a bit again yes

PR #488, opened 2018-06-12
Update bash and zsh completion

Morganamilo:


completions/zsh
```diff
@@ -52,11 +55,18 @@ _pacman_opts_common=(
...
+  '--answerupgrade[Set a predetermined answer for the upgrade menu]:answer'
   '--noanswerclean[Unset the answer for the clean build menu]'
   '--noansweredit[Unset the answer for the edit pkgbuild menu]'
   '--noanswerupgrade[Unset the answer for the upgrade menu]'
-
+  '--cleanmenu[Give the option to clean build PKGBUILDS]'
+  '--diffmenu[Give the option to show diffs for build files]'
+  '--editmenu[Give the option to edit/view PKGBUILDS]'
+  '--upgrademenu[Show a detailed list of updates with the option to skip any]'
+  '--nocleanmenu[Dont clean build PKGBUILDS]'
```
Morganamilo: Don't has been changed to Dont. This is because the `'` ended the string. Doing `\'` still caused it to end the string but I think this might have just been an error in the editor highlighting.

Morganamilo: I'll test it later to see what's really happening. If I really can't use `'` I might just write `Do not`.

Morganamilo: Just ended up uses `"` for the outer quotes instead. A bit inconsistent but it works.

PR #504, opened 2018-06-22
Remove default usage of --ask

### LINKED ISSUE #490: Make --ask optional
Not sure how to fill in the template on feature requests but I'm pretty sure this is valid for the master branch...

As discussed in https://github.com/actionless/pikaur/issues/201 usage of `pacman --ask` can have some undesired effects, like confirm the removal of conflicting packages that were not found by `yay` or presented to the user. As such the command was classified as "potentially harmful" on ArchWiki and `aurman` (the other active helper using `pacman --ask`) made it optional: https://github.com/polygamma/aurman#make-use-of-the-undocumented---ask-flag-of-pacman

I think it would be a good idea for `yay` to do the same, especially considering existing flags like `--answerclean` where a flag for `pacman --ask` would be a natural extension to.

----
#### Affected Version
https://github.com/Jguer/yay/commit/1002a4f5e5f82ab7480a91751d78b7b4c516c2eb

#### Issue
`--ask` used by default

#### Steps to reproduce

1. `yay -S <pkg>`
2.
3.

Morganamilo:
--ask is no longer used when installing AUR packages, instead pass no
confirm when we know there are no conflicts and wait for manual
confirmation when there are.

This means that when there are no conflicts there should be no change in
behaviour and the user will not need to intervene at all.

The old behaviour can still be used with --useask.

fixes #490

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -341,6 +341,10 @@ func handleConfig(option, value string) bool {
 		config.EditMenu = true
 	case "noeditmenu":
 		config.EditMenu = false
+	case "useask":
+		config.UseAsk = true
+	case "nouseask":
+		config.UseAsk = false
 	case "a", "aur":
 		mode = ModeAUR
 	case "repo":
--- a/config.go
+++ b/config.go
@@ -69,6 +69,7 @@ type Configuration struct {
 	CleanMenu     bool   `json:"cleanmenu"`
 	DiffMenu      bool   `json:"diffmenu"`
 	EditMenu      bool   `json:"editmenu"`
+	UseAsk      bool   `json:"useask"`
 }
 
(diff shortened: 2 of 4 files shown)


Jguer:
Can't forget to do fish completion for this once it's merged. I think we're missing a few options completion across the different shells.

Morganamilo:
Might want to figure out a decent order for the flags in the completion/help/man page so it is easier to compare them. Alphabetical is nice but similar flags should also be grouped together which gets a little tricky.

Jguer:
Having `--[no]editmenu [Disables]/Enables edit menu` is nice for --help
having them grouped apart in completions is ok because it is not user sensitive. Still wish there was an easier way

PR #516, opened 2018-06-30
Separate Pacman upgrade and AUR Upgrade by default

Morganamilo:
Currently When performing a system upgrade, Yay will first refresh the
database then perform the repo and AUR upgrade. This allows Yay to add
some features such as better batch interaction, showing potential
dependency problems before the upgrade starts and combined menus
showing AUR and repo upgrades together.

There has been discussion that this approach is a bad idea. The main issue
people have is that the separation of the database refresh and the upgrade
could lead to a partial upgrade if Yay fails between the two stages.

Personally I do not like this argument, there are valid reasons to Yay
to fail between these points. For example there may be dependency or
conflict issues during the AUR upgrade. Yay can detect these before any
installing actually starts and exit, just like how pacman will when
there are dependency problems.

If Yay does fail between these points, for the previously mentioned
reasons or even a crash then a simple refresh will not cause a
partial upgrade by itself. It is then the user's responsibility
to either resolve these issues or instead perform an upgrade using
pacman directly.

My opinions aside, The discussions on the Arch wiki has reached
a decision, this method is not recommended. So to follow the decided
best practises this behaviour has been disabled by default.

This behaviour can be toggled using the --[no]combinedupgrade flag

It should be noted that Yay's upgrade menu will not show repo packages
unless --combinedupgrade is used.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -345,6 +345,10 @@ func handleConfig(option, value string) bool {
 		config.UseAsk = true
 	case "nouseask":
 		config.UseAsk = false
+	case "combinedupgrade":
+		config.CombinedUpgrade = true
+	case "nocombinedupgrade":
+		config.CombinedUpgrade = false
 	case "a", "aur":
 		mode = ModeAUR
 	case "repo":
--- a/config.go
+++ b/config.go
@@ -35,41 +35,42 @@ const (
 
 // Configuration stores yay's config.
 type Configuration struct {
-	BuildDir      string `json:"buildDir"`
-	Editor        string `json:"editor"`
-	EditorFlags   string `json:"editorflags"`
-	MakepkgBin    string `json:"makepkgbin"`
-	PacmanBin     string `json:"pacmanbin"`
-	PacmanConf    string `json:"pacmanconf"`
-	TarBin        string `json:"tarbin"`
-	ReDownload    string `json:"redownload"`
-	ReBuild       string `json:"rebuild"`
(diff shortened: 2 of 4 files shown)


Morganamilo:
Commit message is really long. I intend to be the new wiki link for native pacman.

Jguer:
In favour. I'll keep using combined upgrade as I also think it's benefits outweigh "potential problems" and as I also did not contribute to the discussion on the Arch wiki I don't get to complain.

Jguer:
Yes I think we can make this a very short one, `7.887` is looking very solid. Wondering if this pull request qualifies for a major version bump

polygamma:
I really like, how both of you are always trying to further improve yay, be it with small changes or bigger changes like this one. I am reading almost every commit/PR and the constant improvement of yay is really great to see.

Morganamilo:
Thanks! maybe it's your turn for a PR  ;)

polygamma:
I have already thought about contributing to yay, I guess that's definitely going to happen.

Jguer:
Hope to see that PR Soon ;)
Mostly it's @Morganamilo doing the innovation, I'm a bit stretched on time (and I've been complaining about this for the last 5 months or so so I guess I just have to find time somewhere) but mostly I've kept to quality assurance, questions & answers ( QA and QA :smirk: ) and giving solicited and unsolicited opinions

Morganamilo:
And closing all the issues that I leave open.

PR #518, opened 2018-07-01
Document --[no]useask and --[no]combinedupgrade

Morganamilo:


cmd.go
```diff
@@ -94,6 +94,10 @@ Permanent configuration options:
     --noprovides          Just look for packages by pkgname
     --pgpfetch            Prompt to import PGP keys from PKGBUILDs
     --nopgpfetch          Don't prompt to import PGP keys
+    --useask              Automatically resolce conflicts using pacman's ask flag
```
prettyv: I'm sorry that my drive-by contribution is just pointing out a spelling error, but this should be "resolve". ;)

Morganamilo: Little late on that ;) https://github.com/Jguer/yay/pull/521/files#diff-0cf3d18277fde705f1a4eb69fddfefc1L97

prettyv: Oops, never mind then, and continue the great work! ;)

PR #524, opened 2018-07-02
Support commas in parserNumberMenu

AlexWayfer:
Please note, that it drops support for some weird untested runes like `\n`, `\v`, `\r`, etc.: https://golang.org/pkg/unicode/#IsSpace

UPD: Now not.


parser.go
```diff
@@ -629,7 +629,9 @@ func parseNumberMenu(input string) (intRanges, intRanges, stringSet, stringSet)
 	otherInclude := make(stringSet)
 	otherExclude := make(stringSet)
 
-	words := strings.Fields(input)
+	words := strings.FieldsFunc(input, func(c rune) bool {
+		return strings.ContainsRune(" \t,", c)
```
Morganamilo: Why not?

```go
return unicode.IsSpace(c) || c == ','
```

AlexWayfer: 1.  `import "unicode"` adding: https://github.com/Jguer/yay/pull/329#issuecomment-378956359; oh, there was `regexp`; oh, `unicode` doesn't increase the binary size, OK.
2. I wanted to make it simpler.

AlexWayfer: Changed.

Jguer: That graph always scares me, but I guess if you want clean graphs, you rewrite everything yourself and after that's done port to C for performance sake :laughing:. 
I agree that it's simpler to read.

AlexWayfer: Big graph like this is normal for stdlib dependencies, not for third-party. But looks scary, yes.

Jguer:
I'm okay with this, @Morganamilo ?

Morganamilo:
Yep that was my only gripe.

AlexWayfer:
Thank you. :heart:

Morganamilo:
No need to thank us when you're writing code for us :P

AlexWayfer:
> No need to thank us when you're writing code for us :P

I'm writing it for myself especially, and for community (all users).

Thank you for merging — not every project accepts every code (you should support this code in the future :yum:).

But now `yay` is more user-friendly.

Morganamilo:
Fun fact. Miscrosoft word only supports commas when choosing what pages of a document to print. I was so used to pac at the time I kept trying with spaces and I just didn't print with no error message. Took a while for it to get pointed out to me.

Maybe you should make a pull request to microsoft :P

AlexWayfer:
Oh, no, thank you. I'd rather use more user-friendly software, like LibreOffice or Google Docs. :sweat_smile: 

(But I'll be surprised if Microsoft open source code of Word; I think it's easier to be employed by them).

PR #526, opened 2018-07-02
run pacman -Qk(k) as root

### LINKED ISSUE #525: run pacman -Qk(k) as root
#### Affected Version
yay v7.887

#### Issue
when yay is used with the "-Qk(k)" switch, pacman isn't called with sudo so there are warning's about insufficient permissions:
```
pacman -Qk
warning: systemd: /usr/share/polkit-1/rules.d/systemd-networkd.rules (Permission denied)
systemd: 1442 total files, 1 missing file
```
So when running as root, the warnings are gone:
```
sudo pacman -Qk
systemd: 1442 total files, 0 missing files
```
So i would propose the "-Qk" and "-Qkk" should run as root for verifying the system consistence.

MorphBonehunter:
running the "k" commands as root ensures the right permissions on reading the files of the packages and avoid warnings about missing files

fixes #525

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -136,6 +136,9 @@ func (parser *arguments) needRoot() bool {
 		}
 		return false
 	case "Q", "query":
+		if parser.existsArg("k", "check") {
+			return true
+		}
 		return false
 	case "R", "remove":
 		return true


Jguer:
Tested, works as intended.
Well spotted, good job

Morganamilo:
I feel like 99% of the time you're not going to run that command as root, and I don't particularly want to have to type my root password every time I use this command. And by this logic `-Qo` and `-Qp` should also require root as they could act on files that the user does not have perms for.

Given that these are purely pacman commands the user could always just use `sudo pacman -Qkk` if they need.

And you merged while I was typing this :(

Morganamilo:
I'll wait until someone complains about this now always requiring root. I do use `-Qo` a lot though it's a really use full command. So as long as that's not touched I'm fine.

Jguer:
same, I only use `-Qo` in day to day setting, so if anyone else complains we revert this

MorphBonehunter:
`-Qo` doesn't need root privileges, not even when the mentioned file isn't readable:
```
ls /usr/share/polkit-1/rules.d/systemd-networkd.rules
ls: cannot access '/usr/share/polkit-1/rules.d/systemd-networkd.rules': Permission denied

yay -Qo /usr/share/polkit-1/rules.d/systemd-networkd.rules
/usr/share/polkit-1/rules.d/systemd-networkd.rules is owned by systemd 238.133-4
```

For  the `-Qp` case..."normally" the Package files are world readable, maybe there are some cases were this is not the case but all in all this option should not need root privileges.

My intention for changing `-Qk` was that i have to use this command often on an old system which crashes irregular while doing an system upgrade and after that not all files are written to disk and i check with `-Qkk` if there are corrupted files which i must reinstall.
So if pacman can not read the file, i eventually got an incomplete result.
So i was thinking this should be the generally running this as root to avoid such an behavior.

If this is not the desired behavior, feel free to revert this one.

Morganamilo:
That's fair with the other two, turns out `pacman -Qo` doesn't touch the file, just looks in the database.

My main point was `-Qk` is not normally run with root. But like I said I won't bother reverting unless some one actually gets upset over this.

AladW:
Consider that if someone has the specific use case that they want to **always** run pacman -Qk as root, they can easily create an alias. With the new change, it's impossible to run pacman -Qk through `yay` as a regular user.

PR #528, opened 2018-07-03
 Replace gopkgbuild with go-srcinfo

Morganamilo:
go-srcinfo seems ready. This is for the next release but may as well post it now.


Jguer:
I'm starting to feel this `v7x` is solid, let's give it maybe 2/3 more days

PR #535, opened 2018-07-05
Properley seprate aur and repo targets.

### LINKED ISSUE #534: Repo packages not being installed
#### Affected Version
`yay v7.887.r24.g18bbc98`
`yay-git`

#### Issue
Tried on a clean config, 
`yay -S` and `yay` both fail to install repo packages

If the package is already installed `yay` will offer to reinstall the package. This regression is only valid on packages without an existing install (could explain why upgrade is not affected)

#### Steps to reproduce
1. `yay numlock` and selection 1

1. `yay -S numlockx`

(Does not affect `yay -Si`)

@Morganamilo unless off the top of your mind you know what caused this regression, I'll check it later

Morganamilo:
When not doing combined upgrade the repo targets need to be extracted
for the early pacman call.

Fixes #534

### DIFF
--- a/install.go
+++ b/install.go
@@ -59,12 +59,44 @@ func install(parser *arguments) error {
 			parser.clearTargets()
 			arguments.clearTargets()
 
-			//seperate aur and repo targets
-			for _, target := range targets {
-				if localNamesCache.get(target) {
-					arguments.addTarget(target)
-				} else {
-					parser.addTarget(target)
+			syncDb, err := alpmHandle.SyncDbs()
+			if err != nil {
+				return err
+			}
+
+			if mode == ModeRepo {
+				arguments.targets = targets
+			} else {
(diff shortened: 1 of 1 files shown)


Morganamilo:
This is kinda a mess and is duplicating code from the dependency code. I'll do a refactor of install.go at some point. Like a real refactor. I'm happy with the way it works to I'm not gonna go rewriting it from scratch or anything.

Jguer:
I'm merging this fix on the premise you present of a proper refactor soon:tm:, also merging this because I don't like the git branch this broken

Jguer:
Took a weird while to realize this bug, I think this release is going to take a bit more

Jguer:
Didn't get it on testing and didn't even register it when I read it. That's what I get for being in a rush to get a fix. :laughing:

PR #537, opened 2018-07-06
Fix -Y and remove debug messages

Morganamilo:


Jguer:
I'm okay with tagging this release now. @Morganamilo Okay on your end?

Morganamilo:
That's all I've got.

PR #539, opened 2018-07-08
Add config option to enable or disable AUR completions

ipha:
Personally, I find the initial delay when it's building the cache annoying and the extra clutter makes it harder to find the repo package I'm looking for. So let's add an option to disable it!

Two parts to this pull request:
1) Split the cache file into separate aur and repo caches
2) Add a config option to disable aur completions

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -105,6 +105,9 @@ Permanent configuration options:
     --timeupdate          Check packages' AUR page for changes during sysupgrade
     --notimeupdate        Do not check packages' AUR page for changes
 
+    --completeaur         Include AUR packages in completions
+    --nocompleteaur       Do not include AUR packages in completions
+
 Print specific options:
     -c --complete         Used for completions
     -d --defaultconfig    Print default yay configuration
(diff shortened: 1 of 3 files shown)


Morganamilo:
Instead of a config option how about making the completion respect --aur/--repo? No idea how hard that is to implement though.

ipha:
Could do that, but the config would just have to move to the completion scripts. Not quite sure how that should work, an env var maybe?

I could probably update zsh, but I'm not familiar with bash or fish completion.

Jguer:
@ipha I wouldn't separate into 2 files. We already have a good file for repo, it's the database. We just add it to aur_fish.cache because it's convenient.
If there's an option to disable it makes far more sense to build the AUR completions on the go like pacman does. 

To alleviate network side pains I propose the following:
- Bump the update time from 48h to a user controlled value, I didn't add this option initially because I'm not such a fan of battering the AUR with requests.
- When grabbing the AUR packages file (from network), use a very small timeout so we can ignore the AUR completion if it's taking too long.

If it's not network sided, I suppose you're using bash which is painfully slow when it comes to completions, that has no cure unfortunately. 

@Morganamilo It's a pain on the completion side but doable and easy on the yay side. Making it respect that makes more sense

ipha:
Any sort of network action on completion will have a noticeable delay -- even with a short timeout.
Even with the (repo) tags, for my use-case, I never want aur completion. 

How do propose I proceed if I want a feature that just turns off aur completions?

Morganamilo:
The way I see it, the point of Yay is to add AUR support, if you're only interested in repo packages whats wrong with pacman?

Morganamilo:
Hopefully #584 is good enough for you. The network delay should be gone at least. If not you have force it away by setting `--completioninterval -1`.

PR #558, opened 2018-07-17
Fix typo

Morganamilo:


doc/yay.8
```diff
@@ -118,7 +118,7 @@ Print current yay configuration\&.
 .PP
 \fB\-n \-\-numberupgrades\fR
 .RS 4
-Depreciated, use \fByay -Qu\fR and \fBwl -l\fR instead\%.
+Deprecated, use \fByay -Qu\fR and \fBwl -l\fR instead\%.
```
StephenBrown2: Oh nooo, forgot `wl` -> `wc`!

Morganamilo: Can I use being ill as an excuse? Your turn to make a pr I guess.

PR #562, opened 2018-07-17
Add --makepkgconf

### LINKED ISSUE #560: Feature request: allow setting PKGDEST makepkg's variable for yay's usage only
Continuation of https://github.com/Jguer/yay/issues/477

Just allow setting makepkg's PKGDEST variable inside yay's configuration. This way will be possible to have different PKGDEST for makepkg generally and separate only for yay's usage of makepkg.
For me, it will be very helpful. I am often building my own AUR packages using makepkg directly and in these cases, I expect package file next to PKGBUILD.

Morganamilo:
Fixes #560

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -273,6 +273,10 @@ func handleConfig(option, value string) bool {
 		config.EditorFlags = value
 	case "makepkg":
 		config.MakepkgBin = value
+	case "makepkgconf":
+		config.MakepkgConf = value
+	case "nomakepkgconf":
+		config.MakepkgConf = ""
 	case "pacman":
 		config.PacmanBin = value
 	case "tar":
--- a/config.go
+++ b/config.go
@@ -39,6 +39,7 @@ type Configuration struct {
 	Editor          string `json:"editor"`
 	EditorFlags     string `json:"editorflags"`
 	MakepkgBin      string `json:"makepkgbin"`
+	MakepkgConf     string `json:"makepkgconf"`
 	PacmanBin       string `json:"pacmanbin"`
 	PacmanConf      string `json:"pacmanconf"`
 	TarBin          string `json:"tarbin"`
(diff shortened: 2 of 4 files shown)


TomaszGasior:
Maybe… instead of adding a new setting just check the existence of `makepkg.conf` inside ~/.config/yay and use it instead if it exists.

Jguer:
I don't like to fill up the `.config/yay` folder. We have users who store a lot of their config files in external disks without symlinks, so `--makepkgconf` is nice and in line with our other settings. If `makepkgconf` is set to "" and there's a makepkg.conf file in `~/.config/yay` we could take it as the default value.

cmd.go
```diff
@@ -672,6 +678,10 @@ func passToMakepkgCapture(dir string, args ...string) (string, string, error) {
 	mflags := strings.Fields(config.MFlags)
 	args = append(args, mflags...)
 
+	if config.MakepkgConf != "" {
```
Jguer: I know we leave a lot of things to user responsibility but here would be an easy spot to check if file exists and:
1. Warn user file does not exist
2. Don't pass --config if file does not exist

Morganamilo:
Ah so you want to do both, didn't see that coming.

Jguer:
:laughing: if you implement the file existence check I suggest, it could even be set as default

TomaszGasior:
Yay! ;) Great!

Morganamilo:
I'd much rather keep the empty default. If a user wants to set an alternative config they can do it explicitly once and save it.

Morganamilo:
Anyway the way I currently see it. An empty string should just be ignored and nothing passed to makepkg. The user can always set the config manually to `~/.config/yay/makepkg.conf`, removing a need on us to make sure it exists. (We don't check the existence of gpgbin gitbin and so on anyway)

I'll add `--nomakepkgconf` as a reset just like `--noansweredit`.

PR #569, opened 2018-07-19
Make output more similar to `pacman`s

AlexWayfer:
Before:

* Empty line after `Searching AUR`
* `There is` starts with capital letter without space before

```
> yay --combinedupgrade -Syu
 core is up to date
 extra is up to date
 community is up to date
 multilib is up to date

 there is nothing to do

> yay --nocombinedupgrade -Syu
 core is up to date
 extra is up to date
 community is up to date
 multilib is up to date
 there is nothing to do

There is nothing to do
```

After:

```
> yay --combinedupgrade -Syu
 core is up to date
 extra is up to date
 community is up to date
 multilib is up to date
 there is nothing to do

> yay --nocombinedupgrade -Syu
 core is up to date
 extra is up to date
 community is up to date
 multilib is up to date
 there is nothing to do
 there is nothing to do
```

### DIFF
--- a/install.go
+++ b/install.go
@@ -106,8 +106,6 @@ func install(parser *arguments) error {
 			arguments.options["ignore"] = ignoreStr
 		}
 
-		fmt.Println()
-
 		for pkg := range aurUp {
 			parser.addTarget(pkg)
 		}
@@ -127,7 +125,7 @@ func install(parser *arguments) error {
 
 	if len(dp.Aur) == 0 {
 		if !config.CombinedUpgrade {
-			fmt.Println("There is nothing to do")
+			fmt.Println(" there is nothing to do")
 			return nil
 		}
 
@@ -160,7 +158,7 @@ func install(parser *arguments) error {
 	}
 
 	if len(do.Aur) == 0 && len(arguments.targets) == 0 && (!parser.existsArg("u", "sysupgrade") || mode == ModeAUR) {
-		fmt.Println("There is nothing to do")
+		fmt.Println(" there is nothing to do")
 		return nil
 	}
 


install.go
```diff
@@ -106,8 +106,6 @@ func install(parser *arguments) error {
 			arguments.options["ignore"] = ignoreStr
 		}
 
-		fmt.Println()
```
AlexWayfer: I'm not sure about this. Does it affect anything else?

install.go
```diff
@@ -160,7 +158,7 @@ func install(parser *arguments) error {
 	}
 
 	if len(do.Aur) == 0 && len(arguments.targets) == 0 && (!parser.existsArg("u", "sysupgrade") || mode == ModeAUR) {
-		fmt.Println("There is nothing to do")
```
AlexWayfer: Should I make some variable or function for unifying this output or this duplicating is normal?

install.go
Morganamilo: There's a couple of new lines sprinkled over to make things read better. What each one does exactly I can not remember.

install.go
Morganamilo: This only shows up twice right? https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

AlexWayfer: Good article, thanks! I'm personally have something like "Rule of two" (**dup**licate, like **two**), but arguments for "three" make sense.

install.go
AlexWayfer: I'm looking at the 75th line and think this empty line only for upgrades.

I like readability, including via empty lines.

But `pacman` doesn't have them. And, I think, if we want to make `yay` more native with `pacman` — we should make output more similar too.

At least, the "jumping" empty line, depending on the `combinedupgrade` option, is strange (for me).

Morganamilo: I think you're confusing the native pacman wiki criteria with looking like pacman. The former is all about how AUR helpers interact with pacman and don't do dangerous things.

Personally I don't want Yay to mimic pacman's look too much. Sure I want to make sure it seems familiar but I also want to remind users that this is a different application.

But back on topic the empty line does not depend on `combinedupgrade`. It's applied either way. But yes only the upgrade menu has this line while the clean/diff/edit do not. So just for the sake of consistency I say remove it.

install.go
Morganamilo: Of course it's just a rule of thumb so don't follow it to closely. If only I could `#define` a return on error :(

install.go
AlexWayfer: > I think you're confusing the native pacman wiki criteria with looking like pacman. The former is all about how AUR helpers interact with pacman and don't do dangerous things.

No, I understand.

> Personally I don't want Yay to mimic pacman's look too much. Sure I want to make sure it seems familiar but I also want to remind users that this is a different application.

1. I like versions diff output by `yay`, and I don't want to see this super-simple output from `pacman`.
2. Just my eye clings at the same phrase (sentence) with a slightly different spelling.

> But back on topic the empty line does not depend on combinedupgrade. It's applied either way.

Yes, so I mentioned "jumping" empty line. There is it. But at different places. Also a bit strange (looks like a bug) for me.

PR #570, opened 2018-07-20
Fix devel updates

Morganamilo:
Commit 474cc56f8d77acc70ab7df264d15beea0a05cc91 refactored the
getCommit() function to use the standard passToGit() and capture()
functions.

Unlike every other external call, here we use .Wait() and kill the
command after 5 seconds. This was missed and ended up breaking the
function.

So instead don't use capture but the manual .Wait() call. passToGit() is
still used.

### DIFF
--- a/vcs.go
+++ b/vcs.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


Morganamilo:
It's just mostly refactoring. Devel updates were broken in git and fixed in git. That never made it to release if that's what you're asking.

Jguer:
I kept checking the backlog, I noticed that. 
949 commits, 954 stars, we could even marry a 1000commits, 1000 stars release and rethink the release versioning as you talked a while ago :laughing: although this scheme hasn't worked bad at all for now (except for the overly arbitrary major version bumps).

PR #573, opened 2018-07-21
Wait for db.lck to become available before starting a db operation

Jguer:
On split updates some funkiness can happen where an `yay` instance is unlocked before the other one gets to the 2nd pacman call but it should be working resulting in:
```
yay-1 -Syu
yay-2 -Syu
yay-1 pacman
yay-2 pacman
yay-1 pacman
yay-2 pacman
```

It should be stable enough for forced deployment without a flag though

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -54,9 +54,9 @@ func updateSudo() {
 }
 
 // waitLock will lock yay checking the status of db.lck until it does not exist
-func waitLock() (err error) {
+func waitLock() {
 	if _, err := os.Stat(alpmConf.DBPath + "db.lck"); err != nil {
-		return nil
+		return
 	}
 
 	fmt.Println(bold(yellow(smallArrow)), "db.lck is present. Waiting... ")
@@ -64,7 +64,7 @@ func waitLock() (err error) {
 	for {
 		time.Sleep(3 * time.Second)
 		if _, err := os.Stat(alpmConf.DBPath + "db.lck"); err != nil {
-			return nil
+			return
 		}
 	}
 }
--- a/exec.go
+++ b/exec.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"time"
 )
@@ -55,7 +56,7 @@ func updateSudo() {
 
 // waitLock will lock yay checking the status of db.lck until it does not exist
 func waitLock() {
-	if _, err := os.Stat(alpmConf.DBPath + "db.lck"); err != nil {
+	if _, err := os.Stat(filepath.Join(alpmConf.DBPath, "db.lck")); err != nil {
 		return
 	}
 
@@ -63,7 +64,7 @@ func waitLock() {
 
 	for {
 		time.Sleep(3 * time.Second)
-		if _, err := os.Stat(alpmConf.DBPath + "db.lck"); err != nil {
+		if _, err := os.Stat(filepath.Join(alpmConf.DBPath, "db.lck")); err != nil {
 			return
 		}
 	}
--- a/exec.go
+++ b/exec.go
@@ -53,6 +53,17 @@ func updateSudo() {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


exec.go
```diff
@@ -53,6 +53,17 @@ func updateSudo() {
 	}
 }
 
+// waitLock will lock yay checking the status of db.lck until it does not exist
+func waitLock() (err error) {
+	for {
+		if _, err := os.Stat("/var/lib/pacman/db.lck"); os.IsNotExist(err) {
```
Morganamilo: This should be join pacmanconf.DbPath + db.lck

parser.go
```diff
@@ -177,6 +177,50 @@ func (parser *arguments) needRoot() bool {
...
+		return true
+	case "F", "files":
+		if parser.existsArg("y", "refresh") {
+			return true
+		}
+		return false
+	case "R", "remove":
+		return true
+	case "S", "sync":
+		if parser.existsArg("y", "refresh") {
```
Morganamilo: These are redundant, they'll be caught by the default return true at the end.

Jguer: The default is return false though, I did notice needRoot does have a lot of redundancy

Morganamilo: The default inside of sync is true though.

exec.go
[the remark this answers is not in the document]
Jguer: That else if would return nil as well and abort the lock so any error actually will return nil

exec.go
```diff
@@ -53,6 +53,17 @@ func updateSudo() {
...
+			return nil
+		}
+		fmt.Println(bold(yellow(smallArrow)), "db.lck is present. Waiting 3 seconds and trying again")
```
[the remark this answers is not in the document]
Jguer: I'll do an initial check, print the message and then loop

parser.go
Jguer: True. :innocent:

Jguer: Maybe needRoot needs a touch up as well

Morganamilo: Probably, been a long time since I've even looked at it.

exec.go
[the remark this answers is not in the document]
Morganamilo: Yeah I just noticed you're not even handling errors. In that case why have a return value at all?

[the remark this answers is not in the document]
Jguer: out of habit

exec.go
```diff
@@ -53,6 +53,22 @@ func updateSudo() {
...
+func waitLock() {
+	if _, err := os.Stat(alpmConf.DBPath + "db.lck"); err != nil {
+		return
+	}
+
+	fmt.Println(bold(yellow(smallArrow)), "db.lck is present. Waiting... ")
+
+	for {
+		time.Sleep(3 * time.Second)
+		if _, err := os.Stat(alpmConf.DBPath + "db.lck"); err != nil {
```
Morganamilo: filepath.Join ;)

Morganamilo:
I think it's good now apart from my last comment (and that I have not actually ran the code)

Morganamilo:
Testing this out I have the most nit picky ocd complaint. The cursor is put on the next line while it waits. This annoys me for some reason. Apart from that works well.

What do you think about this instead?

```diff
diff --git a/exec.go b/exec.go
index 2a97ae3..a5b95dc 100644
--- a/exec.go
+++ b/exec.go
@@ -60,11 +60,12 @@ func waitLock() {
                return
        }
 
-       fmt.Println(bold(yellow(smallArrow)), "db.lck is present. Waiting... ")
+       fmt.Print(bold(yellow(smallArrow)), " db.lck is present. Waiting...")
 
        for {
                time.Sleep(3 * time.Second)
                if _, err := os.Stat(filepath.Join(alpmConf.DBPath, "db.lck")); err != nil {
+                       fmt.Println()
                        return
                }
        }
```

Jguer:
That is the most nit picky ocd complaint :laughing: , merge it with #576

PR #575, opened 2018-07-21
Add --{no,ask}removemake

Morganamilo:
Fixes 574


AladW:
With all the command-line options `yay` has now, I propose to rename it to `GNU Yay` on the next release. :wink:

Morganamilo:
Should we also refactor the code to?

```go
func
foo() {

}
```

PR #576, opened 2018-07-21
Remove unused code, fix typo for -g flag

Morganamilo:
As it turns out, the times you need root also tend to be the time you
need to manipulate the database. So the needWait() function can be
removed and repllaced by needRoot()

Also this brings us back down to more lines lost than gained since last release ;)

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -405,20 +405,6 @@ func handleYogurt() (err error) {
 func handleSync() (err error) {
 	targets := cmdArgs.targets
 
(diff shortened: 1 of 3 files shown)


Morganamilo:
Whoops still had the code from https://github.com/Jguer/yay/pull/573#issuecomment-406804637 checked out. Guess it can be part of the commit anyway.

Jguer:
> Also this brings us back down to more lines lost than gained since last release ;)

That makes me so happy.

PR #577, opened 2018-07-21
yay.8: typo in --nopgpfetch flag

AladW:


AladW:
And of course, I messed up the commit message

Morganamilo:
image

Thanks for the laugh, I'll see you in an hour.

AladW:
Hey, you and @polygamma can see the future, I *live* in it.

Morganamilo:
Is it sad that this is the funnies thing I've seen today? I figured it out though. Windows and Linux store time differently, so every time I boot windows it's an hour behind. Although I do it so rarely I don't care enough to fix it. And I also forget this fact every time.

Morganamilo:
Anyway, thanks for proving at least some people read the man page ;)

AladW:
I guess you can set Linux to use localtime... or [set Windows to use UTC](https://wiki.archlinux.org/index.php/Time#UTC_in_Windows).

Morganamilo:
I really can not be bothered to deal with any of it. Hell I was using the default background until my bf changed it behind my back. Andhttps still works even though I feel it shouldn't.

Plus on top of that I think it's a daylight saving issue, so it only happens half of the year lol

PR #578, opened 2018-07-22
Refacotr continueTask()

Morganamilo:
The main reason behind this is for future localisation. yes and no can
be set to the localized equivalent and it should just work.

This Refactor also changes the function in ways which make it much less
confusing.

The param is no longer reversed and changed to a boolean. Before you had
to pass in Yy to get a default of no and vice versa.

The function now retuens false for no and true for yes. Before it would
return true if the default option was choosen.

### DIFF
--- a/clean.go
+++ b/clean.go
@@ -82,7 +82,7 @@ func syncClean(parser *arguments) error {
 	fmt.Println()
 	fmt.Printf("Build directory: %s\n", config.BuildDir)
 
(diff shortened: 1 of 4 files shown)


Jguer:
On-topic but with no contribution, just complaining: it was annoying (af) and easy to break.

Morganamilo:
Was kinda wondering what the hell you were thinking when you made it to be honest.

PR #584, opened 2018-07-24
 Rework completion

Morganamilo:
Bash seperates on whitespace, so the fish completion file
actually works for bash and zsh. So remove the concept of shells
entirley and just use the singular aur_sh.cache file.

If for some reason output without the repository data is needed, the
user could always just pipe it into awk like so
`yay -Pc | awk '{print $1}'`. Or perhaps a --quiet option could be added
where yay will strip the output itself.

The completion cache now updates when installing AUR packages. This is
done as a goroutine with no wait groups. This ensures the program will
never hang if there is a problem.

The completion is stil updated during -Pc but as long as an AUR package
has been installed recently it should not need to update.

The cache will now also wait 7 days instead of 2 before refreshing.
A refresh can be forced using -Pcc.

Make the completion refresh time configurable
The default setting is 7 days. The user can specify a different time in
days. -1 can be set to never refresh while 0 can be used to always
refresh.

Alternative to #539

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -361,13 +361,10 @@ func handlePrint() (err error) {
 		err = printUpdateList(cmdArgs)
 	case cmdArgs.existsArg("w", "news"):
 		err = printNewsFeed()
+	case cmdArgs.existsDouble("c", "complete"):
+		complete(true)
 	case cmdArgs.existsArg("c", "complete"):
-		switch {
-		case cmdArgs.existsArg("f", "fish"):
-			complete("fish")
-		default:
-			complete("sh")
-		}
+		complete(false)
 	case cmdArgs.existsArg("s", "stats"):
 		err = localStatistics()
 	default:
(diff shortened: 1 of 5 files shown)


Morganamilo:
`--completionrefreshtime` is quite the mouthful. A better name would be appreciated.

Morganamilo:
I've already got more commits on the way but I'll make this one the last one for this release.

Jguer:
Getting a name for that is a bit of a pain. A few alternatives I don't like as much as the present option:
- `cacheduration` (can be misinterpreted with pkg cache
- `completioninterval`

Morganamilo:
`completioninterval` is a bit long but an improvement over `completionrefreshtime`.

Morganamilo:
Also because there is now only one completion cache and it how has the fish format, the current aur_sh cache will need to be refreshed to get the proper formatting. There's a couple solutions I see.

- As well as the 7 day and file not exist check. Force a refresh if file.lastmodified is before current date
- Keep the aur_fish.cache file and use that for every shell. Instead using aur_sh.cache
- Wait for the 7 days to be up and it will take care of itself, also suggest users run `yay -Pcc`

Jguer:
aur_sh is actually a non descriptive name because it contains both repo and aur pkgs, I would  move it to sh.cache and deprecate all others

PR #585, opened 2018-07-25
Support Usage in pacman.conf

Morganamilo:
This also increases out reliance on alpm. The code is now faster, more correct and there is less of it. (Although you can argue there's more code on the go-alpm side)

Depends on https://github.com/Jguer/go-alpm/pull/17

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -11,13 +11,13 @@
   branch = "master"
   name = "github.com/jguer/go-alpm"
   packages = ["."]
-  revision = "1114f773cdfb05f577438f7a0538eccabc9cf012"
+  revision = "c3ee958efac942186012cc67de8fe5e7a5b3685d"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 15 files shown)


Jguer:
I'm okay with moving stuff to `go-alpm`, seems cleaner code anyway

Morganamilo:
Reading over this again in the morning, I've realized a typo or two and I think this stops the "ignoreing upgrade" mesage for repo upgrades. But that can probably be fixed by letting the logcb or however pacman prints them.

Morganamilo:
Sadly we can not control the output of alpm's warnings. So it looks a little out of place.

image

We could restructure out warnings to be  `pkgname: text (version1 => version2)` and it would be close, we would lose the colour though which is kind of annoying.

Jguer:
That is kind of problematic, We could keep the colour and still change to `=>` anyway and change to example: -> linux-ck: ignoring package upgrade  (version1 => version2) with the colour

PR #588, opened 2018-07-25
Fixed typos and improved punctuation consistency in README

muesli:
Just a couple of nitpicky grammar, typo and punctuation fixes for the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,11 +10,11 @@ So say hi to 20+1.
 
 Yay was created with a few objectives in mind and based on the design of [yaourt](https://github.com/archlinuxfr/yaourt), [apacman](https://github.com/oshazard/apacman) and [pacaur](https://github.com/rmarquis/pacaur):
 
-* Have almost no dependencies.
-* Provide an interface for pacman.
-* Have yaourt like search.
+* Have almost no dependencies
+* Provide an interface for pacman
+* Have yaourt-like search
 * Minimize user input
-* Know when git packages are due for an upgrade.
+* Know when git packages are due for an upgrade
 
 ## Features
 
@@ -23,7 +23,7 @@ Yay was created with a few objectives in mind and based on the design of [yaourt
 * Ask all questions first and then start building
 * Search narrowing (`yay linux header` will first search linux and then narrow on header)
 * No sourcing of PKGBUILD is done
-* The binary has no dependencies that pacman doesn't already have.
+* The binary has no dependencies that pacman doesn't already have
 * Advanced dependency solving
 * Remove make dependencies at the end of the build process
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,7 +44,7 @@ makepkg -si
 
 Contributors are always welcome!
 
-If you plan to make any large changes or changes that may not be 100% agreed
+If you plan to make larger changes or changes that may not be 100% agreed
```
Morganamilo: I prefer the old wording. Is there anything wrong with it?

README.md
```diff
@@ -107,11 +107,12 @@ Currently yay Depends on:
   `yay -{OPERATION} --repo`
 
 ### `Out Of Date AUR Packages` message is displayed, why doesn't `yay` update them?
-  This means the package has been flagged out of date on the AUR but maintainer has not updated the `PKGBUILD` yet.
+  This means the package has been flagged out of date on the AUR, but its
+  maintainer has not updated the `PKGBUILD` yet.
```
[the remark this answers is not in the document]
muesli: Sounds fine to me, both work. Will amend.

README.md
muesli: The combination of "any" with the plural "changes" sounds a bit weird to some people, even though it's not technically incorrect as far as I know. I believe this change makes it sound a bit more natural (native speakers: please correct me here).

Morganamilo: Well larger means bigger than something. So to me when you say larger without comparing it to anything it does not quite make sense.

muesli: Then we should probably remove the word "large" altogether, as you can't really quantify code changes that way (where does large begin?). How about "If you plan to make more involved/complex/elaborate/intricate changes..."?

Morganamilo: You could say the same thing about all of these. What makes a change complex? Yes large is subjective but that's fine. And a change does not have to be complex to merit an issue anyway. You could be tearing out a large piece of the code base in order to make it less complex. Basically what #585 is doing right now.

Morganamilo: @polygamma you're stalking the repo right now. Got any opinions with your German-English?

polygamma: my 2 cents: `If you plan to make any large changes` is fine

muesli: As mentioned I'm not a native speaker either, but I have spend half my life in english speaking countries. The word 'any' with an innumerable plural object simply sounds a bit awkward.

A 'large change' by definition stands in contrast to a smaller change, hence why I would use the comparative "larger" here.

"Edits from maintainers" are enabled, so you do you here.

Jguer: Both of them sound fine. Personally I prefer `any large changes` even if it's more Americanised

Morganamilo: @muesli There's three of us in agreement on keeping it. It would be much easier if you edited your own commits than asking us to do it.

Morganamilo:
Didn't realize you made the change already. (rebasing a PR does not give email notifications). Anyway thanks for the changes.

PR #593, opened 2018-07-28
'Packages to cleanBuild?' prompt has misspellings

awareness481:
The following prompt as of now looks like this
`+ cyan(" [N]one ") + "[A]ll [Ab]ort [I]nstalled [No]tInstalled or (1 2 3, 1-3, ^4)"))`

However in the [No]tInstalled the 't' is incorrectly placed outside the brackets. I changed it from
[No]**t** to [No**t**]

I use the following command to trace the misspelling
`$ grep -RiI "No]t" .`
so to my understanding this error happens only in the install.go file. However my knowledge and experience with grep is limited.

### DIFF
--- a/install.go
+++ b/install.go
@@ -521,7 +521,7 @@ func cleanNumberMenu(pkgs []*rpc.Pkg, bases map[string][]*rpc.Pkg, installed str
 	}
 
 	fmt.Println(bold(green(arrow + " Packages to cleanBuild?")))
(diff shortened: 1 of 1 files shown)


awareness481:
While taking a further look into the `install.go` file it seems that my PR is inconsistent with the following code:
```go
if !installed.get(pkg.Name) && (cOtherInclude.get("no") || cOtherInclude.get("notinstalled")) {
    toClean = append(toClean, pkg)
       continue
}
```

However, I'm leaving the PR open for the reviewers.

Morganamilo:
This simply means: type `a` for abort, just like type `y` for yes. None and NotIntalled both happen to start with N so one needs an extra letter to differentiate. The exact same thing is done with All and Abort.

PR #594, opened 2018-07-28
Update man page and usage

Morganamilo:


cmd.go
```diff
@@ -52,8 +52,11 @@ Permanent configuration options:
     --gpg         <file>  gpg command to use
     --gpgflags    <flags> Pass arguments to gpg
     --config      <file>  pacman.conf file to use
+    --makepkgconf <file>  makepkg.conf file to use
+    --nomakepkgconf       Use the default makepkg.conf
 
     --requestsplitn <n>   Max amount of packages to query per AUR request
+    --completioninterval  <n> Time in days to to refresh completion cache
```
Morganamilo: This flag is annoyingly long :(

Jguer:
Good because I was out for the weekend and out of data. I would really like a flag smaller than completion interval but everything else is non descriptive

Morganamilo:
That's why I have yet to merge this. Just hoping a better name appears out of nowhere. Theres `cinterval` but like you say its not very descriptive.

PR #595, opened 2018-07-28
Use no-progress when cloning packages

Morganamilo:
The output is annoying and takes up many lines. All AUR packages clone
extremley fast anyway, making the progress unneeded.

### DIFF
--- a/download.go
+++ b/download.go
@@ -60,7 +60,7 @@ func gitHasDiff(path string, name string) (bool, error) {
 func gitDownload(url string, path string, name string) (bool, error) {
 	_, err := os.Stat(filepath.Join(path, name, ".git"))
 	if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


Jguer:
With this out of the way, should we concurrently git clone?

Morganamilo:
That's part of the reason I added the flag. I have this working well with the aurlib repo I have been experimenting. Which normally I don't really push to because it's just me messing around. The fetcher seems to work pretty nicely though.

https://github.com/Morganamilo/aurlib

Make sure you have `vifm` installed and run `go run test/fetch.go`. Once the initial fetch is done, you can wait for one of the pkgbuilds to update (or in mush less time: cd into `.clone/pkgname` and `git reset --hard HEAD^`). Then next time you run the test you will also get a diff of the changes.

PR #600, opened 2018-07-30
Update completion

Morganamilo:


Morganamilo:
I think this is it. Nothing more I want to push into this release.

PR #601, opened 2018-07-31
Only minor changes

xyproto:
* 50 fewer lines
* introducing the `must` function and `defer` in `main.go`
* favoring the "early return" style, which allows for reduction of indentation levels
* avoiding named returns in a couple of places (but there is not much of a difference for these particular cases)

Accept or reject at will.

Thanks for creating yay!

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -396,7 +396,7 @@ func handleYogurt() error {
 	return numberMenu(cmdArgs.targets)
 }
 
-func handleSync() (err error) {
+func handleSync() error {
 	targets := cmdArgs.targets
 
 	if cmdArgs.existsArg("s", "search") {
--- a/cmd.go
+++ b/cmd.go
@@ -171,15 +171,10 @@ func handleCmd() (err error) {
 }
 
 func handleQuery() error {
-	var err error
-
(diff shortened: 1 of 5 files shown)


cmd.go
```diff
@@ -373,35 +367,33 @@ func handlePrint() (err error) {
...
-func handleYogurt() (err error) {
+func handleYogurt() error {
 	config.SearchMode = NumberMenu
-	err = numberMenu(cmdArgs.targets)
-
-	return
+	return numberMenu(cmdArgs.targets)
 }
 
 func handleSync() (err error) {
```
Jguer: If these changes are made, it is not needed to name the return

xyproto: Agree, I can update the pull request.

xyproto:
The desired granularity for pull requests differ from project to project, and is hard to specify in the contribution guidelines.

xyproto:
Split out the changes to `main.go` to #602.

Morganamilo:
As long as this is done consistently I don't mind. Might wanna rebase this against #585 though, as that's gonna be the next thing that gets merged as soon a we get a release.

xyproto:
Thanks for merging!

PR #602, opened 2018-07-31
Use a `cleanup` function instead of `goto`

xyproto:
* Uses a `cleanup` function instead of `goto cleanup`.
* Also introduces a `must` function for less verbose error handling.

Split out from https://github.com/Jguer/yay/pull/601, as requested by @Jguer.

### DIFF
--- a/main.go
+++ b/main.go
@@ -45,43 +45,43 @@ func setPaths() error {
 	return nil
 }
 
-func initConfig() (err error) {
+func initConfig() error {
 	defaultSettings(&config)
 
-	if _, err = os.Stat(configFile); os.IsNotExist(err) {
+	if _, err := os.Stat(configFile); os.IsNotExist(err) {
 		err = os.MkdirAll(filepath.Dir(configFile), 0755)
 		if err != nil {
 			err = fmt.Errorf("Unable to create config directory:\n%s\n"+
 				"The error was:\n%s", filepath.Dir(configFile), err)
-			return
+			return err
 		}
 		// Save the default config if nothing is found
 		config.saveConfig()
-	} else {
-		cfile, errf := os.OpenFile(configFile, os.O_RDWR|os.O_CREATE, 0644)
-		if errf != nil {
-			fmt.Printf("Error reading config: %s\n", err)
-		} else {
-			defer cfile.Close()
-			decoder := json.NewDecoder(cfile)
-			err = decoder.Decode(&config)
-			if err != nil {
-				fmt.Println("Loading default Settings.\nError reading config:",
-					err)
-				defaultSettings(&config)
-			}
-			if _, err = os.Stat(config.BuildDir); os.IsNotExist(err) {
-				err = os.MkdirAll(config.BuildDir, 0755)
-				if err != nil {
-					err = fmt.Errorf("Unable to create BuildDir directory:\n%s\n"+
-						"The error was:\n%s", config.BuildDir, err)
-					return
-				}
-			}
+		return err
+	}
+
+	cfile, err := os.OpenFile(configFile, os.O_RDWR|os.O_CREATE, 0644)
+	if err != nil {
+		fmt.Printf("Error reading config: %s\n", err)
+		return err
+	}
+	defer cfile.Close()
+
+	decoder := json.NewDecoder(cfile)
+	if err := decoder.Decode(&config); err != nil {
+		fmt.Println("Loading default Settings.\nError reading config:", err)
+		defaultSettings(&config)
+		return err
+	}
+
+	if _, err := os.Stat(config.BuildDir); os.IsNotExist(err) {
+		if err = os.MkdirAll(config.BuildDir, 0755); err != nil {
+			return fmt.Errorf("Unable to create BuildDir directory:\n%s\n"+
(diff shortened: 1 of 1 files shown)


Morganamilo:
The entire point of the goto was to make sure cleanup is called even if an error occurs. This kinda undoes that.

Morganamilo:
Scratch that, I was just going by the comment in d1c83a3fe784f30c42cd489d547ebc62996f5e38. 

Personally I would remove the defer and just stick the cleanup at the end. That way there is no defer to ignore, which I find more clear.

xyproto:
If the cleanup happens at the end, there needs to be a way to get to the end, either by `goto` or `defer`. As long as `os.Exit` is used, there is no way to make sure the code reaches the end of the main function.

I personally think `goto` is ugly, that `defer` is a step up, but that the best way would be to structure the code so that neither would be needed.

There is also `runtime.Goexit` which can be used instead of `os.Exit` to make sure the functions set up with `defer` are called at exit. If using `runtime.Goexit`, one might have to end the program manually after that, using `os.Exit`.

xyproto:
See also: https://stackoverflow.com/a/39755730/131264

main.go
```diff
@@ -171,93 +169,57 @@ func initAlpm() (err error) {
...
-	if err != nil {
-		if err.Error() != "" {
-			fmt.Println(err)
-		}
-
-		status = 1
-		goto cleanup
-	}
+	// Ensure release of alpmHandle, if os.Exit is not called
+	defer cleanup()
```
Morganamilo: I'm talking about this cleanup specifically. This could be to the end of the function and it will not change the behaviour. No defer = no possibility of getting confused about os.Exit cancelling it.

xyproto: I agree that the combination of `defer` and `os.Exit` is unfortunate, since `os.Exit` does not trigger the deferred functions. Placing `os.Exit` in the final deferred function would solve this.

xyproto:
Found a solution that I think is a slight improvement over the previous `goto cleanup`:

For the current proposed changes:

* There is no duplication of error-checking code.
* There is no duplication of the cleanup code, and it is ensured to be run since it's a deferred function.
* `os.Exit` is only called after any deferred functions are run.

Tested with both successfull runs and runs where I introduced errors to stop the execution, and the error handling and cleanup works correctly here for all cases.

xyproto:
@Morganamilo

What do you mean by "avoid unexpected behavior" in the context of `defer`?

Moving the cleanup to the end requires checking each and every critical function call for errors within the main function, with a lot of duplication of code. If this is the desired behavior, then at least the code should use the standard `if err != nil` pattern instead of using `goto cleanup`.

Renaming `must` to `exitOnErr` is possible, but for the case of the proposed changes in this pull request, `exitOnErr` does not actually exit, it only stops the execution (with the exception of the deferred functions), so it would be very misleading to name it something with `exit`.

Morganamilo:
>What do you mean by "avoid unexpected behavior" in the context of defer?

Little gotchas like defer not kicking in because os.Exit was called, or being confused because the main function is over complicated and hard to read,  

>exitOnErr does not actually exit

It did when I was suggesting the change originally.

>Moving the cleanup to the end requires checking each and every critical function call for errors within the main function, with a lot of duplication of code. 

Why would it? You wrap each call in `must/exitOnErr`, if there's on error then the function will call cleaup and exit. Therefore you have one cleanup call in `must/exitOnErr` and another at the end of main.

xyproto:
> Little gotchas like defer not kicking in because os.Exit was called, or being confused because the main function is over complicated and hard to read,

I think the code using the `must` function is easier to read. It also avoids `goto`, results in less duplication of code and also fewer lines of code. These are all good things. However, I agree with you that the downside is that the combination of `runtime.Goexit` + `os.Exit` + `defer` may be to subtle and/or confusing. On the other hand, all the code involved is close to each other, so it should be relatively easy to untagle.

But still, it's a downside that it's not as straightforward as it could be, I agree, so I'm closing this pull request.

Morganamilo:
You don't have to have the downsides though. I was mostly happy with d1c83a3 all I was asking was for you to move one line and a couple of other things. No need to throw the baby out with the bathwater.

Morganamilo:
It's cumbersome for us too. Like this PR, you named it `Use a cleanup function instead of goto` but it does a lot more than that, in fact it does a lot of the refactoring that #601 does. Really the refactor of the syntax and the main() cleanup should be separate but you decided to split out the entire file instead.

I know @Jguer said to split it to a new PR but if it was one PR with multiple commits that would work just as well. Or at the very least describe your commit.

And you seemed to ignore my comments of just moving the deferred line to the end of the function and came up with some over complicated solution. af08ae954d2acaa3bb9d68576f10ff0b1ea10c59 seems to work fine.

Also not to mention you still have not applied the changes mentioned here https://github.com/Jguer/yay/pull/603#issuecomment-409339920

You're welcome to contribute, but I would recommend you stick to the basic git principles of, one change per commit and describe your changes.

xyproto:
Let me rephrase, with stronger wording: For me, the review process is centered too much around what you said or didn't say, and what you like and didn't like. I don't care about your opinion, only about contributing to a project that I like. If I have to care about your opinion, I would rather contribute to a different project.

PR #603, opened 2018-07-31
Add a link/shield to Go Report Card

xyproto:


Morganamilo:
The badges are all under the package heading. How about we move the licence and report card next to the  yay heading? Plus I prefer the licence on the right.

Jguer:
Agreed, package heading is for packages, it shouldn't have had the license there to begin with.
Moving the license to yay heading or if it looks good as part of subtitle.

StephenBrown2:
Also, it appears that the badge URL used is hardcodeed to A+. To get a live badge, use this: https://goreportcard.com/badge/github.com/Jguer/yay

Morganamilo:
Honestly I'm surprised we got an A. I mean have you ever looked at install.go? *shudders*

xyproto:
@StephenBrown2 The badge link URL was correctly set to https://goreportcard.com/badge/github.com/Jguer/yay, but the image had the wrong one. Thanks for pointing that out, updated the pull request.

Morganamilo:
I've heard that phrase enough times (it's even his username) that I think it's finally time i asked. What exactly did Allan do to get that tittle?

AladW:
The last I remember: https://lists.archlinux.org/pipermail/arch-dev-public/2015-October/027467.html

StephenBrown2:
@xyproto Yeah, I accidentally the word "image" in my sentence. Thanks for the quick fix!

Jguer:
Badges are a bit silly anyway, the package version ones are useful-ish (I would to put the manjaro yay ones around because even in unstable they are constantly outdated, even when group install was broke it took 4 days for manjaro users to get it in their repos). 

The go report card is going to give you an A+ as long as your team has the habit of following the linter.

xyproto:
Having badges is an indicator that the project is having a minimum of testing and code standards. Much more so than the actual value on the badge, be it "A+" or something else.

PR #605, opened 2018-07-31
Ensure devel packages exist when updating

### LINKED ISSUE #604: Devel update is broken when package is removed from the AUR
#### Affected Version
yay v8.922.r74.g892d9cc - libalpm v11.0.1

#### Issue
If there is devel package installed that is then removed from the AUR. Yay will still attempt to update the package when a new commit happens, leading to a target not found error.

Devel updates should also check if the package exists on the AUR and remove them from vcs.json otherwise.

Morganamilo:
Fixes #604 

Big diff because variable rename and some blocks got indented an extra layer. But it's really not that bad.

### DIFF
--- a/upgrade.go
+++ b/upgrade.go
@@ -121,7 +121,7 @@ func upList(warnings *aurWarnings) (aurUp upSlice, repoUp upSlice, err error) {
 	var repoErr error
 	var aurErr error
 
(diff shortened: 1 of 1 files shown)


Jguer:
It's simple enough logic wise I don't think it will break `yay-git`. My hope would be tag and release tomorrow

PR #614, opened 2018-08-02
Refactor main

Morganamilo:
Instead of #602

### DIFF
--- a/main.go
+++ b/main.go
@@ -45,98 +45,88 @@ func setPaths() error {
 	return nil
 }
 
-func initConfig() (err error) {
+func initConfig() error {
 	defaultSettings(&config)
 
-	if _, err = os.Stat(configFile); os.IsNotExist(err) {
-		err = os.MkdirAll(filepath.Dir(configFile), 0755)
-		if err != nil {
-			err = fmt.Errorf("Unable to create config directory:\n%s\n"+
+	if _, err := os.Stat(configFile); os.IsNotExist(err) {
+		if err = os.MkdirAll(filepath.Dir(configFile), 0755); err != nil {
+			return fmt.Errorf("Unable to create config directory:\n%s\n"+
 				"The error was:\n%s", filepath.Dir(configFile), err)
-			return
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -45,220 +45,168 @@ func setPaths() error {
...
+			return fmt.Errorf("Unable to create config directory:\n%s\n"+
 				"The error was:\n%s", filepath.Dir(configFile), err)
-			return
 		}
 		// Save the default config if nothing is found
-		config.saveConfig()
-	} else {
-		cfile, errf := os.OpenFile(configFile, os.O_RDWR|os.O_CREATE, 0644)
-		if errf != nil {
-			fmt.Printf("Error reading config: %s\n", err)
```
Morganamilo: @Jguer  There's a ton of instances in this function where errors are either printed or ignored entirely. Do you have any logic to what to ignore and what to report? I've return most of the errors here apart from the vcs stuff as I guess that is non essential.

Jguer: For example, cfile unable to load won't prevent normal function since default options are loaded, I think as long as `yay` can install a package even if the error occurs, no return and just print was the logic

Morganamilo: Alright I'll change it to work like that. Although In the current release, I'm pretty sure it exits if it cannot make ~/.config/yay, which kind of defeats the point, I'll change that too.

Morganamilo: So playing around with stuff. I can't seem to think of a good reason not to return if the config file can not be opened/parsed. If the json is invalid then the user should probably fix that, if there's a disk error it's probably best not to go on and install a bunch of stuff.

What I have done is make it so that yay will never write to the config file unless `--save` is used. The same has also been done for vcs data and `--gendb`. We don't really need to make sure they exist, lets just make it whenever we write to it, instead of during initialisation.

the cachedir, configdir and builddir will still be crated at initialisation though.

Jguer: That seems fair enough, in one week I'll be back in home base and ready to check on a release

PR #615, opened 2018-08-02
 Add yay specific args, drop conflicting args 

Morganamilo:


Morganamilo:
Got any suggestions to replace `--print` or are you fine with just the short option?

AladW:
I guess that would make it the only short option without corresponding long option. No idea is this of any importance. I'd consider `--yayconfig` instead of `--currentconfig` though, to emphasize the difference with `--config`.

Morganamilo:
>I guess that would make it the only short option without corresponding long option. No idea is this of any importance.

It's very important for my OCD. 

>I'd consider --yayconfig instead of --currentconfig though, to emphasize the difference with --config.

We already have `--defaultconfig`, `--currentconfig` matches it very well.

Jguer:
Only alternative of `--print` I can think right now is `--puke` information. Only the short option is not very satisfying, even if I've never used the long option

Morganamilo:
```
--show
--output
--out
--echo
```

Are all I can think of right now. --show is my pick, even if it is a little annoying that it does not start with p.

```
show news
show currentconfig
show stats
show completion
```

All make perfect sense to me (although the completion flag is actually complete). If there's no objection I'll update the PR and merge it when I get round to it.

Jguer:
`--show` is okay I guess, just going to be a nice laugh when someone asks why `-Sc` is deleting their cache instead of printing completions

Morganamilo:
Eh, -S is probably the most well known pacman operation. If they confuse --show for -S they've only got themselves to blame.

PR #619, opened 2018-08-03
Align version strings right

Jakeler:
It is imho easier to spot the version differences on varying lengths if they are aligned right.
For example what only a new package release and what a real software update is.
comparison

### DIFF
--- a/print.go
+++ b/print.go
@@ -174,7 +174,7 @@ func (u upSlice) print() {
 	}
 
 	namePadding := fmt.Sprintf("%%-%ds  ", longestName)
-	versionPadding := fmt.Sprintf("%%-%ds", longestVersion)
+	versionPadding := fmt.Sprintf("%%%ds", longestVersion)
 	numberPadding := fmt.Sprintf("%%%dd  ", len(fmt.Sprintf("%v", len(u))))
 
 	for k, i := range u {
@@ -184,7 +184,7 @@ func (u upSlice) print() {
 
 		fmt.Printf(namePadding, i.StylizedNameWithRepository())
 
-		fmt.Printf("%s -> %s\n", fmt.Sprintf(versionPadding, left), right)
+		fmt.Printf("%s -> %s\n", fmt.Sprintf(versionPadding, left), fmt.Sprintf(versionPadding, right))
 	}
 }
 
--- a/print.go
+++ b/print.go
@@ -164,17 +164,18 @@ func (u upgrade) StylizedNameWithRepository() string {
 
 // Print prints the details of the packages to upgrade.
 func (u upSlice) print() {
-	longestName, longestVersion := 0, 0
+	longestName, longestVersionLeft, longestVersionRight := 0, 0, 0
 	for _, pack := range u {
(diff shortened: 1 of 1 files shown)


Morganamilo:
Right alignment makes sense for regular numbers, although in this case each line is a different package with a different version format, so I don't really know if your justification makes sense.

I can't tell what I prefer, I think it looks weird like this but that could just be because I'm used to the old way. So I'll reserve judgement for now.

As a slight bug though, padding is way too much for devel updates
```
:: 2 Packages to upgrade.
2  aur/chromium-widevine          1:1.4.9.1088-1 ->        1:4.10.1146.0-2
1  devel/latte-dock-git   0.7.96.r37.gec9ad19c-1 ->          latest-commit
```

Jakeler:
> each line is a different package with a different version format

This is the reason why i prefer right alignment, the only thing we really know for sure is the pkgrel at the end. Also many have a single digit minor version. Of course there are still a lot of cases where it doesn't help further, but then i see no disadvantage comapared to the old style.
Might be just my personal preference though, would be interesting to read other opinions.

> As a slight bug though, padding is way too much for devel updates

Yes, i have changed it now to do the padding on each side independently.


Btw, the alignment looks off when updating with `--timeupdate` (also without this patch):
```
5  community/electron             2.0.2-1  -> 2.0.6-1
4  aur/brightnessctl              0.3.2-1           -> 0.3.2-1
3  aur/jupyterlab                 0.33.4-1 -> 0.33.6-1
```

Morganamilo:
>This is the reason why i prefer right alignment, the only thing we really know for sure is the pkgrel at the end. Also many have a single digit minor version. Of course there are still a lot of cases where it doesn't help further, but then i see no disadvantage comapared to the old style.
Might be just my personal preference though, would be interesting to read other opinions.

If you want to read all the version numbers, it Is a lot easier to do it with left alignment. In the same way you would read a shopping list.

Also you can't assume the pkgver is one digit:
```
:: 7 Packages to upgrade.
7  aur/aurutils                            1.5.3-9 ->        1.5.3-10
6  aur/chromium-widevine           1:4.10.1146.0-1 -> 1:4.10.1146.0-2
5  devel/luvi-git             2.8.0.r20.g1ec62db-1 ->   latest-commit
4  devel/menuconfig-git             r801.54a3824-1 ->   latest-commit
3  devel/python-acme-git   0.24.0.r74.g09a28c7a2-1 ->   latest-commit
2  devel/python2-acme-git  0.24.0.r74.g09a28c7a2-1 ->   latest-commit
1  devel/uefi-shell-git         24434.3a0c1bf64b-1 ->   latest-commit
```

Also check out the pkgrel here https://www.archlinux.org/packages/multilib/x86_64/lib32-fontconfig/

>Btw, the alignment looks off when updating with --timeupdate (also without this patch):

Fixed

Jguer:
Both look okay for me, the new proposed one maybe looks slightly better, at least in the case presented

Jguer:
screenshot from 2018-09-15 20-01-34

Master is on the right, personally I prefer it, since I'm more keen on looking on the new version then the old.

Issue #621, opened 2018-08-04
yay fails when trying to install AUR package


LaserEyess:
#### Affected Version
```
yay -V
yay v8.998.r23.g6515a9b - libalpm v11.0.1
```

#### Issue
As the title says, I get the following when trying to do `yay -S rofi-pass`

```
yay -S rofi-pass
panic: runtime error: index out of range

goroutine 1 [running]:
main.splitDep(0xc42026ebe5, 0x1, 0x7ffdae8decd4, 0x9, 0x0, 0x0, 0x0, 0x0)
	/tmp/makepkg/yay-git/src/yay/dep.go:60 +0x100
main.provideSatisfies(0xc42026ebe5, 0x1, 0x7ffdae8decd4, 0x9, 0x7ffdae8decd4)
	/tmp/makepkg/yay-git/src/yay/dep.go:75 +0xaa
main.(*depPool).findSatisfierAurCache(0xc4204b2000, 0x7ffdae8decd4, 0x9, 0x9)
	/tmp/makepkg/yay-git/src/yay/depPool.go:423 +0x63a
main.(*depPool).resolveAURPackages(0xc4204b2000, 0xc4200550c0, 0x1, 0x0, 0x1)
	/tmp/makepkg/yay-git/src/yay/depPool.go:282 +0x204
main.(*depPool).ResolveTargets(0xc4204b2000, 0xc420486660, 0x1, 0x1, 0x0, 0xc420089050)
	/tmp/makepkg/yay-git/src/yay/depPool.go:157 +0xc7b
main.getDepPool(0xc420486660, 0x1, 0x1, 0xc42009e550, 0x0, 0x0, 0xc4200ba000)
	/tmp/makepkg/yay-git/src/yay/depPool.go:362 +0x7e
main.install(0xc4200a01c0, 0xc420055d38, 0x2)
	/tmp/makepkg/yay-git/src/yay/install.go:113 +0x476
main.handleSync(0xc42009a458, 0x1)
	/tmp/makepkg/yay-git/src/yay/cmd.go:273 +0x472
main.handleCmd(0x0, 0x0)
	/tmp/makepkg/yay-git/src/yay/cmd.go:152 +0x32c
main.main()
	/tmp/makepkg/yay-git/src/yay/main.go:239 +0x2b3
```
#### Steps to reproduce
`yay -S rofi-pass`

[...]

PR #622, opened 2018-08-04
Fix panic on invalid version strings

ZenixTheHusky:
Fixed issue #621. Fixed out of bounds error that causes yay to panic when splitting invalid version strings.


Morganamilo:
It looks like things like `provides('=')` now error out when trying to build with makepkg. So hopefully something like that never pops up again. Although there can still be old orphans lurking out there. And if they do pop up it's nice to not panic.

PR #624, opened 2018-08-05
Give an example of how to update only AUR packages

Morganamilo:
I've seen `yay -Syua` given as an example all over the place. I'm not
sure if this is left over as a habbit from yaourt. In yay, -a restricts
the upgrade to AUR packages ONLY while I believe in yaourt it means to
also include the AUR.

The problem with `yay -Syua` is that you are doing a database refresh
and then only upgrading the AUR packages. Essencially:

pacman -Sy && update-aur

This is a recipy for partial upgrades.

Either upgrade everything `yay -Syu` or just the AUR `yay -Sua`.

### DIFF
--- a/doc/yay.8
+++ b/doc/yay.8
@@ -474,6 +474,10 @@ yay \-Syu
 Update package list and upgrade all currently installed repo and \fBAUR\fR.
 
 .TP
+yay \-Sua
+Update all currently installed \fBAUR\fR packages.
+
+.TP
 yay \-S \fIfoo\fR
 Installs package \fIfoo\fR from the repos or the \fBAUR\fR.
 


polygamma:
Why not simply leave out the `-y` part, when one calls `yay -Syua`?

Morganamilo:
One of the most baffling places I found this advice is the manjaro wiki. I would correct this myself, but it seems as if you can't just go and make an account, you must ask the admins for one (really)?

Anyway I know @excalibur1234 has edited the wiki before and is the only name I recognise. If you would edit it for me I would great full.

Morganamilo:
@polygamma I prefer the program do what the users tell it to, the users just need to tell it to do the right thing. Although if it continues like this I'll consider it.

polygamma:
On the other hand: If users do not know what they are doing, and you know better, you may teach them how to do it right with e.g. a warning. There's basically no reason to actually call `yay -Syua`, so it's a sensible thing to give at least a warning, even if executing the `-y`. If one really wants the `-Sy` part, the user may simply call `pacman`, no AUR helper needed for that.

PR #625, opened 2018-08-05
Update bash completion

### LINKED ISSUE #616: Completion stuck for a long time if no prefix of pkg name
#### Affected Version

- yay version: yay v8.998 - libalpm v11.0.1
- my shell is `bash`

#### Issue
Pressing TAB twice after `yay_-S_` (where _ = space) cause a long not responding time.
TAB after any letter (e.g. `yay_-S_g`) can respond in resonable time.

I tend to press TAB a lot (since completion makes no harm). It is not a good experience that an extra TAB make things stuck.

If this is due to large data to download (60000+ possibilities), it might be better to disable completion when there is no leading letter.

My network condition
```
$ ping aur.archlinux.org
PING aur.archlinux.org (5.9.250.164): 56 data bytes
64 bytes from 5.9.250.164: icmp_seq=0 ttl=46 time=375.987 ms
```

Morganamilo:
Update bash completion, based of the current pacman bash completion.
Organized to try and keep the yay additions seperate from the main
pacman stuff and unused functions removed.

Don't complete packages for empty string

Fixes #616

### DIFF
--- a/completions/bash
+++ b/completions/bash
@@ -1,4 +1,3 @@
-# vim:fdm=marker foldlevel=0 tabstop=2 shiftwidth=2 filetype=bash
 # This file is in the public domain.
 
 _arch_compgen() {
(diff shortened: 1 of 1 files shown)


Morganamilo:
I still have no idea how to tell bash to suggest nothing. Suggesting files is still better than locking up until bash is killed. So I'll probably merge this and leave it up to somebody else to submit the full fix.

Jguer:
Maybe it's for the best

PR #626, opened 2018-08-07
Support the NO_COLOR environment variable

qjcg:
Support the `NO_COLOR` environment variable for disabling colorized
output.

From http://no-color.org/ :

> By adopting this standard, users that prefer to have plain,
> non-colored text output can set one environment variable in their shell
> to have it automatically affect all supported software.

### DIFF
--- a/main.go
+++ b/main.go
@@ -162,7 +162,7 @@ func initAlpm() (err error) {
 	value, _, _ = cmdArgs.getArg("color")
 	if value == "always" || value == "auto" {
 		useColor = true
-	} else if value == "never" {
+	} else if value == "never" || os.Getenv("NO_COLOR") != "" {
 		useColor = false
 	} else {
 		useColor = alpmConf.Options&alpm.ConfColor > 0


Morganamilo:
There's a little problem that the commands we call don't support this variable. The biggest example being the pacman calls. Do you think this is a pacman problem or something we should handle?

Jguer:
If pacman doesn't respect it, it will create a bit of inconsistency :/

Morganamilo:
And git, makepkg, editor, ect.

qjcg:
I see your point about wrapped commands.

On the other hand, there is certainly colorized printing that `yay` does on its own (ex: `yay -Ps`).

Perhaps a note could be added along the following lines:

"`yay` supports the `NO_COLOR` environment variable for disabling colorized output it produces. However wrapped commands (pacman, git, ...) may require further configuration to fully disable colored printing."

Morganamilo:
Alright if that's how you wanna do it, you might wanna add that to the man page as part of this PR.

Also If you want support in pacman consider talking about it at https://bugs.archlinux.org/ or `pacman-dev@archlinux.org`.

PR #630, opened 2018-08-09
Add --needed config option

c0d3z3r0:
Add --needed config option

### DIFF
--- a/doc/yay.8
+++ b/doc/yay.8
@@ -327,6 +327,14 @@ only GitHub packages are supported.
 Do not check for development packages updates during sysupgrade.
 
 .TP
+.B \-\-needed
+Pass --needed option to pacman by default.
+
+.TP
+.B \-\-noneeded
+Do not pass --needed option to pacman by default.
+
+.TP
 .B \-\-gitclone
 Use git to download and update PKGBUILDs. PKGBUILDs previously downloaded
 using tarball will continue to use tarballs until the package is clean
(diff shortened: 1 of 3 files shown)


doc/yay.8
```diff
@@ -326,6 +326,14 @@ only GitHub packages are supported.
 .B \-\-nodevel
 Do not check for development packages updates during sysupgrade.
 
+.TP
+.B \-\-needed
+Pass --needed option to pacman by default.
```
AladW: What does "by default" mean here? Doesn't that only apply if you set it in the config file?

c0d3z3r0: Exactly. This is under "PERMANENT CONFIGURATION SETTINGS" so were indeed speaking about the config file

Morganamilo: They only become permanent when `--save` is used. Without that it will not be by default.

Morganamilo:
I don't think I'm a fan of incorporating pacman's flags into yay's config. If needed should be a config option it should be in pacman.conf. Although I don't see much of a use case anyway. The only time I ever use needed tends to be in group installs.

Also I'm afraid all this PR does is make needed not work. you've changed it from a pacman flag to a yay flag. So instead of passing it to pacman, yay now consumes it when setting the config option. But you don't do anything with the config option, it just kind of exists there, doing nothing.

c0d3z3r0:
Ouch. I just realized I always used --needed but completely misunderstood (YES, I read the man page ;) ) what it does. "group installs" was the missing hint... Thank you!

PR #637, opened 2018-08-15
Refactor install

Morganamilo:
Merged depOrder.Aur and depOrder.Bases, fixed some bugs along the way.

This allows the functions in install.go more easily be used elsewhere. Notably, using -G to download pkgbuilds just like how -S does. Which will be done in another PR.

See commits for details.


Jguer:
Haven't reviewed but I approve the concept

PR #613, opened 2018-08-02
 Parallelize pkgbuild download and use MultiError

Morganamilo:

PR #641, opened 2018-08-18
 Use goroutines for fetching packages via -G

Morganamilo:
Continuation of #613. -G is now also run in parallel and uses the existing AUR pkgbuild functions.

### DIFF
--- a/doc/yay.8
+++ b/doc/yay.8
@@ -31,7 +31,8 @@ Perform yay specific print operations.
 
 .TP
 .B \-G, \-\-getpkgbuild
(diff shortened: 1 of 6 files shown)


Morganamilo:
I looked into getting rid of the `mv` command. It seems that os.Rename doesnt work across devices and go does not provide os.Copy or anything similar and it's probably not worth writing a recursive copy function.

Although if/when we finally move to tar.Reader+gzip.Reader instead of the call to tar, then we could selectively write the files we want, where we want. No need to move the files after.

Issue #627, opened 2018-08-08
Cannot use yay after renaming the username


Borderliner:
#### Affected Version
All

#### Issue
Cannot use `yay` after renaming my user, even after removing it, I recompiled and installed it again, but it fails to run, throwing the following error:
```
$ yay -V
Unable to create BuildDir directory:
/home/[old_username]/.cache/yay
The error was:
mkdir /home/[old_username]: permission denied
```
#### Steps to reproduce
1. sudo usermod -l newUsername oldUsername
2. sudo usermod -d /home/newHomeDir -m newUsername
3. yay -V

I can confirm that I have a home directory with the new username, and the old one is removed completely. Please fix this, thank you.

Jguer:
What is the output of `yay -Pg`?

Morganamilo:
Build dir is usually `/home/user/.cache/yay/`. I don't I need to see the output to guess that this is the problem.

```
yay --builddir $home/.cache/yay --save
```

Will fix it.

This is just a downside of the config system. A better way would be to have builddir to stay as an empty string in the config and have that imply the default. In the same way that makepkgconf stays as an empty string, implying the default.

Morganamilo:
Also for my command I think stuff happens in a weird order, so yay bails out before the --builddir can take effect. You may have to manually edit the config file `~/.config/yay/config.json`.

Borderliner:
@Morganamilo Awesome! Manually changing the path from `~/.config/yay/config.json` fixed this. It would be great if yay could update this path automatically. Thank you.

PR #678, opened 2018-08-31
Expand environment variables in config

### LINKED ISSUE #639: Yay should respect ~ for $HOME in config file
Version is yay v8.998 - libalpm v11.0.1

Yay is one of the few programs which does not respect a '`~`' as an alias for `$HOME` for file paths specified in `~/.config/yay/config.json`. So, for example, you have to hard code the path `/home/mark/.cache/yay/` for `buildDir` instead of being able to specify `~/.cache/yay/`. When I copy my config files to other users (e.g. on other machines) I have to remember to manually edit yay's config for the new path.

BTW, a separate issue I know but `buildDir` should default to the above, like pacaur does.

Morganamilo:
All string variables are expanded because why not? Would be weird to have it just for builddir.

`-Pg` and `-Pd` will show the expanded versions of the config, a little annoying but no big deal. `--save` will correctly keep the variables though which is the important thing.

Fixes #639 #627

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -132,6 +132,8 @@ func handleCmd() (err error) {
 		config.saveConfig()
 	}
 
+	config.expandEnv()
+
 	if cmdArgs.existsArg("h", "help") {
 		err = handleHelp()
 		return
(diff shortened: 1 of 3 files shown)


Jguer:
It looks ok, the config system keeps growing in an unhealthy way but I see no way around this in the current system

Morganamilo:
I don't mind code like this, it's the same thing over and over so it's readable even if it is a lot of lines. You might be able to achieve the same with less lines using reflection but I hate reflection.

Jguer:
Yes, I thought about reflection, but I really want to avoid importing reflection into `yay`

Issue #663, opened 2018-08-27
Is it possible to install all dependencies of a package, but not the package itself?


stevenroose:
Didn't find an IRC channel or discussion forum. Sorry for the question.

Morganamilo:
You can always grab the dependencies via -Si and install with -S --asdeps. But apart from that, not really.

stevenroose:
Hmm, there is no way to get the deps from -Si alone without the rest?

Morganamilo:
Oh yeah silly me, didn't think about recursive dependencies. You could `yay -S pkg` and it will tell you which repo/aur make/runtime deps you need. Although this could be an x y problem. What exactly do you want to achieve? Maybe aurmansolver is what you want?

stevenroose:
I wanted to install a different version from sources, without having to find out what all the dependencies were. The info is already there.

Morganamilo:
Why not just edit the pkgver during the install?

Morganamilo:
Closing because It's not something I really want to implement. This could be achieved via the `--print` flag if someone implemented it for AUR packages.

`yay -Sp pkg | filter out pkg`

Issue #652, opened 2018-08-23
Exit codes differ from pacman


EvanPurkhiser:
#### Affected Version
yay v8.998.r80.g0a136a3 - libalpm v11.0.1

#### Issue

Exit codes differ from pacman. I'm trying to use yay with `makepkg` by setting the `PACMAN` environment variable. But when makepkg tries to test for dependencies it just immediately fails with

```
==> ERROR: 'yay' returned a fatal error (1): <some-package>
```

#### Steps to reproduce
```
$ pacman -T invalid
invalid
$ echo $?
127
$ yay -T invalid
invalid
$ echo $?
1
```

Morganamilo:
We are not pacman. Yay can fail internally, as can makepkg, pacman, git, ect. So returning error codes becomes a little pointless when you don't know what program the error came from.

Although for cases were yay wraps purely pacman it would make sense. Using yay in makepkg certainly is a nice idea. Unfortunately makepkg tries to run yay as root and it fails.

EvanPurkhiser:
> Using yay in makepkg certainly is a nice idea. Unfortunately makepkg tries to run yay as root and it fails.

Realized this after opening this :)

Seems like this is probably still a valid issue.

Jguer:
#1095 supersedes this

Issue #69, opened 2017-12-08
Run makepkg as non-root when yay is run with sudo or as root


lectrode:
Currently in `apacman` you have the ability to install or update AUR packages as root. It gets around the inability to run `makepkg` as root by creating a new user ("aurbuild") and running makepkg as that user.

Snippet from `apacman` code for reference:
```
  # Installation (makepkg and pacman)
  rm -f *$PKGEXT
  if [[ $UID -eq 0 ]]; then
    id aurbuild >/dev/null 2>/dev/null || useradd -r -d /var/empty aurbuild
    mkdir -p /var/empty/.gnupg
    chown -R aurbuild:aurbuild . /var/empty/.gnupg

    if [[ $validpgpkeys ]]; then
      su aurbuild -c "gpg --refresh-keys ${validpgpkeys[*]}"
      keyfound=$(su aurbuild -c "checkpgpkeys ${validpgpkeys[*]}" 2>/dev/null)
      if ! [[ $keyfound ]]; then
        askpgpkeys "${validpgpkeys[*]}"
        su aurbuild -c "addpgpkeys ${validpgpkeys[*]}"
      fi
      su aurbuild -c "makepkg $MAKEPKGOPTS -f"; buildcheck=$?
      [[ $keyfound ]] || su aurbuild -c "removepgpkeys ${validpgpkeys[*]}"
      [[ $purgekeys ]] && [[ $keyfound ]] && su aurbuild -c "removepgpkeys ${validpgpkeys[*]}"
    else
      su aurbuild -c "makepkg $MAKEPKGOPTS -f"; buildcheck=$?
    fi
```

This has proven to be immeasurably valuable in my work, and the main reason I'm still using `apacman` instead of any other AUR package manager (despite it's apparent abandonment). This functionality allows me to completely update the systems I manage remotely without having to touch each one individually, save for the occasional manual update steps, which I'll discover and script before updating the rest of the systems.

Would it be possible to add this functionality to `yay`?

Jguer:
It is possible to implement this functionality in yay. This would involve allowing yay to run as root so I'll still have to go over the current logic too see what else will get broken by this.

lectrode:
That would be awesome. I'm happy to test any changes and provide feedback as well.

julianxhokaxhiu:
I really would like to have this implemented. It was really an awesome feature with apacman, and as long as I used it in those last 2 years, I never had a draw-back.

takaomag:
+1

[...]

PR #679, opened 2018-09-03
Allow running yay via sudo

Morganamilo:
This commit allows running yay via sudo. It does not however let you run
yay as root. Instead the SUDO_UID and SUDO_GID are used to sudo back
into the previous user that originally called yay.

This could allow for `PACMAN=yay makepkg` as a way to build packages
with makepkg while having dependency resolution for the AUR.

The return code problem would also need to be solved for this to work.

This could help with use cases such as #663 #652 and partially implements #69.

Not too sure if this is a good idea. Just an idea I had and it's literally 8 lines of code. Opinions welcome.

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
@@ -195,6 +196,13 @@ func cleanup() int {
 func main() {
 	if 0 == os.Geteuid() {
 		fmt.Println("Please avoid running yay as root/sudo.")
+	}
+
+	if 0 == os.Geteuid() && os.Getenv("SUDO_UID") != "" && os.Getenv("SUDO_GID") != "" {
+		args := []string{"-g", "#" + os.Getenv("SUDO_GID"), "-u", "#" + os.Getenv("SUDO_UID")}
+		args = append(args, os.Args...)
+		exitOnError(show(exec.Command("sudo", args...)))
+		os.Exit(cleanup())
 	}
 
 	exitOnError(setPaths())


Jguer:
It seems like a very simple and nice solution to a complex problem, I like it, although I would like @eli-schwartz opinion because last time I messed with yay and sudo he was right on all points

eli-schwartz:
As far as I understand, the sole purpose of this is in order to let people use yay to build a PKGBUILD, rather than an AUR package namespace. Wouldn't it be a lot more logical, then, to allow yay to build a local package?

For example, yaourt has the `-P <directory>` option to build a local PKGBUILD using AUR dependency resolution.

Morganamilo:
>As far as I understand, the sole purpose of this is in order to let people use yay to build a PKGBUILD, rather than an AUR package namespace.

That as well as to somewhat answer all the requests for allowing yay to run as root.

>Wouldn't it be a lot more logical, then, to allow yay to build a local package?

I don't know, maybe? I know that's what yaourt and some others do, but I don't know if that makes it more logical.

I mostly just found this approach interesting. I have yet to see if it's actually useful. It will be a while before I decide on merging it.

Jguer:
I think if it works for the sudo case it's mergeable material.

> I mostly just found this approach interesting. I have yet to see if it's actually useful. It will be a while before I decide on merging it.
:+1: 

Off-topic: Have you seen binary size increases in go1.11, it's crazy

eli-schwartz:
Except do you actually want to encourage people to do anti-patterns?



> Off-topic: Have you seen binary size increases in go1.11, it's crazy

How can you possibly tell the difference at this point...

Morganamilo:
>Except do you actually want to encourage people to do anti-patterns?

Personally no, that's why I put it out here to see if it was useful in any way.

>How can you possibly tell the difference at this point...

Well the package is ~20% (78mb) bigger so it's probably not too hard to notice.

eli-schwartz:
> Well the package is ~20% (78mb) bigger so it's probably not too hard to notice.

FWIW I've just tried building this and it's gone from 6.5 MB to 7.5 MB, when using the non-default go-pie compiler, and 5 MB to 5.7 MB when using a non-pie toolchain.

This trend of increasing bloat is more noticeable for some of my other binaries which have grown from 76 MB to 88 MB.

Jguer:
I was referring the compiled project binary sizes, I saw the `yay` package increase from `4.95 MiB` to `5.70 MiB`, even after stripped. Although to the trusting non go developer I would suggest installing `yay-bin` instead of pulling `go` from the repository which on my system is the largest single package

Jguer:
Back on topic, should we drop this, merge it or is there a better solution?

Morganamilo:
I haven't found this to actually be useful so I'll close it.

PR #682, opened 2018-09-04
 Use go-pacmanconf for config parsing

### LINKED ISSUE #680: Yay doesn't like includes in /etc/pacman.conf
I have an include in /etc/pacman.conf:
```Include = /etc/pacman.d/*.conf```

When this include is placed **above** the mirrorlist include(s), yay doesn't interact with the repo anymore. Searching and such doesn't work anymore. Updating and installing still works since that simply shells out to `pacman`.
When this include is placed **below** the mirrorlist include(s), everything works as expected.

### PR DESCRIPTION
**Pull request by Morganamilo**
This moves the config parsing from out of alpm and into the
go-pacmanconf libary plus some boilerplate code to get it into our alpm
config.

This makes sense as many config options such as UseColor and CleanMethod
have nothing to do with alpm and only relate to pacman.

pacman-conf is used instead of direct config parsing. This tool resolves
defaults and includes for us, so we don't need to handle it.

It is now safe to drop all the config parsing from go-alpm.

Should fix the broken Include handling That I've been ignoring for too long.
Fixes #680

### CODE REVIEW

File: config.go [hunk 1]
```diff
@@ -333,3 +321,112 @@ func (config Configuration) String() string {
...
+			ret |= alpm.UsageUpgrade
+		case "All":
+			ret |= alpm.UsageAll
+		}
+	}
+
+	return ret
+}
+
+func configureAlpm(conf *pacmanconf.Config) error {
```
Jguer: Passing the configuration to go-alpm and removing go-alpm's internal parser is not an option?

I would tag the current version of go-alpm as the last with an internal parser, import `go-pacmanconf` in `go-alpm` and avoid this alpm config in `yay, give me your thoughts on it

Jguer: Read your mid sentence comment here `It is now safe to drop all the config parsing from go-alpm.`, I guess we agree on this, just the boilerplate code is my question, wouldn't it still be useful for users of `go-alpm`

Morganamilo: Well The way I see it go-alpm should strive to just be go bindings for alpm. The config belongs to pacman and has nothing to do with alpm. For our use as a pacman wrapper it's useful to have config parsing.

With go-pacmanconf as a separate library you can still get config parsing without needed to write a ton of code.

Plus doing it this way, all out dependencies have no dependencies, it's just yay that depends on the individual pieces.

Jguer: Fair enough, I'll go and tag go-alpm still, so the parser removal is well documented

Morganamilo: I think you may have forgotten this isn't merged yet.

Morganamilo: Anyway I think this solution might actually be less code. It's +1000 but around 670 of those lines are the licence. Not sure now much the parsing code in alpm is though.

Morganamilo:
Added tests to this and go-pacmanconf. I've not used pacman-conf much so I ended up spotting a ton of little bugs (including a bug in pacman). Should be good to merge now.

Jguer:
We should keep new features in this release as small as possible as to only have a clean release of the new go-pacmanconf. Is it safe to gut go-alpm's config parser now?

Morganamilo:
Yeah feel free to delete config and config_test and maybe lint for more unused stuff. With the releases I've thought about using trunk based development. Would make it a lot easier to push bug fixes when there's a high volume of commits.

Jguer:
It could work, more than once we've had to push out semi ready features because of a bug fix

PR #684, opened 2018-09-04
Fixup some download bugs

Morganamilo:
Fix redownload logic
--redownload was reversed and redownloaded the deps instead of the
targets.

Limit download concurrency to 25 threads
Trying to download the ros packages, my system didn't like trying to download 250 packages at once. 

Currently it does it in "batches", so it fetches the first 25 packages, once they all finish it goes on to the next 25. Ideally it would launch a new thread as soon as one finishes. I don't really want to spend time on that and this works well enough. If there's an elegant solution I'd be happy to implement it.

### DIFF
--- a/depPool.go
+++ b/depPool.go
@@ -301,14 +301,12 @@ func (dp *depPool) resolveAURPackages(pkgs stringSet, explicit bool) error {
 			continue
 		}
 
-		//has satisfier installed: skip
-		_, isInstalled := dp.LocalDb.PkgCache().FindSatisfier(dep)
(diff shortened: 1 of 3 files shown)


Jguer:
Logic seems fine, I'm ok with including it in this hotfix if you are as well

Morganamilo:
It works fast enough, feel free to make a release.

Jguer:
Let's hope not ;), everyone already know our release window works in 1 long cycle and then 1 hotfix release

Issue #375, opened 2018-04-28
Revamp config system


Jguer:
### Origin
#373 

### Constraints
- Most Configuration libraries are overkill for `yay`

### Requirements
- Only output configuration file if default values are override
- Exported configuration must only contain overridden default values
- All relevant directories must be checked for existence and created on initialization.
- Configuration variables should only be set on `yay` init and `yay` should only read the values thereafter
- Easy to add new flags/config options
- Testing for the config system
  * Test for missing files
  * Test for missing directories
  * Test for illegal inputs on file
  * Test reading of a full, empty and semi-filled config file.

Morganamilo:
Does this include moving away from json and to something key pair?

Morganamilo:
I do prefer `key = value` over json for configs, I just use `--flag --save` anyway though it shouldn't matter that much.

One think I've wanted for a while but avoided because it breaks compatibility is separating the build files from pkg and package. So instead of `~/.cache/yay/name/PKGBUILD` It would be separated out to `~/.cache/yay/build/name/PKGBUILD` `and ~/.cache/yay/pkg/name.pkg.tar.xz`.

This gives more control over cache cleaning and will make #178 easy to implement with the keep last x version stuff.

Jguer:
I like it, I suppose it won't break as much as expected if moving config systems. All I do to configure yay is `yay -Syu --devel  --timeupdate --save` so underlying format is not so important to me but #373 (although technically not a problem of legacy configs) did show the limitations of the current system

Morganamilo:
Also if we do anything that breaks stuff I'll put together a .install that warns the user that their config settings have been reset and they may need to be reconfigured.

Morganamilo:
Now that we depend on go-pacmanconf we can use the sub package https://github.com/Morganamilo/go-pacmanconf/blob/master/ini/ini.go to implement a pacman style config. With no extra dependencies.

Jguer:
`yay.conf` it is :) 
Want to handle it or want me to do it, I should have a bit of time for it

Jguer:
I've started work on it, I was thinking with this new config system should we abandon `--save`, have a default `/etc/yay.conf` aka emulate pacman config entirely or implementing write equivalent to `go-pacmanconf/ini`

My `yay.conf` test sample
https://github.com/Jguer/yay/blob/reimplement-config/testdata/yay.conf

Feel free to suggest changes, I'm not sure I'm entirely happy with it

Jguer:
[...]

PR #703, opened 2018-09-13
add path expansion for '~' in 'config.BuildDir'

jshap70:
currently, setting `buildDir` in `~/.config/yay/config.json` to something like `"buildDir": "~/.cache/yay"` fails, as go doesn't expand the tilde to `$home`. This makes sense from a program standpoint, as the tilde only expands in shell and not in the environment, but to a user it is unclear why this setting would create a folder in the current directory named `~`.

Also, I am unashamed to admit when I was first starting out on linux I lost an install by running `rm -rf ~` because there was a local folder called `./~`, so I can personally attest that it is potentially dangerous for yay  to be making folders with this naming convention :confounded:

### DIFF
--- a/config.go
+++ b/config.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	pacmanconf "github.com/Morganamilo/go-pacmanconf"
@@ -184,6 +185,9 @@ func (config *Configuration) defaultSettings() {
 
 func (config *Configuration) expandEnv() {
 	config.AURURL = os.ExpandEnv(config.AURURL)
+	if strings.HasPrefix(config.BuildDir, "~"+string(os.PathSeparator)) {
+		config.BuildDir = filepath.Join("$HOME", config.BuildDir[2:])
+	}
 	config.BuildDir = os.ExpandEnv(config.BuildDir)
 	config.Editor = os.ExpandEnv(config.Editor)
 	config.EditorFlags = os.ExpandEnv(config.EditorFlags)


config.go
```diff
@@ -184,6 +185,9 @@ func (config *Configuration) defaultSettings() {
 
 func (config *Configuration) expandEnv() {
 	config.AURURL = os.ExpandEnv(config.AURURL)
+	if strings.HasPrefix(config.BuildDir, "~"+string(os.PathSeparator)) {
```
jshap70: this looks weird, but `os.PathSeparator` is a rune not a character, so it needs to be converted/cast to a string. using `filepath.Join()` is also an option here, but because it would need to be written as `filepath.Join("~", string(os.PathSeparator))` I decided there were no benefits over `+`.

Morganamilo:
What's the benefit over just using `$HOME` other than a little less typing? This implementation is also missing the ability to resolve the home dir of other users which one would expect to work. It also fails to expand when just using `~` because of the prefix check requiring a `/`.

jshap70:
> What's the benefit over just using $HOME other than a little less typing? 
it's not replacing the `$HOME` functionality, and like I said in the comment I just think its a little non-obvious 

> This implementation is also missing the ability to resolve the home dir of other users which one would expect to work. 
is that expected? if would resolve to whatever `~` is for the current user, which is probably where this config file is already. 

> It also fails to expand when just using ~ because of the prefix check requiring a /.
you're right, I'll go ahead and change that.

Morganamilo:
`~[user]` resolves to the home directory of user, or if user is not specified, the current user. Sure specifying a user is probably not too useful here but I would rather not implement a feature then have it work differently to how it works everywhere else.

If go had an `ExpandUser()` function I would be happy to use it, I would not however like to implement it here. I talked about this in #639.

PS. you need an empty line between the quote and your reply.

jshap70:
one thing we could do is query out to `readlink -f ...`, which could just automagically resolve (afaik) all paths for us, as long as they didn't rely on things `sh` wouldn't source. what do you think about that? 

PS: yeah I was replying on my phone and didnt remember if I did or not, and then couldn't fix it in time, lol

Morganamilo:
Now you're getting way ahead of yourself. Calling external commands for such little functionality is way too hacky (we currently call out to mv and tar and I hate it). Calling sh though is a whole other level and has lead to problems in the past https://github.com/Jguer/yay/issues/417.

I appreciate your effort but it comes down to lack of support in the stdlib. maybe suggest adding `ExpandUser()` to the stdlib? I for one would like to see it there.

jshap70:
sure, I'll take a swing at it :slightly_smiling_face:, I'll reopen something once it's taken shape

Morganamilo:
As I said in the last issue, I don't think it's worth implementing. Expanding environment variables already gives the functional.

Morganamilo:
PS. The reason I don't think it's work it because 99% of people are going to have `~` expanded by the shell anyway. The only real use for it is for the default builddir.

With #375 the main way of editing the config will be via the file instead of `--flag --save`. Where the shell can't expand '~'. I'd probably change my mind on this for the ini file.

jshap70:
that makes sense. for me, I was that 1% who just manually edited the config over of using the cli. but if it's changing in the future then yeah much better to wait. 

fwiw, I know you were joking about the stdlib thing, but I was bored and looked into how it works in python (https://github.com/python/cpython/blob/28658485a54ad5f9df52ecc12d9046269f1654ec/Modules/pwdmodule.c#L92). Actually looks pretty easy, just parsing `/etc/passwd`.

Morganamilo:
I wasn't joking about the stdlib thing. I do wish it was there.

Anyway there's no need to resort to horrible hacks. https://golang.org/pkg/os/user/#Lookup

PR #705, opened 2018-09-15
Convert to go modules

### LINKED ISSUE #704: Installing yay broken when GO111MODULE=on
#### Affected Version

```
$ yay -V
yay v8.1115 - libalpm v11.0.1
```

This is after I worked around the problem and upgraded. It's more about the build than the version of yay though, so it will still happen on this version upgrading to a newer one or when attempting to rebuild.

#### Steps to reproduce

1. `export GO111MODULE=on` – enable [Go modules] (even if we're in GOPATH)
2. `yay -Su yay` (or similar: attempt to upgrade yay from an older version anyways)

```
…
==> Starting build()...
go build -v -ldflags '-s -w -X main.version=8.1115' -o yay
go: cannot determine module path for source directory /home/sam/.cache/yay/yay/src/yay-8.1115 (outside GOPATH, no import comments)
make: *** [Makefile:46: build] Error 1
```

Adding an import directive (eg. `package main // import "github.com/Jguer/yay"`) should fix this. Alternatively, switching to go modules from dep (or adding a step to the build to generate go mod files from the dep files if you're not ready to switch entirely) would fix it.

[Go modules]: https://golang.org/doc/go1.11#modules

Morganamilo:
In go 1.11 modules are considered experimental. So it might be best to wait for 1.12 before merging. Also sadly there seems to be no support for pruning (hence the additions).

Fixes #704


Morganamilo:
Also I've changed the default prefix to `/usr/local`. I had been meaning to for a while. The pkgbuilds should have `PREFIX=/usr` after `DESTDIR` (they should have specified it from the start really).

Given that this change will currently not effect the built package you can do it without a pkgver bump.

Jguer:
Coincidentally I also tested `go mod` today and I think we should convert to the soon to be default tool. Also, being included in the toolchain I would finally remove the vendor directory from the repo and start tagging upstream according to semantic versioning.

It would be most prudent to wait for 1.12 but still I'm very hyped to merge this

Morganamilo:
So does that mean `yay 9.0.0`?

SamWhited:
If you want a temporary patch to make the build work on machines running Go 1.11, adding a package import path (`package main // import "github.com/Jguer/yay"`) will let the Go tool automatically generate a go.mod even if one doesn't exist. Since Yay will typically be built outside of GOPATH (I'm assuming it will be built from the PKGBUILD mostly), on Go 1.11 machines modules will be used by default (unless the user has disabled them entirely) so having a temporary fix would fix updates until the full support can be merged.

Jguer:
Yay doesn't need to transition to semantic versioning because it's not imported but if we want to change it, we should do it with this change

Morganamilo:
I checked the roadmap and it seems the 1.12 is due for January 2019 :/

You mentioned using this in production. So as long as you are comfortable I don't mind merging this soon.

SamWhited:
> You mentioned using this in production. So as long as you are comfortable I don't mind merging this soon.

FWIW I've been using go modules in production for a while too (including in several PKGBUILDs). It has a few bugs, but for the most part it's been faster and more stable than dep for me.

> Also sadly there seems to be no support for pruning (hence the additions).

I think you want `go mod tidy`

Morganamilo:
> I think you want go `mod tidy`

What I meant is pruning unneeded files, readmes, test files, etc.

I just checked this out again and tried to build. It gave me this output:
```
morganamilo@Vinyl modules ~/git/yay % make
go build -mod=vendor -v -ldflags '-s -w -X main.version=8.1160' -o yay
golang_org/x/crypto/cryptobyte/asn1
golang_org/x/net/dns/dnsmessage
golang_org/x/crypto/curve25519
github.com/Morganamilo/go-pacmanconf/ini
github.com/Morganamilo/go-srcinfo
github.com/Morganamilo/go-pacmanconf
github.com/jguer/go-alpm
golang_org/x/crypto/cryptobyte
golang_org/x/crypto/internal/chacha20
net
golang_org/x/crypto/poly1305
golang_org/x/text/transform
golang_org/x/crypto/chacha20poly1305
golang_org/x/text/unicode/bidi
golang_org/x/text/unicode/norm
golang_org/x/text/secure/bidirule
golang_org/x/net/http2/hpack
golang_org/x/net/idna
golang_org/x/net/http/httpproxy
net/textproto
crypto/x509
golang_org/x/net/http/httpguts
crypto/tls
net/http/httptrace
net/http
github.com/mikkeloscar/aur
github.com/Jguer/yay
```

Which is confusing, I have no idea why it's building all that.

Jguer:
```
 env GO111MODULE=on go mod tidy
go: finding github.com/Morganamilo/go-pacmanconf v0.0.0-20180910220353-9c5265e1b14f
go: finding github.com/mikkeloscar/aur v0.0.0-20180725140122-f998dbf94dc4
go: finding github.com/jguer/go-alpm v0.0.0-20180914002751-643c287316a5
go: finding github.com/Morganamilo/go-srcinfo v0.0.0-20180702014333-368edc79b2c5
go: downloading github.com/mikkeloscar/aur v0.0.0-20180725140122-f998dbf94dc4
go: downloading github.com/jguer/go-alpm v0.0.0-20180914002751-643c287316a5
go: downloading github.com/Morganamilo/go-srcinfo v0.0.0-20180702014333-368edc79b2c5
go: downloading github.com/Morganamilo/go-pacmanconf v0.0.0-20180910220353-9c5265e1b14f
```

I believe it's purely listing all dependencies

Morganamilo:
Okay so I've found out why it was building all that I guess. 

I knew that `golang_org/x/net/http2/hpack` and similar are outside of the stdlib but I guess the stdlib actually pulls them in as dependencies or something?

```
morganamilo@Octavia modules ~git/pacman % yay -Qo /usr/lib/go/src/vendor/golang_org/x/net/http2/hpack
/usr/lib/go/src/vendor/golang_org/x/net/http2/hpack/ is owned by go 2:1.11.1-1
```

So that at least shows it's not our projects pulling them in and downloading them somehow. And these `golang_org/x/` packages actually do show up on the current master if you use `go build -v -a`. I guess the modules mode is just more print happy or something.

So with that mystery out of the way, happy to merge this now?

Jguer:
Although I did find this in the `Go1.11.2` stack https://github.com/golang/go/issues/27285 but I didn't check it with attention enough.

Still reluctant and afraid as usual with this type of change but ok let's do it, even before the config parser merge so it can mature in `yay-git`. It shouldn't require PKGBUILD changing except for the minimum go version

Jguer:
Let's wait for 1.11.2, It musn't be that far anyway.

Well, manjaro is on 1.11 so we can consider everyone is on 1.11 now :laughing:

Morganamilo:
Manjaro like to place a decade old version of yay in their repos anyway. So even if we released now, go 1.12 would probably be out by the time they update.

PR #708, opened 2018-09-15
Allow building with bmake and don't rebuild unnecessarily

SamWhited:
These are just minor suggestions from glancing at the makefile; I won't be offended if they're not wanted :)

I'll leave comments inline with what everything does since this clutters things up a bit if you're not all that familiar with make.
Since this project is pretty specifically targeted at Arch, it's not likely that gnu make won't be available, but as an aside this also makes it possible to build with BSD make and similar (which is actually what I have as the default make on my machine).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,21 +1,22 @@
-.PHONY: all default install uninstall test build release clean
+.PHONY: all default install uninstall test build release clean package
 
 PREFIX := /usr
 DESTDIR :=
 
-ifndef VERSION
-MAJORVERSION := 8
-MINORVERSION ?= $(shell git rev-list --count master)
-endif
-VERSION := ${MAJORVERSION}.${MINORVERSION}
+
+MAJORVERSION$(VERSION) := 8
+MINORVERSION$(VERSION) != git rev-list --count master
+VERSION ?= ${MAJORVERSION}.${MINORVERSION}
 
 LDFLAGS := -ldflags '-s -w -X main.version=${VERSION}'
 ARCH := $(shell uname -m)
 PKGNAME := yay
 BINNAME := yay
 PACKAGE := ${PKGNAME}_${VERSION}_${ARCH}
+GOFILES != find . -name '*.go'
 
-export GOPATH=$(shell pwd)/.go
+CURDIR?=${.CURDIR}
+export GOPATH=$(CURDIR)/.go
 
 default: build
 
@@ -41,7 +42,9 @@ test:
 	go vet
 	go test -v
 
-build:
+build: ${BINNAME}
+
+${BINNAME}: $(GOFILES) Gopkg.toml Gopkg.lock
 	go build -v ${LDFLAGS} -o ${BINNAME}
 
 release: | test build
@@ -54,6 +57,7 @@ release: | test build
 
 package: release
 	tar -czvf ${PACKAGE}.tar.gz ${PACKAGE}
+
 clean:
 	rm -rf ${PKGNAME}_*
 	rm -f ${BINNAME}


Makefile
```diff
@@ -3,47 +3,54 @@
...
 DESTDIR :=
 
-ifndef VERSION
-MAJORVERSION := 8
-MINORVERSION ?= $(shell git rev-list --count master)
-endif
-VERSION := ${MAJORVERSION}.${MINORVERSION}
+
+MAJORVERSION$(VERSION) := 8
+MINORVERSION$(VERSION) != git rev-list --count master
```
SamWhited: Removing the ifndef makes this compatible with BSD make (which uses a slightly different syntax). It should also be compatible with any POSIX compliant make system now. This is probably less necessary than the other changes in this file since it's unlikely anyone will be using anything but GNU make on arch.

SamWhited: Changing to `!=` instead of the shell expansion means that this is evaluated once, and not every time the variable is used. Probably doesn't matter since it's only used once that I see here. It was just habit.

Makefile
```diff
@@ -3,47 +3,54 @@
...
+VERSION ?= ${MAJORVERSION}.${MINORVERSION}
 
 LDFLAGS := -ldflags '-s -w -X main.version=${VERSION}'
 ARCH := $(shell uname -m)
 PKGNAME := yay
 BINNAME := yay
 PACKAGE := ${PKGNAME}_${VERSION}_${ARCH}
 
-export GOPATH=$(shell pwd)/.go
+GOPATH=$(shell pwd)/.go
```
SamWhited: `export` isn't valid outside of GNU make, but it's unnecessary in either. This should be propagated into any subshells, and I think this behavior is the same for GNU make and BSD make.

Makefile
```diff
@@ -3,47 +3,54 @@
...
 
 default: build
 
+.PHONY: all
```
SamWhited: Setting the various rules that don't correspond to files to `.PHONY` means that make won't be trying to check if ifles with that name exist which could cause rules to not be run (eg. if I did `touch all` for some reason, the rule would fail before).

Makefile
```diff
@@ -3,47 +3,54 @@
...
+.PHONY: test
 test:
 	gofmt -l *.go
 	@test -z "$$(gofmt -l *.go)" || (echo "Files need to be linted" && false)
 	go vet
 	go test -v
 
-build:
+.PHONY: build
+build: ${BINNAME}
```
SamWhited: Kept for backwards compatibility.

Makefile
```diff
@@ -3,47 +3,54 @@
...
+
+${BINNAME}:
```
SamWhited: Prevent rebuild when the file already exists and is up to date. Really lots of `.PHONY:` rules is generally a bad sign in a makefile, I didn't want to touch too much though so I didn't extract out the generation of any other build artifacts except the main one.

Makefile
SamWhited: Interesting, let me double check and I'll add it back in if there's not a standard way to do this. Thanks!

Jguer: Could be a darwin thing (or super outdated bash thing). I think it's safe to remove it

Makefile
Morganamilo: > Really lots of .PHONY: rules is generally a bad sign in a makefile

I find this true when using C and such. In this case though `go build` already knows when to rebuild and when not to.

Makefile
SamWhited: Odd, I thought I tested this but apparently I'm completely wrong about the behavior. export does work in both, it's the `$(shell pwd)` that's not working in bsd make for some reason. I've updated it and tested with both `gmake` and `bmake`: the absolute path is correct in both now.

Makefile
Morganamilo: Also my preference is all the recipes under one .PHONY declaration. But that's about all I have to complain about as long as it builds.

SamWhited: That's a fair point; it still feels poor to shell out unnecessarily if you don't have to, but I suppose it matters less in Go where the build is smart and fast.

SamWhited:
An alternative since it's not likely that anyone but me isn't using GNUmake might be just to rename this to `GNUMakefile` (or whatever it is that GNU expects). That way if you're not using it your tool won't work or will know to shell out to the GNU version if you're using a smart wrapper around make instead of throwing a syntax error. This is all minor nit-picky stuff either way though, so take it all with a grain of salt.

Makefile
SamWhited: Oh gosh, was that there the whole time? I didn't even notice that they were already .PHONY, sorry about that. Fixing.

Morganamilo: Haha line 1 and all. Don't worry about it.

Morganamilo:
I don't disagree with the idea of this but now that I've actually read through it, it's clear it won't work.

`${BINNAME}` has no dependencies so there are no files that can be changes to convince make to rebuild.

SamWhited:
*facepalm* good point. Fixed.

Morganamilo:
So I'm testing this again, I assumed the `!=` was some sort of conditional but looking it up it's just used for shell commands. The point of the if was to stop the running of git if version is already set. For example When building from a tarball `make VERSION=1.1` could be done and the git call would be skipped.

This also shows that the yay build can be altered using `VERSION=` or `LDFLAGS+=` in the make command. The recipe doesn't understand this and will sill claim to be up to date. It's probably best to just let go handle knowing when to build or not.

Also could you explain `CURDIR?=${.CURDIR}`? I'm not quite sure how it works.

SamWhited:
> So I'm testing this again, I assumed the != was some sort of conditional but looking it up it's just used for shell commands. The point of the if was to stop the running of git if version is already set. For example When building from a tarball make VERSION=1.1 could be done and the git call would be skipped.

Ah, I see, yes, git will always be run in this case. The better way to do this would be to use the `$(shell)` expansion inside an assignment using `?=`, which I think will cause the shell expansion to only be run if the variable did not exist. I'll double check and fix that in a follow up if it works.

> Also could you explain CURDIR?=${.CURDIR}? I'm not quite sure how it works.

On BSD make the variable is called `.CURDIR` and on GNU Make it's `CURDIR`, so we're setting `CURDIR` (if it's not already set, I.E. we're on BSD make) to the value of `.CURDIR` so that we have one identifier to use throughout the file. If we're on GNU Make CURDIR will already be set and this won't do anything (and we can still use CURDIR throughout the file).

Morganamilo:
It appears that `!=` is the same as `:= $(shell ...)`. The `:=` causing the variable to be expanded immediately.

Just using `= ${shell ...}` fixes it. But then that turns out to be a GNU extenction... No idea if there's an equivalent way to do it in BSD's make.

Also with the `CURDIR` thing you should have spaces around the operator. That's why I got so confused over it.

>I forgot what a pain writing portable makefiles was

Yeah I'm starting to feel that now. Which is sad because I do like having things be portable.

SamWhited:
Closing this for now; sorry it got away from me.

Morganamilo:
Giving up eh? I don't blame you, make can be difficult to figure out. It's not all for nothing though, I'll merge some of the changes. It doesn't make it bmake comparable but it does at least improve it. 

Maybe we can follow pacman and move to meson :P

Jguer:
I have nightmares about most build systems already, specially cmake. I guess this adds make & bmake compatibility to the list

SamWhited:
Yah, sorry, I don't have time to get back to this right now. Thanks for bearing with me though, and of course feel free to use whatever chunks you'd like!

PR #717, opened 2018-09-22
Reimplement config

Jguer:
Reimplementation of config using the `ini` package and maps

TODO:

- [x] Remove the save system
- [x] Decide a `yay.conf` to package by default /etc/yay.conf
- [x] Reimplement `-Pd` and `-Pg`
- [x] To achieve a new system where to add a new config option all you need to do is add it to defaultSettings move all keys to lowercase and use that to parse `yay.conf` and the command-line options, this will rename some cli options.
- [ ] Man page writing
- [ ] Update completions
- [x] Add warning for users who still have yay.json

To debate:
- Removal of some underused flags from config and cli like `--timeupdate` and maybe `--nogitclone`


Morganamilo:
I still don't get the point of having maps. We know all the valid options ahead of time, maps should be for datasets that can change in size during runtime. Also the great thing about structs is they will give us a compile error if you typo the option name.

You might want to consider adding this somewhere:
```
if config.json exists then
    print "Yay has moved to a pacman style config system. If you have configured yay
    you may want to configure yay.conf to match the settings you previously used.
    Delete config.json to remove this warning."
end if

Do you wish to continue?
```

I get removing tarball support although time update can be handy when used with redownload and rebuild.

Morganamilo:
Here's what I would do for a struct based parser. It's incomplete but just to show the type of thing I'd go for.

```go
package main

import (
	"fmt"
	"io/ioutil"
	"os"

	"github.com/Morganamilo/go-pacmanconf/ini"
)

func parseCallback(fileName string, line int, section string,
	key string, value string, data interface{}) (err error) {
	if line < 0 {
		return fmt.Errorf("unable to read file: %s: %s", fileName, section)
	}
	if key == "" && value == "" {
		return nil
	}

	conf := data.(*Configuration)
	value = os.ExpandEnv(value)

	switch section {
	case "options":
		return setOption(fileName, line, section, key, value, conf)
	case "answer":
		return setAnswer(fileName, line, section, key, value, conf)
	default:
		return fmt.Errorf("%s:%d: directive %s is not in a valid section", fileName, line, key)
	}
}

func setOption(fileName string, line int, section string,
	key string, value string, conf *Configuration) error {

	switch key {
	case "AURURL":
		conf.AURURL = value
	default:
		return fmt.Errorf("%s:%d: unkown directive %s in section %s", fileName, line, key, section)
	}

	return nil
}

func setAnswer(fileName string, line int, section string,
	key string, value string, conf *Configuration) error {

	switch key {
	case "Clean":
		conf.AnswerClean = value
	case "Diff":
		conf.AnswerDiff = value
	case "Edit":
		conf.AnswerEdit = value
	case "Upgrade":
		conf.AnswerUpgrade = value
	default:
		return fmt.Errorf("%s:%d: unkown directive %s in section %s", fileName, line, key, section)
	}

	return nil
}

func parseConfig(file string) (*Configuration, error) {
	iniBytes, err := ioutil.ReadFile(file)

	if err != nil {
		return nil, fmt.Errorf("Failed to open config file '%s': %v", file, err)
	}

	conf := &Configuration{}
	err = ini.Parse(string(iniBytes), parseCallback, conf)

	return conf, err
}
```

Jguer:
For me the map based approach allows you to add a new option to the defaultconfig and knowing it's configuration and command line flag will be automatically created (to be done) in a way that is only possible with reflection in a static struct. 

The not knowing if you typoed a config variable is a very big concern (although making them all lowercase has helped)

I guess it boils down to being more tailor made or more general purpose (viper uses maps as well understandably since it doesn't know flags beforehand)

Morganamilo:
>For me the map based approach allows you to add a new option to the defaultconfig and knowing its configuration and command line flag will be automatically created (to be done) in a way that is only possible with reflection in a static struct.

I don't really see how you'd do this. We have the `rebuild` config option and the flags controlling it being `--rebuild --rebuildall --norebuild`. Also you'd need to mark if a flag takes an argument.

Also the struct will no longer be tied to json so I we could dump all the non savable flags in there too. For example the variable "mode" can be moved inside the struct.

Jguer:
For the flags, I would take this approach:
In the config file it's `Rebuild = no` therefore it will be saved as `config.value["rebuild"] = "no"`, the flag will subsequently be `--rebuild no`, `--rebuild=all` or `--rebuild=yes`.

When you check for the `config.value["rebuild"]` in usage you check for `all`, `yes` and any other value defaults to `no`.

The parser will check if the key exists in one of the maps, if it exists in the value map you take the value, if it exists in the boolean map you can either take in the value of `yes` and `no` or we can respect some of the old notation, remove the prefix `no` and check key existence and then set the boolean value with that.

Con: you can't really warn the user for valid values, but it won't break their experience.

Morganamilo:
I don't really want to break compatibility by changing the flag syntax. Also looking up if its a valid value and defaulting if not sounds like it could easily break when some one tries to install a package with the same name as a potential value.

You could take the more standard approach of `--rebuild` meaning `--rebuild=yes` and any other values requiring an explicit `=`.

And of course with the struct no longer being json the type for `rebuild` and such flags can move from a string to an enum.

Jguer:
> You could take the more standard approach of --rebuild meaning --rebuild=yes and any other values requiring an explicit =.

Seems workable to me

> And of course with the struct no longer being json the type for rebuild and such flags can move from a string to an enum.
But  if they are turned into enums they will require specific unpacking or reflection, in the map system or struct system

Morganamilo:
>But if they are turned into enums they will require specific unpacking or reflection, in the map system or struct system

Not really sure what you meant here. What I meant is an enum similar to the targetMode type.

Morganamilo:
About the latest commit we could just drop current config printing. For the default printing we could just compile the config file in. 

```
-ldflags -X main.defaultConfig=`cat yay.conf`

function print_default_config()
    print defaultConfig
end function
```

Jguer:
@Morganamilo Take a look at the new handle config maybe it will be more clear but what I meant about unpacking with enums is this:
```
case "rebuild":
if value == "all"{
config.rebuild = all
} else if value == "no" {
config.rebuild = no
} else {
config.rebuild = yes
}
```
Instead of just 
```
case "rebuild":
config.rebuild = value
```

The new implementation standardizes flags. if it expects values then it has no `norebuild` type options, it's all `rebuild=` but if it is a boolean `nosudoloop` still works which doesn't completely break our flags but standardizes them

Jguer:
Missing remains `is-arg` to follow the same logic as the option parser and it should be alpha quality (and then adding the warnings, the file checks and the best part, checking if it runs smoothly with new configs, old configs, missing dirs, full dirs, half-full dirs, half-empty dirs and invalid files :laughing:)

Jguer:
Usual off-topic: After this my objective would be to implement the localization and after that porting to gomodules, feedback encouraged

Morganamilo:
>porting to gomodules

Well that's basically already one :P

Jguer:
Haha, talking about the merge order because we said we would wait a bit before introducing it.
Maybe config - release and then localization + gomodules release?

Jguer:
You're right, I'll do it tomorrow to avoid nasty surprises

Morganamilo:
When you do, I'd suggest adding `DESTDIR="$pkgdir" PREFIX=/usr` to both make calls. It doesn't actually do anything (at the moment) but is better practice.

Morganamilo:
Also seems as you're changing how flags are parsed, maybe we can finally implement flags that can be passed multiple times?

- cachedir
- hookdir
- overwrtie

Jguer:
Yes, I'm touching up those parts as well. I'll check, today I should have this branch feature complete. Improve on it.
Also, next version is 9.0.0 I guess

Morganamilo:
Menu doesn't show unless I explicitly pass `--upgrademenu`.

That's kinda to be expected as my config is empty.

Jguer:
Fair point

Morganamilo:
Also your boolean map could be a stringSet ;) struct{} takes less memory than bool aswell.

Jguer:
> Also your boolean map could be a stringSet ;) struct{} takes less memory than bool aswell.

True, but if to make it false you need to delete the key it kind of ruins the current system in place

Everything should be working as suppose in this branch, time for behaviour checks and seeing if the new "autogenerated" options are not too much of a pain. Also if we want to change any defaults, this would be a good time.

Morganamilo:
Here's some of the stuff I noticed reading through:

- The expandEnv() function loops through .value so it won't work on any of the other sections.
- I'm against to having a section called `[intoptions]`. Options should be placed where they logically belong not just because it's easier to implement it this way.
- You seem to have added a fallback if home/cache is not set. Setting this to `/tmp` will cause yay to dump everything right in `/tmp` right? It would be better to do `/tmp/yay-username`. I Believe go has a function to get the tmpdir too.
- You should check for both no section and an invalid section in parseCallack. Currently an invalid section will error as no section.
- Pacman has all of its options in upper camel case I would like ours to be the same for consistency. You lower them after the fact which makes the config file case insensitive. This means some one could enter a config flag in camel case but when they `-Pg` it comes out lower case.
- That warning for the .json file should definitely be in here before merging
- Same for the /etc/yay.conf handling

Jguer:
> The expandEnv() function loops through .value so it won't work on any of the other sections.

it shouldn't be necessary to expand numerical or boolean 

> I'm against to having a section called [intoptions]. Options should be placed where they logically belong not just because it's easier to implement it this way.

Easily fixable

> - You seem to have added a fallback if home/cache is not set. Setting this to /tmp will cause yay to dump everything right in /tmp right? It would be better to do /tmp/yay-username. I Believe go has a function to get the tmpdir too.
>- You should check for both no section and an invalid section in parseCallack. Currently an invalid section will error as no section.

TODO

> Pacman has all of its options in upper camel case I would like ours to be the same for consistency. You lower them after the fact which makes the config file case insensitive. This means some one could enter a config flag in camel case but when they -Pg it comes out lower case.

We should provide the file in upper CamelCase but after that I think it's a worthy compromise

> That warning for the .json file should definitely be in here before merging
> Same for the /etc/yay.conf handling

TODO

Morganamilo:
>it shouldn't be necessary to expand numerical or boolean

Sure but it's a little inconsistent. You could call expandEnv() on value inside of parseCallback(), that will catch every value and will remove the need for a loop.

Morganamilo:
One thing I forgot to mention is that like pacman, flags should be additive.

Jguer:
Can you give me an example of yay additive flags, I'm a bit confused on what you mean, I can only think of the pacman flags like `overwrite`

Morganamilo:
The flags append to the already existing options set by the config.

IgnorePkg = a b c

Passing --ignore d would keep the current entries and append d. This would apply to mflags and editorflags.

Jguer:
Only things missing are:
- .json present warning
- the addition flag stuff

Core functionality is working, it has no fs bugs that I see so we should be good for merge and adding on those small fixes. Have you had an opportunity to go through it @Morganamilo ?

Morganamilo:
What warning is a must, otherwise the -git users might be very confused.

Morganamilo:
Nothing wrong with rewriting history as long as it's not on a main branch. That said squashing all the commits into one is just as unreadable as having a ton of fixup commits littered around. The balence is some where in between :P

Might even be a good idea to make a new PR with 47ddaa51af01d69b5002469416156cdcb9bfb7f7, e8f1ca8ea9ed2990e14c56dcc004a445262f09cc and possibly others so that they can be merged now and it will cut down on the size of this PR.

Glad that it's rebased though. Now I can run yay with the new config and the new depSolver.

Morganamilo:
I finally got around to writing a more proof of concept struct based implementation. Maybe I can finally sway you on it lol.

https://github.com/Jguer/yay/compare/master...Morganamilo:struct-config

The config parsing here is pretty complete but it's not hooked up the the main code in anyway. 

The code is verbose but readable. The error handling is pretty great too

For example this config file:

```
[options]
Rebuild = Bar
Something
BuildDir
Devel = foo
```

Produces:

```
/etc/yay.conf:2: invalid value for 'Rebuild' : Bar
/etc/yay.conf:3: unknown option 'Something'
/etc/yay.conf:4: option 'BuildDir' requires a value
/etc/yay.conf:5: invalid value for 'Devel' : foo
```

Jguer:
These ones I can implement in the map system :innocent: 
```
/etc/yay.conf:3: unknown option 'Something'
/etc/yay.conf:4: option 'BuildDir' requires a value
```

It's not that I don't recognize the access performance and syntactic advantages of the struct based system but then having that huge parser case, another one for config and having to almost replicate code for each new added option. 
On the error handling, the only thing that pains me in the map approach is that all flags like `rebuild` need to be checked in if `all`, else if `blabla` and instead of `no` it needs to be an else because the user might have fed rebuild with something else besides those 3 options.

Morganamilo:
For me the thing I hate the most is having to use `config.boolean["combinedupgrade"]`. Compile time checks asides this is pretty horrible to read/write.

But seems as you're still set on this and I don't wanna touch anything until after this is done. If you're hard on time and don't mind I'll rebase and fixup your branch for you.

Also I'd like to restructure the config file to what I have in by example. Notably merging enabling a menu with giving a menu an answer and the [bin] section.

Jguer:
> But seems as you're still set on this and I don't wanna touch anything until after this is done. If you're hard on time and don't mind I'll rebase and fixup your branch for you.

If you have the availability for it, I think it would turn out even better as I think you have a slightly different view on the correctness of it.

Thinking of it, for runtime checks, `config.boolean["combinedupgrade"]` should be wrapped to check for the existence of the key at least, and it is horrible syntax indeed. I'm not happy with either and the least happy that the one that would make me happy is config structs with reflection :laughing:

Morganamilo:
Does that mean reflection is on the table then? I'm fine with it as a compromise between our two methods

Stuff like this already made it into commits: https://github.com/Jguer/yay/blob/e5f6e15c24c8e5ddb9feef51360b3ad8a5253ce5/parser.go#L549

Which won't raise a compile time or run time error.

The json stuff I would assume uses reflection so it's not like we're not already using reflection.

Jguer:
True, I would say reflection is on the table for this, if the cli option parsing and config option parsing can be programmatically determined instead of static determined (possibly in the same method even) I would be satisfied with it

Morganamilo:
I pushed my cleaned up version. I could make it neater but I'm gonna go look into a reflection based method so I pushed what I have.

The code is mostly the same, it's just the commits that have been tidied. There are minor tweaks though, such as fixing this to be lower case like it should https://github.com/Jguer/yay/blob/37adf5086137af1557fe9739545944e5e739a932/install.go#L787

Jguer:
Point taken indeed

Morganamilo:
I came up with https://github.com/Morganamilo/yay/blob/reflect-config/config_parser.go as a prototype. Way uglier and hackier (as expected with reflection) than the pure struct based one but this is at least less code and allows for new config options to be dropped in with a single line. 

On top of this, the flag system could pretty much be directly piped into this, all you would need is add some handling for the `no` prefix just like how your branch does.

The error handling is also just as good as before (even better now that it can tell you what values an option accepts).

The only problem I have is that non boolean `--no` flags like `--topdown --norebuild` would no longer work. `--norebuild` could be implemented via some `default:` tag, while --topdown is impossible to implement without adding special handling. Which is not hard to do, but this kind of goes against the attempt at having all the flags in one place and easy drop in.

That being said maybe there can be an alias list were you define a flag and the key/value it emulates. That way norebuild=`rebuilld=no` rebuildall=`reabuild=all`, topdown=`sortmode=topdown`.

Anyway what do you think of this approach?

Jguer:
It actually looks pretty good, it's very intuitive for adding new options.
I already did it in my branch, but I would remove `--topdown` and use only sortmode and I would move the rebuild flags to a single key,  `rebuild=*` as suggested and the same for `redownload=*`

Morganamilo:
I can accept it for rebuild/download, but for --topdown it adds so many keystrokes, which makes me to loose the shorter option.

But then again does anyone actually use --topdown/--bottomup after configuring it once? In fact does anyone even use --topdown? So I guess it doesn't actually matter, I guess we'll see if anyone complains.

Seems as you like the solution I'll hook it up to the flag system and see how it goes.

Morganamilo:
I've changed my mind, I forgot about --aur/--repo, which are used on the fly and not just configured once like sortmode. So there needs to be some way for --aur/--repo to work and for -a to work as a short flag.

I'd much rather do `yay -Ssa foo` than `yay -Ss --mode=aur foo`

Maybe The alias method should be implemented after all a=`mode=aur`? That in itself seems a little hacky though as I would prefer there be no `--mode` flag but only an `-a` flag.

Jguer:
I agree on `--aur` and `--repo`, I think those since they are rather permanent can actually be checked for it is less work

Morganamilo:
When I was working on it it was 90% percent done. (although the last fiddly 10% will still take an annoying amount off effort to get right).

#746 is nowhere near being merged. I don't want it merged in half baked. The main blocker was a pacman bug. The fix has been merged https://git.archlinux.org/pacman.git/commit/?id=f9eb2a so at least work in the pr can continue (using pacman-git atleast).

#778 I will probably close. I'm pretty sure it's possible to break packages using this. Similar to partial upgrades. And I don't see it being that adventitious anyway so not much point putting it behind a config option.

bobi32:
I just noticed the both of you talking of #746 here. I’m answering to let you know that I don’t think I’ll find the time to work on it before march, but I’m still planning on having it finished and cleaned to get it merged eventually.

PR #732, opened 2018-09-25
Fix provider menu showing when it shouldn't

### LINKED ISSUE #730: yay tries to install packages as dependencies that are already satisfied
#### Affected Version
yay v8.1139

#### Issue
When I try to install [`josm-svn`](https://aur.archlinux.org/packages/josm-svn) from the AUR which make-depends on `java-environment>=8`, yay asks to select a provider for that dependency:

```
$ yay -S josm-svn 
:: There are 2 providers available for java-environment>=8:
:: Repository extra
    1) jdk10-openjdk 2) jdk8-openjdk 
```

But [`jdk8`](https://aur.archlinux.org/packages/jdk8/) is already installed and provides `java-environment=8`.

#### Steps to reproduce

1. Install [`jdk8`](https://aur.archlinux.org/packages/jdk8/): `yay -S jdk8`
2. Install [`josm-svn`](https://aur.archlinux.org/packages/josm-svn): `yay -S josm-svn`

Morganamilo:
If the package is already installed, we need to check if it is in the
repos to see if it aplies to rebuild tree. Normally alpm will not prompt
us to select a provider if we already have one installed. An exception
comes up when we have a provider installed that is from the AUR. In this
case FindSatisfier() still asks us to select a provider.

Now make sure to never to never show the menu if the package exists in
the local repo.

Fixes #730

### DIFF
--- a/callbacks.go
+++ b/callbacks.go
@@ -19,6 +19,10 @@ func questionCallback(question alpm.QuestionAny) {
 		return
 	}
 
+	if hideMenus {
+		return
+	}
+
 	size := 0
 
 	qp.Providers(alpmHandle).ForEach(func(pkg alpm.Package) error {
--- a/config.go
+++ b/config.go
@@ -119,6 +119,8 @@ var alpmHandle *alpm.Handle
 // Mode is used to restrict yay to AUR or repo only modes
 var mode = ModeAny
 
+var hideMenus = false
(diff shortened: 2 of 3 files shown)


Jguer:
I think I'm going to push this as a release before merging the giant config changes

PR #738, opened 2018-09-27
Rework cleanafter

### LINKED ISSUE #736: Cache VCS repositories only
#### Affected Version

yay v8.1139 - libalpm v11.0.1

#### Issue
Currently, `yay` may save everything if the package is successfully built, which includes history versions of built packages that can be very space-consuming.

It would be great if `yay` has an option to cache only VCS repositories to speed up future checkout.

The `SRCDEST` option in `makepkg` also saves all non-VCS sources, which is useless for me in most situations: new tarballs always need to be downloaded, and keeping old ones does not really help.

Morganamilo:
Fixes #736 and a little more.


Jguer:
I'm going to push a new release now +/-. Question, do you want to include this in that release by this way:
- branch master to v8 branch, merge this into said v8 branch, then I handle the tagging and releasing leaving reimplement-config to be merged into master and then this can be rebased on that?

Morganamilo:
I guess if you're releasing right now, leave this for the next one. Although the other thing I'm about to push can be merged right away if you like.

Also I'd like to read #717 before you merge it. Although that's on me for not reading it yet.

Jguer:
Sure thing, we can still delay the config for a while, so if you want to merge this into master we can then give it a few more days to cooldown and rebase reimplement-config again against master when it's time

Jguer:
Ok, I'll be checking the stability of master to release, and bringing reimplement-config to parity

PR #746, opened 2018-10-03
[Fixes #5] --build option

### LINKED ISSUE #5: Request: Equivalent of Yaourt's Sync Option: --build
```
SYNC OPTIONS
-b, --build
    Build from sources, ABS for official packages, or AUR if packages is not found. Specify this option twice to build all dependencies.
```

bobi32:
Attempt to add a build from ABS options for official packages.
Non-failing yay -S[y][u] --build [\<pkg\>], as long as \<pkg\> is an official or AUR package or there are only official & AUR packages to update.

May not be merged as is, as thorough tests are most certainly needed to make sure all cases are covered!
An important point to address is the management of additional non-official repositories: as is they are silently ignored with --build option, and I don’t know how to deal with them in a tidy manner.

Any comments and hints are welcome towards improving this PR before potential merging.

### DIFF
--- a/config.go
+++ b/config.go
@@ -65,6 +65,7 @@ type Configuration struct {
 	SudoLoop           bool   `json:"sudoloop"`
 	TimeUpdate         bool   `json:"timeupdate"`
 	NoConfirm          bool   `json:"-"`
+	Build              bool   `json:"-"`
 	Devel              bool   `json:"devel"`
 	CleanAfter         bool   `json:"cleanAfter"`
 	GitClone           bool   `json:"gitclone"`
--- a/depSolver.go
+++ b/depSolver.go
@@ -200,10 +200,8 @@ func (ds *depSolver) ResolveRepoDependency(pkg *alpm.Package) {
 	if ds.Seen.get(pkg.Name()) {
 		return
 	}
-	ds.Repo = append(ds.Repo, pkg)
-	ds.Seen.set(pkg.Name())
 
-	pkg.Depends().ForEach(func(dep alpm.Depend) (err error) {
+	doResolve := func(dep alpm.Depend) (err error) {
 		//have satisfier in dep tree: skip
 		if ds.hasSatisfier(dep.String()) {
 			return
@@ -223,7 +221,18 @@ func (ds *depSolver) ResolveRepoDependency(pkg *alpm.Package) {
 
 		ds.ResolveRepoDependency(repoPkg)
 		return nil
-	})
+	}
+
+	// Manage check & make dependencies before package itself to ensure they are built or installed beforehand.
+	if config.Build {
+		pkg.CheckDepends().ForEach(doResolve)
+		pkg.MakeDepends().ForEach(doResolve)
+	}
+
+	ds.Repo = append(ds.Repo, pkg)
+	ds.Seen.set(pkg.Name())
+
+	pkg.Depends().ForEach(doResolve)
 }
 
 // This is mostly used to promote packages from the cache
@@ -278,10 +287,6 @@ func (ds *depSolver) findSatisfierAurCache(dep string) *rpc.Pkg {
 				continue
 			}
 		}
-	}
-
-	if !config.Provides && providers.Len() >= 1 {
-		return providers.Pkgs[0]
 	}
 
 	if providers.Len() == 1 {
--- a/download.go
+++ b/download.go
@@ -60,6 +60,40 @@ func gitHasDiff(path string, name string) (bool, error) {
 	return head != upstream, nil
 }
 
+func aspDownload(path string, name string) (bool, error) {
+	_, err := os.Stat(filepath.Join(path, name, ".git"))
+	if os.IsNotExist(err) {
+		os.MkdirAll(path, os.ModePerm)
+		cmd := exec.Command("asp", "checkout", name)
+		cmd.Dir = path
+		cmd.Env = append(os.Environ(), "GIT_TERMINAL_PROMPT=0") // useful with asp? :/
+		_, stderr, err := capture(cmd)
+		if err != nil {
+			os.Exit(1)
+			return false, fmt.Errorf("error cloning %s: %s", name, stderr)
+		}
+
+		return true, nil
+	} else if err != nil {
+		return false, fmt.Errorf("error reading %s", filepath.Join(path, name, ".git"))
+	}
+
+	cmd := exec.Command("asp", "update")
+	cmd.Dir = filepath.Join(path, name)
+	cmd.Env = append(os.Environ(), "GIT_TERMINAL_PROMPT=0")
+	_, stderr, err := capture(cmd)
+	if err != nil {
+		return false, fmt.Errorf("error updating %s: %s", name, stderr)
+	}
+	cmd = passToGit(filepath.Join(path, name), "pull")
+	_, stderr, err = capture(cmd)
+	if err != nil {
+		return false, fmt.Errorf("error pulling %s: %s", name, stderr)
+	}
+
+	return false, nil
+}
+
 func gitDownload(url string, path string, name string) (bool, error) {
 	_, err := os.Stat(filepath.Join(path, name, ".git"))
 	if os.IsNotExist(err) {
--- a/install.go
+++ b/install.go
@@ -847,19 +847,33 @@ func downloadABS(packages []*alpm.Package, buildDir string) (stringSet, error) {
 
 		var url string
 
-		switch base.DB().Name() {
-		case "core", "extra", "testing":
-			url = "https://git.archlinux.org/svntogit/packages.git/snapshot/packages/" + pkg + ".tar.gz"
-		case "community", "multilib", "community-testing", "multilib-testing":
-			url = "https://git.archlinux.org/svntogit/community.git/snapshot/packages/" + pkg + ".tar.gz"
-		default:
-			return
-		}
+		if shouldUseGit(filepath.Join(config.BuildDir, "packages", pkg)) {
+			clone, err := aspDownload(filepath.Join(config.BuildDir, "packages"), pkg)
+			if err != nil {
+				errs.Add(err)
+				return
+			}
+			if clone {
+				mux.Lock()
+				cloned.set(pkg)
+				mux.Unlock()
+			}
+		} else {
 
-		err := downloadAndUnpack(url, buildDir)
-		if err != nil {
-			errs.Add(err)
-			return
+			switch base.DB().Name() {
+			case "core", "extra", "testing":
+				url = "https://git.archlinux.org/svntogit/packages.git/snapshot/packages/" + pkg + ".tar.gz"
+			case "community", "multilib", "community-testing", "multilib-testing":
+				url = "https://git.archlinux.org/svntogit/community.git/snapshot/packages/" + pkg + ".tar.gz"
+			default:
+				return
+			}
+
+			err := downloadAndUnpack(url, buildDir)
+			if err != nil {
+				errs.Add(err)
+				return
+			}
 		}
 
 		mux.Lock()
--- a/install.go
+++ b/install.go
@@ -35,9 +35,11 @@ func install(parser *arguments) error {
 				}
 			}
 		} else if parser.existsArg("y", "refresh") || parser.existsArg("u", "sysupgrade") || len(parser.targets) > 0 {
-			err = earlyPacmanCall(parser)
-			if err != nil {
-				return err
+			if !parser.existsArg("build") {
+				err = earlyPacmanCall(parser)
+				if err != nil {
+					return err
+				}
 			}
 		}
 	}
@@ -119,7 +121,7 @@ func install(parser *arguments) error {
 		return err
 	}
 
-	if len(ds.Aur) == 0 {
+	if len(ds.Aur) == 0 && !parser.existsArg("build") {
 		if !config.CombinedUpgrade {
 			if parser.existsArg("u", "sysupgrade") {
 				fmt.Println(" there is nothing to do")
@@ -140,6 +142,10 @@ func install(parser *arguments) error {
 	conflicts, err := ds.CheckConflicts()
 	if err != nil {
 		return err
+	}
+
+	if parser.existsArg("build") {
+		downloadABS(ds.Repo, config.BuildDir)
 	}
 
 	for _, pkg := range ds.Repo {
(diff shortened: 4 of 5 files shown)


Morganamilo:
Skimming through the code, I recognise quite a bit of stuff from download.go and other places, you can use the already existing functions, if they don't quite fit then refactor the common code out.

I see no changes to the the dependency resolution so I'm assuming this will fail for packages with {make,check}depends.

The `--build` flag can go into the config struct and added to the switches in parser.go. Then it will be extracted to the struct and you won't have .delArg().

bobi32:
Thank you for your feedback!

Concerning the behavior of the option by now, I don’t know for checkdepends (still have to find an official package with it), but makedepends doesn’t seem to be a problem at all.

FYI I test it with extra/gpaste, removing unused dependencies (with yay -Rcs $(yay -Qdqt), which results in gobject-introspection being uninstalled) before rebuilding it with yay -S --build gpaste.
This results in gobject-introspection — alongside python-mako & python-beaker, as gobject-introspection depends on python-mako which depends on python-beaker — being reinstalled (while being only listed in makedepends in the PKGBUILD) and gpaste flawlessly built.

From there, the question is more about the consistency of the --build option: I assume it should build not only the specified packages (or packages to update), but also their dependencies. But what about make & check dependencies?

Regarding checkdepends, do you know some official packages that might be of help to test it?

Concerning the other points, I moved the --build option handling to handleConfig. I’ll have a look at what can be refactored and might get back to you for some help on this.

Morganamilo:
>From there, the question is more about the consistency of the --build option: I assume it should build not only the specified packages (or packages to update), but also their dependencies. But what about make & check dependencies?

```
--buildall: build every package
--build: build targets only
--nobuild: don't build anything
```

We already have this type of thing on many other flags.

>Regarding checkdepends, do you know some official packages that might be of help to test it?

Pacman has checkdepends

>Concerning the other points, I moved the --build option handling to handleConfig.

This is more correct, but the flag should have its json value set and corresponding flag to disable building. That being said, the entire config system (and flag setting) is going to be redone, so there's no really point in doing it. You can just ignore that and focus on the behaviour.

Again on the behaviour I would also expect this to integrate pretty well will the rest of the system. `Remove make dependencies after install?` should also know how to deal with the repo makedeps for example. And more importantly the pkgbuild diff/editing should also work with repo stuff. If you're building repo packages you probably want to edit them.

That kind of introduces a bigger problem though. How do you save these changes? These pkgbuilds are not git managed to they can not be commited and merged with upstream.

bobi32:
> Pacman has checkdepends

Ok, so checkdepends is not a problem either (in my test environment, fakechroot was missing, and it was installed before yay tries and builds pacman).

> […] the flag should have its json value set and corresponding flag to disable building. That being said, the entire config system (and flag setting) is going to be redone, so there's no really point in doing it. You can just ignore that and focus on the behaviour.

Ok, I assumed (wrongly?) that `\`json:"-"\`` was a way to prevent the option for being settable from the config file, which seemed to me a good idea to avoid bad surprises on the next big upgrade.

> Again on the behaviour I would also expect this to integrate pretty well will the rest of the system. `Remove make dependencies after install?` should also know how to deal with the repo makedeps for example.

I’ll have to get a look at it, as ATM the prompt does not appear if no AUR package with new makedeps is involved.

> And more importantly the pkgbuild diff/editing should also work with repo stuff. If you're building repo packages you probably want to edit them.

I was pretty sure it did work as yay does prompt the edit choice (which I systematically skipped until now), but I just realized that official packages are not listed nore selectable (e.g. with A choice). It turns out the option is still more buggy than I thought ^^

> That kind of introduces a bigger problem though. How do you save these changes? These pkgbuilds are not git managed to they can not be commited and merged with upstream.

Okay, here I think you’re referring to yay or AUR features I don’t even know of. For what I get, does it allows users to keep, for a given package, custom pkgbuilds from one version to another?
I remember that back in the day of yaourt & ABS, I used yet another tool found on AUR that was in charge of keeping local (e.g. on my own system) patches for official packages to be built (IIRC I used it to add some build options to curl).
Is this similar to what you refer to?

Morganamilo:
>Ok, so checkdepends is not a problem either (in my test environment, fakechroot was missing, and it was installed before yay tries and builds pacman).

Yes I see now that you just added -s to the makepkg call. This creates all sorts of problems, yay is totally unaware of these packages so it never lists them before hand. And this breaks yay as a pacman wrapper because makepkg will `pacman -S` these packages ignoring any flags we originally passed. So `yay -S foo --dbpath /tmp/db` will break.

>Ok, I assumed (wrongly?) that the value json:"-" was a way to prevent the option from being settable from the config file, which seemed to me a good idea to avoid bad surprises on the next big update.

Yes this is what "-" does. There's no real reason to not let users save it. They won't get a nasty surprise because --save has to be explicitly used.  

>I’ll have to get a look at it, as ATM the prompt does not appear if no AUR package with new makedeps is involved.

As noted before Yay is totally unaware these dependencies were even installed

>Okay, here I think you’re referring to yay or AUR features I don’t even know of. For what I get, does it allows users to keep, for a given package, custom pkgbuilds from one version to another?
I remember that back in the day of yaourt & ABS, I used yet another tool found on AUR that was in charge of keeping local (e.g. on my own system) patches for official packages to be built (IIRC I used it to add some build options to curl).
Is this similar to what you refer to?

Each AUR package acts as a git repo. This allows you to permanently save changes and merge them with upstream.

```
git clone https://aur.archlinux.org/yay.git
vim PKGBUILD
git commit -am "my changes"
```

Now whenever you pull these changes will remain.

On the other hand ABS packages are downloaded as tarballs. The only solution I see is using `asp` but I'm not too sure how well it can handle merging.

bobi32:
> Yes I see now that you just added -s to the makepkg call. This creates all sorts of problems, yay is totally unaware of these packages so it never lists them before hand. And this breaks yay as a pacman wrapper because makepkg will `pacman -S` these packages ignoring any flags we originally passed. So `yay -S foo --dbpath /tmp/db` will break.

My bad, I forgot I actually did that. I’m pretty sure it goes back to when I was still trying to understand the code more than actually implementing the option.
As I know what is intended towards fixing this, I’ll focus on it first.

> Yes this is what "-" does. There's no real reason to not let users save it. They won't get a nasty surprise because --save has to be explicitly used.

Ok, I’ll take care of it when I add the `--nobuild` & `--buildall` options then.

> Each AUR package acts as a git repo. This allows you to permanently save changes and merge them with upstream.
> 
> ```
> git clone https://aur.archlinux.org/yay.git
> vim PKGBUILD
> git commit -am "my changes"
> ```
> 
> Now whenever you pull these changes will remain.
> 
> On the other hand ABS packages are downloaded as tarballs. The only solution I see is using `asp` but I'm not too sure how well it can handle merging.

Alright, it just never occurred to me to take advantage of AUR being based on git in such a way!
I’ll have a look at `asp` and see what can be done with it.

bobi32:
With further investigation, it seems that make & check depends are available only when the Package object is built from the information loaded from the PKGBUILD, and not from the synchronization DB.

As a consequence, it turns out it is then necessary to download and load the PKGBUILD before taking care of a package’s `[{make,check}]deps`, this instead of requesting the DB (if the PKGBUILD of a package has all the information needed regarding this package, it seems redundant to get half of it from a second source).

Morganamilo:
Actually the information is in the db. `zcat /var/lib/pacman/sync.db | less`

If you take a look at sync_db_read() in besync.c:702 you will see the relevant code to load {make,check}depends is missing with a comment "currently unused".

So my goal would be get it implemented in pacman. Failing that I would rather just use the rpc.

Morganamilo:
From IRC:
```
14:55 <zzorra> Hey, i noticed alpm_pkg_get_{make,check}depends don't work. I tracked it down to
 sync_db_read(), where the fields are commented out with "currently unused". Is there a reason for this 
other than simply being unused? And would simply replacing it with the relevant READ_AND_SPLITDEP
 macro work and be accepted as a patch?
15:36 <elibrokeit> zzorra: we originally added these functions in commit 0994893b0e6b627d45a63884ac01af7d0967eff2
15:36 <elibrokeit> due to someone contributing a patch because, presumably, they wanted it
15:37 <elibrokeit> but, this patch apparently did not touch be_sync.c
15:37 <elibrokeit> so, I'm guessing allan would be fine if you submitted a patch to do exactly this
```

bobi32:
Hi, following your advice, I opened a bug request for libalpm to be patched upstream: https://bugs.archlinux.org/task/60347

Considering this is due for pacman 5.1.2 at the moment, I work with a patched libalpm from now on.

The three last commits take advantage of this to manage {check,make}depends for official packages in the same way as AUR packages, instead of delegating it to `makepkg -s`

For now, check & make dependencies are built alongside the package to build no matter what (_i.e._ in a `--buildall` manner). I was able to make yay propose the removal of `{check,make}deps` afterwards, which does work.

Morganamilo:
Also seems as you're working on it now you can patch pacman yourself. It will be a while before anything gets into pacman-git let alone release.

```patch
From 52fcb2bb8cd59f0986c645f9ee3d80bddd26b43a Mon Sep 17 00:00:00 2001
From: morganamilo <morganamilo@gmail.com>
Date: Mon, 8 Oct 2018 09:14:26 +0100
Subject: [PATCH] libalpm: parse {check,make}depends when building a alpm_pkg_t

Commit 0994893b0e6b627d45a63884ac01af7d0967eff2 added the
alpm_pkg_get_{make,check}depends functions but forgot to include
logic for parsing these fields from the database. As a result these
functions will always return an empty list.

This commit adds the parsing logic.

Signed-off-by: morganamilo <morganamilo@gmail.com>

diff --git a/lib/libalpm/be_local.c b/lib/libalpm/be_local.c
index 7c2f96b8..f8c3947e 100644
--- a/lib/libalpm/be_local.c
+++ b/lib/libalpm/be_local.c
@@ -773,6 +773,10 @@ static int local_db_read(alpm_pkg_t *info, int inforeq)
 				READ_AND_SPLITDEP(info->depends);
 			} else if(strcmp(line, "%OPTDEPENDS%") == 0) {
 				READ_AND_SPLITDEP(info->optdepends);
+			} else if(strcmp(line, "%MAKEDEPENDS%") == 0) {
+				READ_AND_SPLITDEP(info->makedepends);
+			} else if(strcmp(line, "%CHECKDEPENDS%") == 0) {
+				READ_AND_SPLITDEP(info->checkdepends);
 			} else if(strcmp(line, "%CONFLICTS%") == 0) {
 				READ_AND_SPLITDEP(info->conflicts);
 			} else if(strcmp(line, "%PROVIDES%") == 0) {
diff --git a/lib/libalpm/be_sync.c b/lib/libalpm/be_sync.c
index 5009a7da..af94b2d5 100644
--- a/lib/libalpm/be_sync.c
+++ b/lib/libalpm/be_sync.c
@@ -700,17 +700,9 @@ static int sync_db_read(alpm_db_t *db, struct archive *archive,
 			} else if(strcmp(line, "%OPTDEPENDS%") == 0) {
 				READ_AND_SPLITDEP(pkg->optdepends);
 			} else if(strcmp(line, "%MAKEDEPENDS%") == 0) {
-				/* currently unused */
-				while(1) {
-					READ_NEXT();
-					if(strlen(line) == 0) break;
-				}
+				READ_AND_SPLITDEP(pkg->makedepends);
 			} else if(strcmp(line, "%CHECKDEPENDS%") == 0) {
-				/* currently unused */
-				while(1) {
-					READ_NEXT();
-					if(strlen(line) == 0) break;
-				}
+				READ_AND_SPLITDEP(pkg->checkdepends);
 			} else if(strcmp(line, "%CONFLICTS%") == 0) {
 				READ_AND_SPLITDEP(pkg->conflicts);
 			} else if(strcmp(line, "%PROVIDES%") == 0) {
-- 
2.19.1
```

Morganamilo:
Outpit:

```
pacman
	deps:		7
		bash
		glibc
		libarchive
		curl
		gpgme
		pacman-mirrorlist
		archlinux-keyring
	checkdeps:	2
		python2
		fakechroot
	makedeps:	1
		asciidoc
gpaste
	deps:		1
		gtk3
	checkdeps:	0
	makedeps:	7
		intltool
		vala
		appstream-glib
		gobject-introspection
		gnome-shell
		gnome-control-center
		git
```

Morganamilo:
be_sync.c and be_local.c for sync and local databases respectively. But yeah just that one patch is needed.

bobi32:
I updated my bug request accordingly.

Concerning the possibility to keep local changes with `asp`, my last commit provides a behavior similar to AUR packages cloned from git. As is, it makes asp a new mandatory dependency for yay.
I wonder if it would be better to make it an optional dependency and add a check whether or not it is installed in an according shouldUseASP function.

Morganamilo:
You see to have messed up a little, 71f3339 adds a lot of changes that should already be there and 
b37243e removes a merge conflict that you never solved.

For unmerged code it's usually best to go back edit the history rather than add new commits. git rebase is a very useful tool if you learn how to use it.

bobi32:
Done.

BTW, beyond this rebase, my contribution to this PR has been quite on hold, as I assumed it won’t be merged before next release of pacman, as the `--build` option makes use of an already patched version of libalpm.
That + I am waiting for the config overhaul to be done before working on `--build-all` & `--nobuild` options; regarding this point, maybe the reimplement-config branch is advanced enough for me to rely on it?

Morganamilo:
Yep not much you can do for now apart from either build against a patched version or just pretend the functions work but make sure to always have make/check deps installed before hand.

Alan said is schedule clears up next week to maybe the patch will land in master and you can just develop using pacman-git.

The config overhaul needs to be rebased on master. So it's kinda a lose-lose whatever you do unfortunately. 

By the way rebasing usually does not generate an extra commit, so Don't know exactly what you did. It does look neater already though.

bobi32:
The last commit may not have the best comment, but it’s only me patching up parts of my code that did not raise conflicts during the rebase: basically replacing depOrder & depPool calls from my contributions with depSolver ones.
As a consequence, https://github.com/Jguer/yay/pull/746/commits/8e34b718b1f5c35f57bef3769eb10c5e3aa1900b ain’t a compilable state, as there still remains calls to do & dp despite the corresponding source files are already removed. I personally don’t see this as a problem, hence the additional commit instead of having all things (even not conflicting ones) patched up in the rebase.

Morganamilo:
The way to solve that is to do an interactive rebase and and ammend each commit of yours that adds references to dp/do.

I'm not actually that picky though, I probably won't refuse to merge just because of it. But I do bring it up because it is the neater way to do things.

bobi32:
https://bugs.archlinux.org/task/60347#comment176227

Upstream changes are commited and are due on next maintenance release!
This should revive this PR, I hope so.

Avi-D-coder:
@bobi32 Does this pull request enable building any packages from a particular repo?
The use case I have in mind is to be able to build any packages from the `community` repo from source.

PR #756, opened 2018-10-07
Don't show missing packages if they are ignored.  Closes #752

### LINKED ISSUE #752: Non-AUR packages listed as missing in the AUR
Locally created packages, that are not part of the AUR (and the Archlinux repo) are listed as missing in the AUR:

    $ yay -Syu
     core is up to date
     extra is up to date
     community is up to date
     there is nothing to do
     -> Missing AUR Packages:  mypackage1  mypackage2
     there is nothing to do
Adding those packages to IgnorePkg= in file /etc/pacman.conf does not help.

yay v8.1139.r13.g6ee14d4 - libalpm v11.0.1

Jguer:


Morganamilo:
Won't work with wildcards. Also this provides an annoying extra confirmation when installing packages with pacman/makepkg.

I would prefer a dedicated config field for it. But I don't really want to add anything until after the new config system is in place.

Morganamilo:
>Your second point I don't see what you mean

The use case for this is to ignore the warnings for non AUR packages, presumably pkgbuilds that are managed manually by the user. If the package is ignored then whenever I build and install, pacman is going to be like "this package is ignored, do you want to install it?" every single time I update it. This extra prompt would be a little annoying.

Jguer:
> Yes, but this commit doesn't cause that, users bad policies cause that, all this commit does is hide the package from the print warning for the users who choose to do that and are annoyed by having one extra line of output every time they update. 
It's okay like this, it doesn't impact anyone negatively except @willemw12 s positively 

Is what I was writing but after reading #752 again I understand that what should be done is revert this change and the text from `Missing AUR packages` to `Missing packages` and that should fix all confusions. No need to hack `ignorepkg` for this since according to @willemw12 pacman also displays ignored packages in upgrades

Issue #108, opened 2018-01-22
Install Algorithm Progress


Morganamilo:
This issue is to centralise all problems and progress on the new install algorithm. The prior pull requests can be seen here #105 #107 and should be read as some of the know problems are outlined there in detail.

Formatting and printing to the user should be discussed separately in #109 .

If there is a major bug in the install or you wish to discuss a problem in detail a separate issue should be opened. This issue is for general discussion and bug reporting. 

## Testing
If you are willing to test you can either test against the master branch or any open pull requests if there are any. If doing the latter you should comment on the pull request instead of here.

My testing is usually preformed in a clean chroot with only base and base-devel installed as to pull in as many dependencies as possible during each operation. This really helps in spotting missing or incorrect dependency solving. 

I will be opening another issue #110  to keep track of difficult to install packages and packages which currently do not install correctly.

 ## Status

- [x] Reliable solver of complex dependencies
- Nothing has been found yet so I'm amusing this is working
- ~~As far as i know the following issues are fixed but more testing it needed to tick this box.~~
- ~~This is mostly the case, we still miss some dependencies when dealing with packages that are pulled using its provides instead of its package name.~~
- ~~This can be seen when installing `discord` and the dependency tree tries to pull `libjpeg` provided by `libjpeg-turbo`~~

- [x] Split packages
- ~~Split packages are currently all or nothing. Yay will always install every package generated by a split package. If two packages `foo` and `bar` from the same pkgbase `base` are specified, Yay in an attempt to build `foo` will build `base`, install `foo` and `bar` then in an attempt to install `bar`, will rebuild `base` and reinstall `foo` and `bar`.~~
- ~~This can be seen when installing `discord` and the dependency tree pulls `libc++abi` and `libc++`.~~

- [x] Circular dependencies
- ~~After realising packages wont even build without their dependencies installed I'm not supporting circular dependencies and marking this as solved.~~
- Don't seem to be a problem for repo packages, I am currently unaware of any AUR packages with circular dependencies. 

- [x] Versioned dependancies
~~- Checked by goalpm for repo packages. Currently ignored for AUR packages.~~
~~- I am unsure whether yay should handle this, goalpm should handle this or gopkgbuild should handle this.~~

- [ ] IgnorePkg/IgnoreGroup
- Checked during sysupgrade, not during normal install.

- [ ] Respect all Pacman options (--root --dbpath --config...)
- Multiple CacheDirs is not supported
[...]

Issue #42, opened 2017-07-28
Package Install specifications


Jguer:
# Managing install dependencies

## Preamble
Dependency management is a critical part in the install process of packages, pacman's approach is seemingly to install packages with their dependencies in one go since they're prebuilt.  
Most AUR packages require build dependencies without which the install process will fail.

The proposed solution for this problem in yay is creating an install graph and installing by tiers.

## Behavior

The user selects what packages he wants to install from the yay menu.
The install package will be fed a slice of AUR packages and Repo packages.

These AUR and Repo packages will be tier 1, tier 1 is for explicitly installed packages.  
If a tier 1 package is a dependency to another tier 1 package it will be installed on another tier and made an explicitly installed.

Already installed packages will be ignored.

Example:
To install yaourt-git (AUR) and caja (repo)

```
Final result after processing, git and pacman were already installed.
|----------------------Tier 1--------------------------|
|  caja yaourt-git                                     |
|------------------------------------------------------|

|----------------------Tier 2--------------------------|
| package-query-git                                    |
| exempi gvfs libexif libsm mate-desktop libnotify     |
|------------------------------------------------------|

|----------------------Tier 3--------------------------|
| yajl gcc-libs expat zlib                             |
| avahi dconf fuse libarchive libcdio-paranoia libsoup |
| udisks2  libsecret libbluray  libgudev  gcr  psmisc  |
|                       ....                           |
|------------------------------------------------------|

|----------------------Tier 4--------------------------|
| glibc  fuse-common                                   |
|                       ....                           |
|------------------------------------------------------|
```  

Before install all tiers are shown to the user along with tier size for approval. After approval the user will only be prompted again at the tier 1 install. Rejecting the tier 1 install will remove the whole tree.

## Implementation
Important points for implementation:  
- After editing the `PKGBUILD` a new `.SRCINFO` will need to be generated
- Capture Ctrl-C and remove full tree.
- Packages are stored as a graph.
- Graph is DFSed before adding each package since adding the same package twice would be bad.
- Nodes will be an interface with a Type function.
``` 
type Node interface {
    Type() //AUR or Repo (there probably is a better way, awaiting revision)
}  
```
- Transversal will be made through a reverse BFS.  
>For the implementation of a reverse BFS, the graph will have a tier 0 (dummy head). 
- Package details will come from `.SRCINFO` parsing since it is the only one that accounts for split packages.

## Objectives

- Get another check on the wiki ( Resolve complex dependencies)
- #5
- #10

[...]

PR #767, opened 2018-10-13
Merge dep pool order and rework inner conflicts

### LINKED ISSUE #673: 'yay -S code' cause 'inner conflict conflicts'
#### Affected Version
yay v8.998 - libalpm v11.0.1

#### Issue
'yay -S code' cause 'inner conflict conflicts', it is difficult to figure out which conflicts by the error message.

#### Steps to reproduce
1. yay -S code
2. yay -S code-git # also show the same error message

 -> nodejs-lts-carbon: nodejs

My workaround

1. yay -G code
2. cd code
3. makepkg -s  # this seems solved the conflict issue?!
4. yay -S code-git # workable now

### LINKED ISSUE #695: 'powerpc-linux-gnu-gcc' can't be installed correctly
#### Affected Version
yay v8.1115.r2.g8248c80 - libalpm v11.0.1

#### Issue
We can't use `yay` to install `powerpc-linux-gnu-gcc` correctly.
Some reasons may be that `yay` can't check for circular dependency correctly and `yay` doesn't resolve the right install order.

#### Steps to reproduce
1.`yay -S powerpc-linux-gnu-gcc`
2.Install the `powerpc-linux-gnu-binutils` and `powerpc-linux-gnu-linux-api-headers` dependency.
3.When `yay` is about to install `powerpc-linux-gnu-glibc`, we will get the following error:

> ==> Missing dependencies:
  -> powerpc-linux-gnu-gcc-stage2>=8.1.0-1
==> ERROR: Could not resolve all dependencies.
Error making: powerpc-linux-gnu-glibc

Morganamilo:
Title says it all, commits for details.

For some examples of the inner conflict handling.

yay -S powerpc-linux-gnu-gcc
```
==> Inner conflicts found:
 -> Installing powerpc-linux-gnu-gcc-stage2 will remove: powerpc-linux-gnu-gcc-stage1
 -> Installing powerpc-linux-gnu-glibc will remove: powerpc-linux-gnu-glibc-headers
 -> Installing powerpc-linux-gnu-gcc will remove: powerpc-linux-gnu-gcc-stage2
```

yay -S yay yay-git yay-bin
```
==> Inner conflicts found:
 -> Installing yay-git will remove: yay
 -> Installing yay-bin will remove: yay-git
```

yay -S ffmpeg-libfdk_aac 
```
==> Package conflicts found:
 -> Installing ffmpeg-libfdk_aac will remove: ffmpeg
```

Fixes #673 in a nicer way
Fixes #695 

Note: powerpc-linux-gnu-gcc has not been tested apart from reading the conflict message. I'm assuming it's gonna take a long time to build to I'll get to that later.


Morganamilo:
So I know why stuff gets duplicated but coming up with a solution might be a little annoying.

On the plus side, totally ignoring the duplicates yay can indeed solve `powerpc-linux-gnu-gcc`.

yay -S powerpc-linux-gnu-gcc
```
installed powerpc-linux-gnu-binutils (2.30-1)
installed powerpc-linux-gnu-linux-api-headers (4.16.1-1)
installed powerpc-linux-gnu-gcc-stage1 (8.2.0-1)
installed powerpc-linux-gnu-glibc-headers (2.27-1)
removed powerpc-linux-gnu-gcc-stage1 (8.2.0-1)
installed powerpc-linux-gnu-gcc-stage2 (8.2.0-1)
removed powerpc-linux-gnu-glibc-headers (2.27-1)
installed powerpc-linux-gnu-glibc (2.27-1)
removed powerpc-linux-gnu-gcc-stage2 (8.2.0-1)
installed powerpc-linux-gnu-gcc (8.2.0-1)
```

Morganamilo:
Consider the problems solved. This new way is much slower though.

1.9s vs 5.2s resolve time for ros-lunar-desktop-full. Need to fix that.

Jguer:
Ok looking good, when you say above is fixed, have you cut the resolve time back to 1.9s?

bet4it:
Thanks for your great work!
But there seem to be something annoying when I install `powerpc-linux-gnu-gcc`:
```
$ yay -S powerpc-linux-gnu-gcc
:: There are 2 providers available for powerpc-linux-gnu-gcc-stage2>=8.1.0-1:
:: Repository AUR
    1) powerpc-linux-gnu-gcc-stage2 2) powerpc-linux-gnu-gcc 

Enter a number (default=1): 2
:: There are 2 providers available for powerpc-linux-gnu-gcc-stage2>=8.1.0-1:
:: Repository AUR
    1) powerpc-linux-gnu-gcc-stage2 2) powerpc-linux-gnu-gcc 

Enter a number (default=1): 1
:: There are 2 providers available for powerpc-linux-gnu-glibc-headers>=2.27-1:
:: Repository AUR
    1) powerpc-linux-gnu-glibc-headers 2) powerpc-linux-gnu-glibc 

Enter a number (default=1): 1
:: There are 3 providers available for powerpc-linux-gnu-gcc-stage1>=8.1.0-1:
:: Repository AUR
    1) powerpc-linux-gnu-gcc-stage1 2) powerpc-linux-gnu-gcc 3) powerpc-linux-gnu-gcc-stage2 

Enter a number (default=1): 1
```
Can we try to let `yay` select them automatically?

Morganamilo:
The first two are duplicates and should be fixed to show only one prompt.
The fact that you can spam the second option and go in circles forever is a little stupid but not really sure if it's worth fixing.

Apart from that everything is working as intended. There is a --noprovides flag, which stops yay going out of its way to find providers. I added the flag because I thought I thought provider searching may slow things down quite a bit. But in its current state, even ros stuff can be resolved pretty quick will the flag enabled (1.7s vs 3.7s).

So maybe it might be a good idea to re purpose --noprovides to hide yay's provider menu also.

About the first paragraph, the way dependency resolving works now is that the package is not added to our install list until all of its dependencies have been resolved. So when you do `yay -S powerpc-linux-gnu-gcc` and it starts resolving deps, because that package has not been added to the list yet, the provider menus may still suggest that package even if it shouldn't really be suggesting it.

So kinda a lot of effort to figure out a fix for something so minor. Plus once you've bootstrapped it once you shouldn't need to do it again.

Jguer:
> So maybe it might be a good idea to re purpose --noprovides to hide yay's provider menu also.

:+1:

Jguer:
Thinking of making a release of the current state, want to cram this in before?

Morganamilo:
Nah I don't want to merge this PR now and implementing the above point will make a merge conflict so I'll do it after this is merged.

bet4it:
I don't think the provider menus are suitable.
Packages which are provided to choose should have the same effects during installing. But in this case, select `powerpc-linux-gnu-gcc` won't help us with the need of `powerpc-linux-gnu-gcc-stage2`.

I think it's a problem relates with circular dependency which has been mentioned in #108.
I don't know how to solve circular dependency or if it is solvable, but I think it's better to avoid it.

In this case, `powerpc-linux-gnu-gcc` depends on `powerpc-linux-gnu-glibc`, `powerpc-linux-gnu-glibc` depends on `powerpc-linux-gnu-gcc-stage2`, and both `powerpc-linux-gnu-gcc` and `powerpc-linux-gnu-gcc-stage2` can provide `powerpc-linux-gnu-gcc-stage2`.
But if we still choose `powerpc-linux-gnu-gcc` here, there will form a `powerpc-linux-gnu-gcc -> powerpc-linux-gnu-glibc -> powerpc-linux-gnu-gcc` circular dependency.
So it's better for us to exclude packages which will cause a circular dependency problem when they are selected.

Then we are facing the problem that how to detect cycles in a directed graph.
It's easy to do it if the graph is traversed with DFS, we only need to check if we have encountered this package in the call chains, and `aurman` has implemented it:
https://github.com/polygamma/aurman/blob/dcb50aa1bf5296dfadbffbe867d3e7e807442397/src/aurman/classes.py#L502-L534

But it seem to be hard to detect cycles in a directed graph using BFS.
Unfortunately, `yay` use BFS to get all the packages need to be installed, which is faster than DFS because we can get more information with one RPC request and handle more packages in one function, as a result, we can't detect circular dependency easily.

I notice that #42 has mentioned:
> Graph is DFSed before adding each package since adding the same package twice would be bad.

If we can implement it, we can solve this problem conveniently.

bet4it:
Maybe I should open a new issue about it?

Morganamilo:
I'm not really sure what to do. The old depPool/depOrder split was an attempt at that. The depPool would get all dependencies using level order then the depOrder would order them using DFS. But did didn't end up working 100%.

The current system is a little lax, it can warn you about you about dep cycles, and will succeed with handling them as long as future packages don't depend on the ones being removed. And for me I find that good enough.

With the provider menu stuff, yeah it acts a bit dumb when it comes to circular deps but is fine otherwise, which is 99% of the time.

If you can come up with a solution that's great. But I do not want to compromise speed, so the aurinfo requests must remain level order.

PR #771, opened 2018-10-13
Follow --color= conventions from gnu coreutils

tonylambiris:
According to the `ls(1)` manpage:
```
‘--color [=WHEN]’
     Specify whether to use color for distinguishing file types.  WHEN
     may be omitted, or one of:
        • none - Do not use color at all.  This is the default.
        • auto - Only use color if standard output is a terminal.
        • always - Always use color.
     Specifying ‘--color’ and no WHEN is equivalent to ‘--color=always’.
     If piping a colorized listing through a pager like ‘less’, use the
     ‘-R’ option to pass the color codes to the terminal.
```

In addition add **github.com/mattn/go-isatty** for testing terminal output and update the Makefile test target to skip over vendored code when running. For checking the vendor directory I personally use `dep check` for lint/vet/tests and `dep ensure -v` for ensuring all outside dependencies have been properly resolved.

Thanks for writing this! I've been using it for a while now and I love it. I love golang for a multitude of reasons but the gift that keeps on giving for me is being able to replace all the tools I use written in that garbage of a dumpster-fire language python. (deuces yaourt!) :)


Morganamilo:
>Follow --color= conventions from gnu coreutils

But we already do this right? Or at least something similar.

A more fitting tittle would be "replacecall to `test` with github.com/mattn/go-isatty". And while I agree that it's less hacky to do that, I don't think Jguer or I want to add 179k lines worth of dependencies.

Jguer:
My doubts are the same as @Morganamilo exactly

tonylambiris:
You don't do this because of the addition of isTty() here:
https://github.com/Jguer/yay/blob/master/main.go#L156

As far as adding dependencies, I'd rather have it check it with a github repo that also has support for things like Cygwin and such, but if forking processes is an acceptable way of testing for a terminal I'll be more than happy to close this PR.

Morganamilo:
From pacman.conf(5)
```
Color
           Automatically enable colors only when pacman’s output is on a tty.
```

I'm not sure what GNU's conventions are here as `ls` does not have a config file. We are a pacman wrapper so I'd rather follow whatever standards pacman uses.

Jguer:
For now I would rather keep it as it is, following the maxim "Don't import a dependency you're not willing to maintain", because I think it breaks one of yay's core objectives "Minimal dependencies" and because the current method works fine enough unless you add a binary called test to your path (:innocent:). 

Thank you for the effort @tonylambiris and I hope we'll see more pull requests from you soon

tonylambiris:
@Morganamilo kk ill close it out, thanks for the feedback.

Issue #734, opened 2018-09-25
[Feature request] Allow configurable default for questions


marmistrz:
Currently the default answer (chosen after enter is pressed) is "None". It'd be great if it were configurable. 

I don't mean the `--answerdiff` option - this entirely disables the prompt - I still want to see the prompt, just have `All` as a default.

izzy:
I totally second this. I'd like to have it [All] for diffs and cleanbuilds by default but still be able to select another option when I see fit instead of the current (rather problematic) default of being able to accidentally pressing enter and not seeing any diffs and just blindly installing things.

alxpettit:
Thirded!

marmistrz:
Please reopen.

PR #774, opened 2018-10-15
[Solves #734] Allow configurable default for questions

Tardog:
First I want to stress that constructive feedback is _very_ welcome, especially since I started writing Go code only recently and am still learning the intricacies of the language.

I use yay as my daily AUR helper, so contributing here is a great opportunity to get more hands-on experience with Go and give something back to the project at the same time.

This PR for the feature request #734 adds a new flag `--askwithdefault`, and a matching config option `AskWithDefault`, that alters the way `--answerdiff` `--answeredit` and `--answerclean` work.

When the flag is set or the config option is true, yay no longer skips the answer prompts entirely. Instead, the user can override the --answer* value by entering something else, or simply press enter to confirm the default.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -64,7 +64,7 @@ Permanent configuration options:
     --answerdiff    <a>   Set a predetermined answer for the diff menu
     --answeredit    <a>   Set a predetermined answer for the edit pkgbuild menu
     --answerupgrade <a>   Set a predetermined answer for the upgrade menu
-    --askwithdefault      Don't skip the menu when any of the above is set
+    --askwithdefault      Don't skip the menu when --answerclean, --answerdiff or --answeredit are set
     --noanswerclean       Unset the answer for the clean build menu
     --noanswerdiff        Unset the answer for the edit diff menu
     --noansweredit        Unset the answer for the edit pkgbuild menu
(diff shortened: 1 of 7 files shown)


Morganamilo:
I'm not against #734 but I am pretty much ignoring it for now until the new config system is sorted out. With a more human orientated config file I mat drop the restraint of every config option having a flag and just have a config option along the lines of.

```
[menus]
NoConfirm = Upgrade Clean Diff
```

But time will tell.


a3f9f97 is in the right direction but I would prefer a more dynamic approach where only the long names are entered and the short options are figured out on the fly.

For example:
```
all none abort

for each option:
    a = all
    none = n
    a = abort, but a is taken so use ab
``` 

This makes localisation much easier and is something I've been meaning to do independently from  #734.

P.S @Jguer They way I'd like to implement the config part is with a simple enum where as it iterates each value, we | each value together. A string map makes this annoying :(

Jguer:
Also overlooking this until config is merged.

Alternatively you could put it in int map which would be able to use enums, but then we'd need reflection to read it to pass string to enum from the config :worried:.

Tardog:
Thank you for your suggestions! I did a few takes on the menu generator function and was not sure if a dynamic approach would be the best.

When he new config is done, I'll gladly rework the implementation with localisation in mind. Do you want to leave the PR open until then?

Tardog:
All right, I'll leave it open then. While rebasing _might_ work, I prefer to wait until the new config is stable enough to make it into master.

PR #775, opened 2018-10-15
Handle libalpm callbacks when checking DB satisfiers for a package

### LINKED ISSUE #753: How to upgrade an ignored package?
If a package is configured as ignored in IgnorePkg= in file /etc/pacman.conf, then to implicitly install the package you can do:

    # pacman -S networkmanager
    :: networkmanager is in IgnorePkg/IgnoreGroup. Install anyway? [Y/n]

However, yay silently ignores the package upgrade:

    $ yay -S networkmanager
    $ yay -S --rebuild networkmanager

yay v8.1139.r13.g6ee14d4 - libalpm v11.0.1

dvejmz:
This fixes #753 by re-enabling the default libalpm callback handling behaviour right before checking the satisfiers of a given package.

Before this patch, all libalpm callbacks were being suppressed, which meant a user would never see the `ALPM_QUESTION_INSTALL_IGNOREPKG` prompt to confirm the explicit installation request of a package in the `IgnorePkg` list.

### DIFF
--- a/install.go
+++ b/install.go
@@ -356,7 +356,10 @@ func inRepos(syncDb alpm.DbList, pkg string) bool {
 		return true
 	}
 
+	previousHideMenus := hideMenus
+	hideMenus = false
 	_, err := syncDb.FindSatisfier(target.DepString())
+	hideMenus = previousHideMenus
 	if err == nil {
 		return true
 	}
@@ -384,7 +387,6 @@ func earlyPacmanCall(parser *arguments) error {
 	if mode == ModeRepo {
 		arguments.targets = targets
 	} else {
-		alpmHandle.SetQuestionCallback(func(alpm.QuestionAny) {})
 		//separate aur and repo targets
 		for _, target := range targets {
 			if inRepos(syncDb, target) {


Morganamilo:
I can't really remember but that was probably added originally to hide the provider menu when checking if somethings in the repos. Since then we added a hideMenus variable. That should probably be used instead.

Morganamilo:
Do you plan to edit your PR to hide to menu or do you want me to do it?

dvejmz:
@Morganamilo Yes, I will edit the PR today.

dvejmz:
I've updated the pull request to hide the provider menu when checking if a package is in the repos.

Also, I think it'd be nicer if access to `hideMenus` was granted through a getter and setter function instead of directly referencing this global variable.

It doesn't seem to be used in a lot of places (in fact, it's just used in `depPool.go`) so it wouldn't be too hard of a job to refactor it. I'm happy to raise a separate issue for this and do the refactor myself, if you would find that helpful @Morganamilo

Morganamilo:
I don't really see much a point for getters and setters. Good for encapsulation but that would only really apply if this was its own module/library.

You've got the right idea but this should be inside of inRepos() so that any other calls to it will also hide the menus. Preferably wrapped around the .FindSatisfier() call. 

Also I would prefer if you did a backup and restore of the variable instead of setting it to true and false. In future versions hideMenus might be set to false already before the function call and should stay that way.

dvejmz:
Updated PR.

Yeah, it makes more sense to hide the provider menu every time `inRepos()` is called.

As for the getters/setters, fair enough, although I suggested using those not with encapsulation in mind, but having a more controlled, conflict-free way of accessing the variable. For example, the state backup/restore you pointed out could be done in the setter/getter

```
func setHideMenusState(hidden bool) {
    previousHideMenus = hideMenus
    hideMenus = hidden
}

func popHideMenusState() bool {
    currentHideMenus := hideMenus
    hideMenus = previousHideMenus
    
    return hideMenus
}
```

Then all you'd have to do is

```
setHideMenusState(true)    
_, err := syncDb.FindSatisfier(target.DepString())
popHideMenusState()
```

Just wanted to point out what I was aiming for, I made the backup/restore changes you suggested anyway.

Morganamilo:
That doesn't seem like a too bad way to do things but I can imagine it subtlety failing If i were to call setHideMenusState() at the start of my function, then also call inRepos(), then try to restore it.

Anyway the current PR looks perfect. Thanks for the help.

PR #778, opened 2018-10-18
 Allow many aur packages to be installed at once.

Morganamilo:
Allow many aur packages to be installed at once.
After building an AUR package don't install it right away. Instead queue
it for install.

Then when a package comes along that does not have all of its
dependencies satisfied, install the queued packages before builting.

This allows a standard AUR update to be done with one transaction, while
building AUR packages will long dependency chains will work as normal.

---

This depended on #767 so now that it's merged I can open this. I'm a little unsure of it as It fear it make .so issues more likely.

But then again updates are unordered anyway so it's possible that `b` depends on `a` and our -Syu updates b first then a updates and gets a .so bump. This as not been a problem so maybe this won't be either.

I know @maximbaz really wanted this before he moved to aurutils ;)

### DIFF
--- a/clean.go
+++ b/clean.go
@@ -45,7 +45,7 @@ func cleanRemove(pkgNames []string) (err error) {
 		return nil
 	}
 
-	arguments := makeArguments()
+	arguments := cmdArgs.copyGlobal()
 	arguments.addArg("R")
 	arguments.addTarget(pkgNames...)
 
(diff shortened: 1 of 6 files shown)


Jguer:
If this is open again, I guess it's time for a maintenance release

Morganamilo:
Including https://github.com/Jguer/yay/pull/961? Seems a big an very new change. I've also not read the diff at all so have no idea how it works.

Saying that you can create a release from any commit. So merging this anyway and you can choose to not include it in the release.

Jguer:
#961 is a small change, fixes the issue just partially for the very specific case and not the general case, so it can pass, I'll tag that commit as `9.2.1`. 

This one I think is the one that needs to steep more in `-git`

PR #791, opened 2018-10-30
Modified readme to show extra dependency

ZenixTheHusky:


Jguer:
True, forgot to add that in the merge. @Morganamilo I think it's time for 8.1192.0 before jumping over to 9 with the new config.

PR #819, opened 2018-12-01
Fix tests broken in master

Jguer:
- Fix tests broken in master
- Transition to semantic version 8.2.0

I think next release we can simplify the makefiles and PKGBUILDs for semantic versioning, right now it won't do it any harm, but if your configuration reflection refactor is a bit late (I guess it's a major pain in the ass and it isn't inspiring to work on, been there :sweat_smile: ) I would still like to transition yay to go mod and bring go-alpm's renaming to yay which is going to cause a lot of merging nightmare's


Jguer:
I put it in the commit description but forgot to add it here, in those files were only benchmarks of printing topdown and bottomup, I confirmed performance was not affected by the sorting but never removed the benchmarks (they were never ran on make all)

Morganamilo:
That makes sense. Might want to merge #705 at the same time as this. 1.11.2 is here, it does still build stdlib stuff but I guess that might just be intentional. It doesn't seem enough of a hindrance for us to wait even more.

Morganamilo:
Forgot to mention this should be a 9.0 release, otherwise vercmp will be broken because 8.1000 > 8.2.

And uh I don't know how to edit a review ¯\_(ツ)_/¯

Jguer:
True, it will break vercmp forgot that. I can justify it as 9.0 by including go modules

PR #822, opened 2018-12-06
fix copy-paste error

sseering:
trivial change


Jguer:
Thanks for taking the time to fix it after noticing

PR #832, opened 2018-12-12
Match the query date/time to pacman. Fixes #831

### LINKED ISSUE #831: Time part info is missing
#### Affected Version
yay v9.0.1.r0.gd3a9082 - libalpm v11.0.1

#### Issue
The datetime information from `-Si` alike commands contains only ISO formatted date, the time information is absent. See the attached screenshot. Other popular AUR helpers do not have this issue.

#### Possible resolution
It would be reasonable to have a datetime format according to local system and user settings (like `trizen` does), with a safe fail-back displaying format, either in accordance with `pacman` look, or, at least, as a standard ISO format with UTC summertime independent timezone.

yay1
yay2

parkerlreed:
Pacman's query prints the date/time information in localtime. Changed Yay's to match.

### DIFF
--- a/print.go
+++ b/print.go
@@ -227,8 +227,8 @@ func PrintInfo(a *rpc.Pkg) {
 	printInfoValue("Maintainer", a.Maintainer)
 	printInfoValue("Votes", fmt.Sprintf("%d", a.NumVotes))
 	printInfoValue("Popularity", fmt.Sprintf("%f", a.Popularity))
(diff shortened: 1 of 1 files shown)


Morganamilo:
spaces -> tabs ;)

parkerlreed:
I even noticed the spaces. Force of habit :) I clean it up real quick.

PR #854, opened 2019-01-07
add bash completion for -G,--getpkgbuild

1ace:


Jguer:
Seems all good, thanks for your contribution

PR #866, opened 2019-01-15
Revert "Merge pull request #767 from Morganamilo/merge-dep-pool-order"

Morganamilo:
This reverts commit c2aeb4bc4eea0a7870ac1e3712dad4d1459b70f1, reversing
changes made to 1f1cee70239a3dffa1ac3a243812a01aa1699c53.

This should fix the ton of recent issues which I assume are because of the depsolver changes. This will bring back the inner conflicts issues that prevent some circular dependencies from resolving. But they are rare enough compared to the ton of open issues right now.

Due to lack of time I'm just reverting the code, if I have time in future (or some one else wants to take on the task) then the code can be fixed and merged back in.


Jguer:
I'm also in the same situation but I'll take the revert until this is sorted. I'm merging this, counseling every open thread to retry using this version and then we'll go from there

Morganamilo:
I'm all for a release now. There's not too much going on right now and this fix is pretty important.

Jguer:
Let's go for it then

PR #897, opened 2019-03-01
remove unnecessary assignment

sosiska:
Each iteration of "range" instruction already makes copy of element, so 
```go
i := _i
```
instructions is unnecessary in this case.

### DIFF
--- a/download.go
+++ b/download.go
@@ -216,7 +216,7 @@ func getPkgbuildsfromABS(pkgs []string, path string) (bool, error) {
 
 		if pkgDB != "" {
 			if db, err := alpmHandle.SyncDBByName(pkgDB); err == nil {
-				pkg  = db.Pkg(name)
+				pkg = db.Pkg(name)
 			}
 		} else {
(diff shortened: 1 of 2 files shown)


Morganamilo:
Are you sure this doesn't break stuff? I don't understand why I had to do this but I remember having some pointer weirdness and that ended up fixing it.

sosiska:
Tour of go says us https://tour.golang.org/moretypes/16 that:
```When ranging over a slice, two values are returned for each iteration. The first is the index, and the second is a copy of the element at that index.```
So current master code 1st makes copy of element, than makes copy of copy of this element, and than makes pointer to this copy of copy of this element.

sosiska:
You are right, pointer of _i element will be same on each iteration:
https://play.golang.org/p/riJ297NmyqX
But we can do differently:
https://play.golang.org/p/Uu0HoKI38it

Give me a moment.

Morganamilo:
Are you planning on editing the PR to implement this?

PR #901, opened 2019-03-04
rewrite if-else to switch statement

sosiska:
This commit rewrites some if-else-if-else chains as a switch.
Based on Go style guide:
https://golang.org/doc/effective_go.html#switch


Jguer:
I do enjoy the extra readability, good job, merging

PR #903, opened 2019-03-05
code improvements

sosiska:
Some code improvements.

### DIFF
--- a/clean.go
+++ b/clean.go
@@ -68,7 +68,7 @@ func syncClean(parser *arguments) error {
 	}
 
 	if mode == modeRepo || mode == modeAny {
-		if err = show(passToPacman(parser)); err != nil {
+		if err := show(passToPacman(parser)); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


sosiska:
Maybe replace
```go
if err = alpmHandle.SetSomeStuff(pacmanConf.SomeParam); err != nil {
		return err
}
```
to 
```go
if err = alpmHandle.SetSomeStuff(pacmanConf.SomeParam); err != nil {
		return err
}
```
and remove err declaration at the top of configureAlpm function?

Morganamilo:
You're free to do that if you wish, would stop an assignment from being missed like that again.

sosiska:
Done. I just wanted to avoid over allocating memory, but this is not a high load application. So now it looks prettier!
Should I squash commits or it's OK?

Morganamilo:
Honestly I don't even know if it does allocate memory. And there's no need to squash them really. Merge commits and fixups I like to be squashed but small commits in general are fine.

PR #908, opened 2019-03-07
Fix function comments based on best practices from Effective Go

CodeLingoTeam:
Hi, we updated an exported function comment based on best practices from [Effective Go](https://golang.org/doc/effective_go.html#commentary). It’s admittedly a relatively minor fix up. Does this help you?

### DIFF
--- a/print.go
+++ b/print.go
@@ -189,7 +189,7 @@ func (u upSlice) print() {
 	}
 }
 
-// printDownloadsFromRepo prints repository packages to be downloaded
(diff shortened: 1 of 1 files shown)


Morganamilo:
It's an improvement so I'll take it. Although there's plenty of other places where the documentation is lacking so it's a little weird for you to just fix this one specifically.

I'll merge this if you like, but if you plan to do more changes I would prefer you just amend this PR.

PR #909, opened 2019-03-07
Some more code improvements

sosiska:
Some more code improvements.

### DIFF
--- a/config.go
+++ b/config.go
@@ -288,8 +288,7 @@ func continueTask(s string, cont bool) bool {
 
 	fmt.Print(bold(green(arrow)+" "+s), bold(postFix))
 
-	len, err := fmt.Scanln(&response)
-	if err != nil || len == 0 {
+	if _, err := fmt.Scanln(&response); err != nil {
 		return cont
 	}
 
--- a/download.go
+++ b/download.go
@@ -164,15 +164,16 @@ func getPkgbuilds(pkgs []string) error {
 		for _, base := range allBases {
 			name := base.Pkgbase()
 			_, err = os.Stat(filepath.Join(wd, name))
-			if err != nil && !os.IsNotExist(err) {
+			switch {
+			case err != nil && !os.IsNotExist(err):
 				fmt.Fprintln(os.Stderr, bold(red(smallArrow)), err)
 				continue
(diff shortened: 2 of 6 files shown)


Morganamilo:
If you're gonna keep going through the code I would suggest you keep at it and make one bigger than a bunch of tinny ones.

sosiska:
OK, i will look for. Then i will squash commits

sosiska:
I think that all(no more suitable if-else-if-else chains 🙃). Look at the separate commits to see what I have changed. Then, if there are OK, I will squash them to big one.

config.go
```diff
@@ -288,8 +288,8 @@ func continueTask(s string, cont bool) bool {
 
 	fmt.Print(bold(green(arrow)+" "+s), bold(postFix))
 
-	len, err := fmt.Scanln(&response)
-	if err != nil || len == 0 {
+	i, err := fmt.Scanln(&response)
```
Jguer: Maybe `len` was not the best descriptor for the return, but I think `i` is not much better. Maybe the `n` which is the same as the function header `func Scanln(a ...interface{}) (n int, err error)` would be a better choice.

Morganamilo: >It returns the number of items successfully scanned. If that is less than the number of arguments, err will report why.

Maybe we don't even need the variable.

Jguer: True, maybe it can even be removed (if it's like that it would never evaluate it anyway)

sosiska: You are right, judging by the function description, the second comparison will not happen. Variable removed in 0347d1d3d69671a7f482193a1b5e25ef5bd0afc8

sosiska:
Squashed.

depCheck.go
```diff
@@ -9,7 +9,7 @@ import (
 	alpm "github.com/jguer/go-alpm"
 )
 
-func (dp *depPool) checkInnerConflict(name string, conflict string, conflicts mapStringSet) {
```
Morganamilo: I'm not sure what's more idiomatic for go but my personal opinion is to only skip the return types if all returns are the same type.

sosiska: OK. Combined parameters in some functions are rejected in 14ac7562981a20bbc0a91c7fcdafc7a1c6b6ab9c.

sosiska:
Ping

Jguer:
I'm ok with it and merging, thanks for the contribution @sosiska

PR #916, opened 2019-03-20
Add SudoBin

daftaupe:
Hello,

this adds the possibility to use a custom binary for sudo instead of the default one (that might also make yay usable for people using doas or other commands instead of sudo for example, the sudoloop won't work though in that case), much like what's done for GpgBin and GitBin.
I've added a default value so that it makes no difference for most users.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -105,6 +105,8 @@ Permanent configuration options:
     --combinedupgrade     Refresh then perform the repo and AUR upgrade together
     --nocombinedupgrade   Perform the repo upgrade and AUR upgrade separately
 
+    --sudo                <file>  sudo command to use
+    --sudoflags           <flags> Pass arguments to sudo
     --sudoloop            Loop sudo calls in the background to avoid timeout
     --nosudoloop          Do not loop sudo calls in the background
 
--- a/config.go
+++ b/config.go
@@ -58,6 +58,8 @@ type Configuration struct {
 	SortBy             string `json:"sortby"`
 	GitFlags           string `json:"gitflags"`
 	RemoveMake         string `json:"removemake"`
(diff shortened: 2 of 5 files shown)


Morganamilo:
Looks about right. The man page should probably mention that the use of a custom sudo command will probably break sudoloop.

Morganamilo:
Actually looks like you missed an option for SudoArgs. This would make it possible to use `su -c`.

daftaupe:
@Morganamilo shouldn't it be SudoFlags ? As there already is GitFlags, GpgFlags, etc.

Jguer:
Seems about right to me as well, just missing the `SudoFlags`

daftaupe:
> Actually looks like you missed an option for SudoArgs. This would make it possible to use `su -c`.

It's not so easy (or let's say I'm not that good ad go) as su -c doesn't work without quotes such as `su -c 'pacman -Syu'` for example.

I'm pushing a first try for SudoFlags, it works with some options, and doesn't work with others.
I

Jguer:
Merging this and we'll look at fixing these integrations when they are needed, already this does what vanilla does.

Jguer:
Actually, this branch is broken with the `SudoFlags` as the empty string gets appended and sudo tries to run an empty command

daftaupe:
Hmm, yes this wasn't ready :/ looks like it's harder than it looks to handle sudo's arguments.

Morganamilo:
All you need to do is call `SplitWhitespace`.

daftaupe:
Ok @Morganamilo, but now the PR is closed :/ should I open a new one ?

Morganamilo:
Oh how times flies. But yeah, once merged I can't reopen this.

PR #931, opened 2019-04-11
Added --copyleft-only argument

0v3rl0w:
I thought about people that are using distros like Parabola. So I decided to code a kind of arguments to check if a package is copylefted or not. I don't have all the copyleft license in the code, it will need to complete.

Any comments and hints are welcomed

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -36,6 +36,7 @@ New operations:
 New options:
        --repo             Assume targets are from the repositories
     -a --aur              Assume targets are from the AUR
+       --copyleft-only    Will just allow copylefted software
 
 Permanent configuration options:
     --save                Causes the following options to be saved back to the
--- a/depCheck.go
+++ b/depCheck.go
@@ -9,6 +9,44 @@ import (
 	alpm "github.com/jguer/go-alpm"
 )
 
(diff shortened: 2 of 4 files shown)


Morganamilo:
This seems like a weirdly specific thing to add to yay. Parabola patches pacman already iirc. Wouldn't it be better for them to patch makepkg on their end? That would then work with any pkgbuild, from any source, with any helper.

0v3rl0w:
I've never tried parabola, but I tought that could be a good idea to implement this, after if you think that's not that great, no problem ^^

Morganamilo:
Actually thinking about it this could possibly work as a libmakepkg drop in. So no need to patch pacman. If you're truly interested in the feature that's probably the way to go. I'd rather not feature creep yay any more.

0v3rl0w:
I think I don't really understand your thought

Morganamilo:
makepkg supports drop in files to add functionality. That would probably be the way to implement this.

0v3rl0w:
Ok I see Thank you

AladW:
Man, reading these reports is more fun than any soap opera.

PR #935, opened 2019-04-14
Speedup git clone by setting clone depth

0mco:
There seems no need to clone the entire commit history.

### DIFF
--- a/download.go
+++ b/download.go
@@ -63,7 +63,7 @@ func gitHasDiff(path string, name string) (bool, error) {
 func gitDownload(url string, path string, name string) (bool, error) {
 	_, err := os.Stat(filepath.Join(path, name, ".git"))
 	if os.IsNotExist(err) {
-		cmd := passToGit(path, "clone", "--no-progress", url, name)
+		cmd := passToGit(path, "clone", "--depth=1 --no-progress", url, name)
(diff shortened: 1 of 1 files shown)


Morganamilo:
Does this actually give any measurable speedup? AUR repos are already tiny as is. And It may be useful for some to want to read through older commits.

0mco:
@Morganamilo I'd say yes in China. The connection to GitHub is extremely poor. AUR repos is tiny, but it may need to clone from other git server (e.g. GitHub).
`yay` is for building and installing packages. Reading commit history via `yay` doesn't make sense.

Morganamilo:
> but it may need to clone from other git server (e.g. GitHub).

No, yay only clones AUR packages. Makepkg is responsible for downloading sources.

>yay is for building and installing packages. Reading commit history via yay doesn't make sense.

I'm not talking about using yay to read the history. But seems as yay maintains a cache you may as well use that cache to read the history that a duplicate clone some where else.

0mco:
I think I messed up something, cloning from GitHub isn't done by `yay`?

Morganamilo:
No it is not. And even if it was `depth=1` would be a big no no seems as packages need the history to checkout old commits or describe.

0mco:
Got it.

PR #939, opened 2019-04-23
Fixups

Morganamilo:


Jguer:
When I saw `fixups` I was afraid for a moment that it would be a pain to rebase on my modules branch but it's nothing serious. 

The case sensitive imports are a sore to the eyes though

Morganamilo:
I remember the case being an actual problem for me. Mainly because for go commands I tend to copy the url from the browser and paste it into the terminal. This tends to conflict with the lower case version.

Jguer:
Yes, specially if you `go get` different cases it will create 2 separate `go/src/...` and github always seems to default to the capital case url.

PR #942, opened 2019-05-01
Add SudoBin and SudoFlags

daftaupe:
Hello,

this adds the possibility to use a custom binary for sudo instead of the default one (that might also make yay usable for people using doas or other commands instead of sudo for example, the sudoloop won't work though in that case), much like what's done for GpgBin and GitBin.
I've added a default value so that it makes no difference for most users.

This time I've started a draft PR :)

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -45,7 +45,7 @@ func sudoLoop() {
 
 func updateSudo() {
 	for {
-		err := show(exec.Command("sudo", "-v"))
+		err := show(exec.Command(config.SudoBin, "-v"))
 		if err != nil {
 			fmt.Fprintln(os.Stderr, err)
 		} else {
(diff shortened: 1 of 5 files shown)


exec.go
```diff
@@ -45,7 +45,7 @@ func sudoLoop() {
 
 func updateSudo() {
 	for {
-		err := show(exec.Command("sudo", "-v"))
+		err := show(exec.Command(config.SudoBin, "-v"))
```
Morganamilo: Shouldnt sudoflags also be used here?

daftaupe: Well, I thought about it, but I don't know, the only purpose of that function is to refresh the sudo user's cached credentials. But thinking about it know, I guess the SudoFlags should be here too, you're right.

Morganamilo: Seems there's no reason not to. May want to make a wrapper around sudo in exec.go.

Issue #936, opened 2019-04-15
[FR] Search in package names only


efermi:
#### Affected Version
yay v9.1.0 - libalpm v11.0.3

#### Issue
It would be useful to have a built-in way to search in package names only. It's a somewhat frequent request too: https://bbs.archlinux.org/viewtopic.php?id=90164, https://bbs.archlinux.org/viewtopic.php?id=54684, https://bbs.archlinux.org/viewtopic.php?id=236426, https://bbs.archlinux.org/viewtopic.php?id=92102

hopeseekr:
So how do you search yay by package name???

grandchild:
It seems what @hopeseekr actually means is that `--searchby name` doesn't actually work, and returns description hits as well:

```
$ yay --searchby name something
5 aur/something r2894.95ca07bd-1 (+0 0.00) 
    Finally something
4 aur/something-for-reddit-git 0.1-1 (+3 0.00) (Out-of-date: 2019-12-24) 
    A Reddit Client For GNOME (with Gtk+ and Python)
3 aur/something-for-reddit 0.2.2-1 (+10 0.03) (Out-of-date: 2020-10-01) 
    A Reddit Client For GNOME (with Gtk+ and Python)
2 community/ruby-httpclient 2.8.3-7 (360.9 KiB 1.7 MiB) 
    gives something like the functionality of libwww-perl (LWP) in Ruby
1 community/python-hypothesmith 0.1.5-1 (17.6 KiB 46.6 KiB) 
    Hypothesis strategies for generating Python programs, something like CSmith
```

PR #948, opened 2019-05-07
FR #936: Adds config parameter for searching in package names only and not in descriptions

nikosk686:
Implements FR #936.
`--pkgnameonly` and `--nopkgnameonly` config params have been added
As config param, each option can be permanently stored if combined with `--save`

Example usage: `yay -Ss --pkgnameonly pkg1 pkg2 ...`

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -104,6 +104,8 @@ Permanent configuration options:
     --nouseask            Confirm conflicts manually during the install
     --combinedupgrade     Refresh then perform the repo and AUR upgrade together
     --nocombinedupgrade   Perform the repo upgrade and AUR upgrade separately
+    --searchby	<field>	  Filter search results by specific field (currently only 'name' is supported)
+    --nosearchby		  Clear search field filter, search in all package fields
 
     --sudoloop            Loop sudo calls in the background to avoid timeout
     --nosudoloop          Do not loop sudo calls in the background
--- a/completions/bash
+++ b/completions/bash
@@ -57,7 +57,8 @@ _yay() {
           search unrequired upgrades' 'c e g i k l m n o p s t u')
   remove=('cascade dbonly nodeps assume-installed nosave print recursive unneeded' 'c n p s u')
   sync=('asdeps asexplicit clean dbonly downloadonly force groups ignore ignoregroup
-         info list needed nodeps assume-installed print refresh recursive search sysupgrade'
+         info list needed nodeps assume-installed print refresh recursive search sysupgrade
+         searchby nosearchby'
         'c g i l p s u w y')
   upgrade=('asdeps asexplicit force needed nodeps assume-installed print recursive' 'p')
   common=('arch cachedir color config confirm dbpath debug gpgdir help hookdir logfile
--- a/completions/fish
+++ b/completions/fish
@@ -208,6 +208,8 @@ complete -c $progname -n "$sync; and __fish_contains_opt -s u sysupgrade" -s u -
 complete -c $progname -n $sync -s w -l downloadonly -d 'Only download the target packages'
 complete -c $progname -n $sync -s y -l refresh -d 'Download fresh copy of the package list'
 complete -c $progname -n "$sync" -xa "$listall $listgroups"
+complete -c $progname -n $sync -s w -l searchby -d 'Filter search results by specific field (currently only 'name' is supported)'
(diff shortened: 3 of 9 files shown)


Morganamilo:
When I was planning to do this I was thinking of a generic `--searchby` flag. I would much prefer that solution. Also you should update the `aur` dependency.

Morganamilo:
Reading the patch, it seems you manually filter the package list yourself. The AUR RPC supports searching by many different fields. For repo packages i'm fine with `--searchby` only applying to the AUR.

nikosk686:
Didn't know about search by field functionality in AUR RPC.
I will change the code to use this.
So as if I understand correctly, you suggest my solution of `--pkgnameonly` flag to remain as it covers both cases and moreover, the addition of the `--searchby` option, to cover only aur packages.
Is this assumption right?

Morganamilo:
My opinion is that as we are primarily an AUR helper, only having `--searchby` work for the AUR is fine.

For repo searching we use alpm's search function. Unless alpm has a native function for searching by name only, i'm fine on not having the `--pkgnameonly` flag.

nikosk686:
I see your point but I think it would be more useful to have this filtering option in both aur and repo packages. Otherwise, it would seem half-done feature in my opinion.
I do not believe alpm offers this feature though. That's why I handle this by filtering out irrelevant results with a custom solution in yay.
Unless you are against this idea, I would suggest 
1) to keep `--pkgnameonly` flag (I will fix the way aur packages are being filtered by using the native AUR RPC 'search by field' solution)
2) to create a new FR for extended `--searchby` functionality only for AUR packages, as you suggest. I could implement this feature, too.

If you do not want 1, I will not insist. We could go just with 2, but in any case I suggest we open a new FR for this as #936 seems a bit irrelevant.

Morganamilo:
I don't like `--pkgnameonly` at all. Not only because the point of yay is to be an AUR helper but there are already penty of other tools out there that can do this like `pacsift`.

Jguer:
I don't like the name `--pkgnameonly` personally, it's quite not a boolean choice, it's a mode choice on how to filter packages (`--searchfilter=name\namedesc`?) . 

Apart from that, I can see why this feature is needed for the repos actually (`yay st terminal`being my normal workaround`), so in this particular case, if pacman doesn't do it, it's an ux improvement we can deliver, and should be implemented for AUR and Repos so it's not half baked. 

For me this PR would be fine with the change from a boolean choice to a mode setting.

Morganamilo:
https://github.com/mikkeloscar/aur/blob/f956f121b3fa2bdaec629d547f3b6ecb8f68996d/aur.go#L24

These are the seach bys that the AUR RPC supports. I really don't want to support all of these for the repos. So either way it's going to be half baked.

Morganamilo:
Also if we did enable this for the repos then searching for st would still be pretty pointless. See `pacsift --name st `

nikosk686:
@Jguer I understand your point, too. But, we should consider that in real life package name is a more "strong" field than the others when searching..
When I want to do a quick search for a package those that their name matches to my query strings are more likely to be the true positives..I wouldn't search by the author's name, for a quick search, for example.
That's why I think the boolean flag for searching in package name (as a more weighted field) is fine IMO.
That's why the author of #936 asked for package names only and not for a parameterized search filed option, also IMO.

nikosk686:
Hello,
I have updated the PR, according the above conversation.
Specifically,
- I 've changed the boolean `--pkgnameonly` parameter with the `--searchby`. `--searchby` accepts values, but I 've currently implemented support only for `name` field.
- I 've changed aur filtering to use the native aur RPC `SearchByName` method when this is the case.

The support to filter Repo search results remains, and the filtering is still performed in yay's code after all results have been returned from alpm.
I know that @Morganamilo does not want this, but as @Jguer seems to agree with me that it is a useful feature, I leave it up to you two to decide.

Jguer:
@Morganamilo you're right that it wouldn't do much. Actually a new sorting algorithm that sorts according to percentage match of name would give a better result. ("st" - st = 100% "st" - st-info = 28.5%) 

On this particular pull request, it's useful but I probably won't use it and I don't know how many more users are going to change the pretty sane default for it. As it is, If there's more people wanting it I'm willing to merge it. 
Otherwise I trust @Morganamilo s judgement on whether to scrap it or take it to full "searchby" for AUR (but keeping the "searchby=name" as well for repo).

PR #950, opened 2019-05-10
Display what file caused clean errors.

Psykar:
Previously if a clean failed, you'd get an error, but wouldn't know what
directory caused the error, requiring a bunch of manual checking of each
directory until you found it.
This adds an error message displaying which folder failed to clean.

It also continues cleaning the other directories even if one failed.

### DIFF
--- a/clean.go
+++ b/clean.go
@@ -188,8 +188,9 @@ func cleanUntracked() error {
 
 		dir := filepath.Join(config.BuildDir, file.Name())
 		if shouldUseGit(dir) {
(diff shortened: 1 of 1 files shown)


Morganamilo:
What's the reasoning for the capture instead of keeping show and printing an error message after? Also what is the error message git prints and under what circumstances does it happen?

Jguer:
Closing due to staleness, request has been implemented in ab3c66ab98d209e8660068c0

PR #961, opened 2019-06-11
Defer cleanAfter and removeMake

### LINKED ISSUE #785: Yay will not remove make deps after a failed aur installation.
#### Affected Version
Not sure. I'm using `yay v8.1173.0.r17.g431118b - libalpm v11.0.1`
#### Issue
For example, I want to install a aur package named `aur-package-a` by `yay -S aur-package-a`. And yay tell me the aur package has a make dependence named `makedep-a`. Then I choose "y" for 
`Remove make dependencies after install?`. After that, yay installed the `makedep-a` automatically and successfully. But in the procedure of downloading source of `aur-package-a`, I get some network error and yay quits. In this case, the `makedep-a` will not be automatically removed after  you installed  `aur-package-a` again and successfully.

FryDay:
This should fix #785.


Jguer:
Code style seems ok, defers always add some funny bugs so I'll test it out and check the changes on the `go.sum` when I'm long enough in front of a dev platform with Arch. Just acknowledging this will be reviewed.

Jguer:
I tested against `yay -S deluge-stable-git` and killing the makepkg in the middle with a TERM signal and it seems to work fine. Of course when it fails and you want to rebuild it, the make dependencies will have to be reinstalled again but they'll be in cache, so not so worrisome. I'll merge it and give a bit of time for complaints.

PR #967, opened 2019-06-18
Use modules with gcc-go9.1

Jguer:
Waiting on https://bugs.archlinux.org/task/62934, but working even without it on gcc-go 9.1.0-1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,7 @@ BINNAME := yay
 PACKAGE := ${PKGNAME}_${VERSION}_${ARCH}
 
 ifneq (,$(findstring gccgo,$(GOCC)))
-	export GOPATH=$(shell pwd)/.go
 	LDFLAGS := -gccgoflags '-s -w'
-	MOD :=
 endif
 
 default: build


Jguer:
gcc-go 9.1.0-2 is in testing now with the updated provides.
so next version of yay should move to requires `go>=1.11'

Makefile
```diff
@@ -18,9 +18,7 @@ BINNAME := yay
 PACKAGE := ${PKGNAME}_${VERSION}_${ARCH}
 
 ifneq (,$(findstring gccgo,$(GOCC)))
```
Morganamilo: Is this ifneq still needed? I don't remember it being needed before moving to modules.

Jguer: It builds but it will always complains of `go build: when using gccgo toolchain, please pass compiler flags using -gccgoflags, not -gcflags` and I think it ignores the arguments after because this: `-X main.version=${VERSION}'` is not supported by gcc-go.

I would like to remove that ifneq but I haven't found a suitable solution

PR #969, opened 2019-06-19
print: out-of-date format

simon04:
Consistently use the same date format for all dates.

### DIFF
--- a/print.go
+++ b/print.go
@@ -308,7 +308,7 @@ func PrintInfo(a *rpc.Pkg) {
 	printInfoValue("Last Modified", formatTimeQuery(a.LastModified))
 
 	if a.OutOfDate != 0 {
-		printInfoValue("Out-of-date", "Yes ["+formatTime(a.OutOfDate)+"]")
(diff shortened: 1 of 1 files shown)


Jguer:
Then isn't it inconsistent with the search result out of date `toprint += bold(red("(Out-of-date "+formatTime(res.OutOfDate)+")")) + " "`? 

Also removing the yes from it would make more sense if then the date is printed in red() imo

simon04:
True. I guess at this place, one would want a concise (date only) output.

I find the inconsistency within the `-Si` output more distracting (and would happily accept the inconsistency between out-of-date outputs of very different commands):
```
$ yay -Si tippecanoe
...
First Submitted : Wed 17 May 2017 08:10:33 PM CEST
Last Modified   : Tue 16 Apr 2019 12:55:38 AM CEST
Out-of-date     : Yes [2019-05-30]
```

Jguer:
It's well justified :wink:, I'm merging it

Morganamilo:
Does this make format time an unused function?

Jguer:
It's still used for search

PR #1000, opened 2019-07-31
Correct package range in error message

kiendang:
```
# ./yay -S nvidia-docker
:: There are 2 providers available for nvidia-container-runtime-hook:
:: Repository AUR
    1) nvidia-container-runtime-hook 2) nvidia-container-runtime-hook-bin

Enter a number (default=1): 3
error: invalid value: 3 is not between 1 and 3
```

should be `3 is not between 1 and 2`.

### DIFF
--- a/print.go
+++ b/print.go
@@ -663,7 +663,7 @@ func providerMenu(dep string, providers providers) *rpc.Pkg {
 		}
 
 		if num < 1 || num >= size {
-			fmt.Fprintf(os.Stderr, "%s invalid value: %d is not between %d and %d\n", red("error:"), num, 1, size)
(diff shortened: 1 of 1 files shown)


Jguer:
Depends if it's [1,3] (contains 1,2,3) , [1,3[ (contains 1,2 like go slices) or ]1,3[ (contains 2) :laughing:, but it's true for users normally [,] is assumed

kiendang:
Right. I didn't think of all that. My mind just assumes [,] by default haha. Also in SQL `BETWEEN 1 AND 3` means [1,3] so I guess going with [,] is safest here.

PR #1003, opened 2019-08-15
improved install instructions

gcb:
was helping friends use the project, and those two points where common issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,14 @@ cd yay
 makepkg -si
 ```
 
+### sudo
+
+the `-si` flag on makepkg will try to use sudo to install the package. If your user can't use sudo, ommit the flag, and install the package that is generated on the source directory with `pacman -U <file>`
+
+### dependencies
+
+yay makepkg depends on `base-devel` and `go`, which can be installed via pacman.
+
 ## Support
 
 All support related to Yay should be requested via GitHub issues. Since Yay is not


README.md
```diff
@@ -43,6 +43,14 @@ cd yay
 makepkg -si
 ```
 
+### sudo
```
Jguer: can you just change this to `####` to match the FAQs?

README.md
```diff
@@ -43,6 +43,14 @@ cd yay
...
+
+the `-si` flag on makepkg will try to use sudo to install the package. If your user can't use sudo, ommit the flag, and install the package that is generated on the source directory with `pacman -U <file>`
```
Jguer: `omit` instead of `ommit`.

README.md
```diff
@@ -43,6 +43,14 @@ cd yay
...
+
+### dependencies
+
+yay makepkg depends on `base-devel` and `go`, which can be installed via pacman.
```
Jguer: `base-devel` is always assumed as installed for AUR use, so it shouldn't be here. Also `-i` in makepkg installs but `-s` installs missing dependencies, the way this is written will make a user install `go` as a explicit package instead of a make dependency.

If your sudo is really borked `# pacman -S go --asdeps` , `$ makepkg`, `# pacman -U yay*.pkg.*` is how I would do it. 

I can't merge this if it counsels users to leave a 500MB package they won't use in their system

gcb: That makes sense. I added those because, well, i was helping people that didn't really wanted to learn much about builds but wanted to use AUR packages, which I assumed was the audience for yay.

One of the people I was helping had a good linux knowledge because of 'devops'ing as a webdeveloper, so they were wary of using sudo all over the place.

let me rethink all that and redo this PR. I am not a user of yay myself and now i forgot why i thought `go` was needed at that point and not as dependency of something...

gcb:
will do over and reopen.

README.md
Morganamilo: >i was helping people that didn't really wanted to learn much about builds but wanted to use AUR packages, which I assumed was the audience for yay.

I'd rather it be the opposite. Yes yay is stupid easy to use but that's just because I like using a simple tool. That doesn't mean I want people to use it without having some understanding of the AUR.

I'd rather not lower the barrier to entry even further.

Issue #1006, opened 2019-08-23
Variables not preserved between functions.


bartoszek:
I'm having this stupid issue with one of my packages ([magma](https://aur.archlinux.org/packages/magma/)). Users keep reporting that binaries are installed in `/usr/local/magma` instead of `/opt/magma` path. In spite that `PKGBUILD` have explicitly stated `CMAKE_INSTALL_PREFIX=/opt/magma` in `prepare()` function. 

I've just tested this in clean chroot, and it looks like `yay` ~in fact somehow is changing cmake install prefix~ is not preserving variables between `prepare()` and `build()`. This is really strange as plain `makepkg` is working as intended :confused: 

I'm not a `yay` user and a compleat noob in `go`, can't pin point the issue further...

Jguer:
`Yay` splits prepare() and build(), so it really is as you diagnose, it's a conscious choice for implementation. `makepkg` supports this type of operation, so it comes down to `PKGBUILDs` supporting it or not

bartoszek:
Yep, but you could also drop `--noextract --noprepare` flags from last `passToMakepkg` and everyone would be happy. It isn't like you are altering sources any way :thinking:

Morganamilo:
If you dropped those flags then you'd end up with problems from prepare being ran twice (as well as the added time for extracting twice).

As far as I'm concerned pkgbuilds should declare their variables in every function where they need them and to do otherwise is incorrect.

Not doing so also breaks useful makepkg features like `--repackage`.

bartoszek:
You've got a point there. Perhaps moving variable logic outside of functions would be a proffered solution, despite it's ugly and hacky :disappointed:

Morganamilo:
I didn't say outside of functions but instead inside each one. Not sure if there's a difference there but I believe that to be the recommended way to do it.

bartoszek:
And here lays the problem.
* If you specify variable logick outside function (e.g `_pyver=$(expac -Q "%v" python)`). You can't use it with `makepkg -s` or `aurutils` without having `expac` preinstalled, but it would work with `yay` because of dependency resolver strictly following output from aur-rpc.
* If you move it inside prepare(), it would work with `makepkg` and by extend `aurutils` but won't work with `yay`.
* To make it work with all above you have to also put `_pyver` logic in in every function it gets utilized (prepare,build,package,package-doc)

As an `aurutils` user I tend toward second solution, but as `yay` is preferred by majority of users I will have to adopt third one :confounded:

[...]

Issue #1004, opened 2019-08-16
Environment variables, export ... gets ignored


LinuxUserGD:
#### Affected Version
yay v9.2.1.r13.gc9f7105 - libalpm v11.0.3
#### Issue
How can I use environment variables with yay when compiling AUR packages?
Example: https://aur.archlinux.org/packages/blender-2.7/
This only works with makepkg and not with ỳay, it gets ignored (still installs cuda) or says "invalid package DISABLE_CUDA=1"

#### Steps to reproduce

1. export DISABLE_CUDA=1 before build
2. DISABLE_CUDA=1 ~your-aur-helper~
3. makepkg DISABLE_CUDA=1

LinuxUserGD:
Editing PKGBUILD also doesn't work

Jguer:
`env DISABLE_CUDA=1 yay -S blender-2.7` does pull cuda as a makedep.

`blender-2.7` [.SRCINFO](https://aur.archlinux.org/cgit/aur.git/tree/.SRCINFO?h=blender-2.7) mentions cuda as a makedep so any AUR helper that parses `.SRCINFO` will trip on that.
yay uses [RPC](https://aur.archlinux.org/rpc/?v=5&type=info&arg[]=blender-2.7) to handle makedeps , which the package also includes cuda in. 

Adding/Removing makedeps on runtime through the `PKGBUILD` is not supported by yay (arguable if it's best practice from what I've seen), but it is not an issue of Environment Variables, just something that is not implemented.

bartoszek:
Sorry, but I'm not going to adjust `PKGBUILD` for every AUR helper out there. If `yay` diverge form pristine `makepkg` in case of deps handling there's not much I can do :disappointed:
You could always try `aurutils` with -T swich, which just starts `makechrootpkg` and so is immune to this problem.

bartoszek:
@LinuxUserGD Use `yay --mflags "DISABLE_CUDA=1"` to pass configuration variable to the `PKBUILD`

LinuxUserGD:
@bartoszek 
I tried this:
```
[user @ archlinux blender-2.7] $ yay --mflags "DISABLE_CUDA = 1"
[user @ archlinux blender-2.7] $ yay -S blender-2.7 --noconfirm
:: Checking for conflicts ...
:: Checking for inner conflicts ...
[...]

Issue #893, opened 2019-02-21
user variables set in perpare() ignored


skydrome:
yay v9.1.0 - libalpm v11.0.2

Setting variables inside of the prepare() function dont get seen when build() is entered. Is this intentional?

Morganamilo:
Yay run's the prepare and actual build in different makepkg calls so environment variables are lost between the steps.

I would assume that regardless of yay, makepkg supports this (--repackge being especially useful) so it's a bad idea for a pkgbuild to do this so I don't consider it a bug.

skydrome:
ok thank you for explaining

PR #1008, opened 2019-08-24
Add YAY=1 indicator flag to makepkg invocation

bartoszek:
Indicate to PKGBUILD that it is executed within yay context.

refers to #1006 #1004 #893

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -105,6 +105,8 @@ func passToMakepkg(dir string, args ...string) *exec.Cmd {
 		args = append(args, "--config", config.MakepkgConf)
 	}
 
+	args = append(args, "YAY=1")
+	
 	cmd := exec.Command(config.MakepkgBin, args...)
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Please consider not using Arch Linux and not using makepkg, since you plainly hate both.

Do not ever suggest this again, and if I ever find a PKGBUILD published in the AUR which either defines or checks such a variable, or packages software which integrates with such a variable, I will immediately delete the package and indefinitely suspend the user account of the person who uploaded it.

Since `yay` is developed by individuals who don't want their accounts indefinitely suspended from the AUR, this pull request is unmergeable.

Have a nice day.

bartoszek:
_“Even the handle sank in after the blade, and his bowels discharged. @eli-schwartz did not pull the sword out, and the fat closed in over it.”_ :wink:

Issue #759, opened 2018-10-08
Split code into modules


Morganamilo:
Ideally I want yay to be built out of small, "do one thing and do it well" blocks. go-alpm, aur, go-srcinfo, go-pacmanconf are all good examples of this.

I'd like to split yay into more pieces but do not really see anything that should be extracted to a new project. Instead I think it would be a good idea to split this repo into many modules, similar to how go-pacmanconf has its own ini module.

I believe yay was like this a very long time ago, before I started contributing. Any reason you moved to one main module? And would you be apposed to splitting it out again?

Currently the two modules I want to create is one for all the coloured output. And a wrapper around aur that multi threads and respects requestsplitn.

Jguer:
I used to have yay split into `aur`, `print`, `repo` , something along those lines, but `rpc` was delegated to Mikel Oscars AUR, so it lost our `aur` lost all re-usability since it started to contain only yay specific code, that and the interactions between aur and repo made me scrap the modules. Not that the modules were bad, they were just badly divided because yay grew organically and there was no interest in having a public API for them. 

If I were to split it out again I would definitively make a config module and a devel module. 
I would be in favor of it but it has to be well thought out to avoid having to break it apart later again because modules require too much knowledge of each other

Morganamilo:
>and finally get that vendor directory out of the repo

Well the way I setup the pr right now is that it still uses a vendor directory, otherwise it would pollute home or tmp or whatever, which is not good for packaging. We can remove vendor/ from the repo and download it at build time. But then that could be done just as easily with dep.

Morganamilo:
This goes back to #579 a little. Ideally I'd like to create a local repo based helper and I don't want to duplicate code. Even ignoring that, having code separation improves readability and maintainability.

I will go and split the code into modules then (probably after #705 is merged). I'll split just the bits that I think would make good self contained modules. I'm not going to try and split the entire code base just cause.

alxpettit:
If it's split, I would very much like to have it retain a single command for the interface. Otherwise I imagine it'll end up being like apt-get/dpkg with lots of functionality scattered around and poor discoverability... so, definitely want to avoid that.

Jguer:
The module split would not affect current UX, instead it would just make it easier to navigate through the code for developers and compartmentalizing functionality to specific modules normally makes it easier to swap or make changes to components because of a stable application interface. 

A bad module architecture can be self defeating in that you end up having to hack components together or tying them too close therefor annulling all benefits.

[...]

PR #1010, opened 2019-08-27
WIP: Atomic refactoring into modules.

Jguer:
As development has stalled on most fronts and branches have gone stale, this seems like the best time to try and pass to modules (#759). Due to the interdependencies in the code an incremental approach was not as feasible. 

Care was taken not to change application/operational logic and to keep user behaviour exactly the same (even though some points will require this). The main objective is the removal of global variables for ease of testing specially.

It is grossly over packaged but should allow interface logic to be gradually split from core.

TODO:
- Fix callbacks. Current method will definitely break in case of config pointer change or release of handle.
- Fix unit tests and start adding unit tests, starting at lv3 deep packages and moving to interface.
- Add Integration tests using containers
- Document exports

- Remove smallArrow and arrow (they were left as a nag to continue refactoring)
- Coloring should be done in `text`, functions should call predefined methods such as Info, Warning, Error, List... (TODO in TODO: find morganamilo's text on required steps for localisation) 

To review:
- `main.go` was kept in the main directory for compatibility purposes. `pkg/cmd/yay` would be the natural place to put it.
- creation of internal packages for functions that handle mostly interface options.


Jguer:
yay-new

Yay-next main package separation displayed up

Yay master branch without package separation (as there is only one package)
yay-old

Morganamilo:
My intention was to do this one module at a time? Doing this all at once seem like a huge task.

Jguer:
Unglobalization of parameters didn't seem as something as doable as just atomically breaking it up into modules. Runtime for example can still have more modules extracted or collapsed into it. 

As application logic didn't change* it was mostly a refactoring job.

*The only change that broke application logic was removing config.noConfirm and HideMenus overrides from various functions and passing them on directly to the functions. Just the callback code for alpm is giving some headaches with this change as I was not as comfortable with it as I thought

Jguer:
*Although as they are now, modules such as types or completion should be easy to integrate in master. Anything that touches config or the main globals tends to break in a way or two

Jguer:
This branch is not going to be merged and remains only as a possible discussion/preview branch.

PR #1017, opened 2019-09-06
Fix typos and completion array

Ashark:
Hi,
I fixed some minor errors.

### DIFF
--- a/completions/bash
+++ b/completions/bash
@@ -61,9 +61,9 @@ _yay() {
         'c g i l p s u w y')
   upgrade=('asdeps asexplicit force needed nodeps assume-installed print recursive' 'p')
   common=('arch cachedir color config confirm dbpath debug gpgdir help hookdir logfile
-           noconfirm noprogressbar noscriptlet quiet root verbose
+           noconfirm noprogressbar noscriptlet quiet root verbose')
            #yay stuff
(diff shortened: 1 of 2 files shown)


completions/bash
```diff
@@ -61,9 +61,9 @@ _yay() {
         'c g i l p s u w y')
   upgrade=('asdeps asexplicit force needed nodeps assume-installed print recursive' 'p')
   common=('arch cachedir color config confirm dbpath debug gpgdir help hookdir logfile
-           noconfirm noprogressbar noscriptlet quiet root verbose
+           noconfirm noprogressbar noscriptlet quiet root verbose')
            #yay stuff
-           makepkg pacman tar git gpg gpgflags config requestsplitn sudoloop nosudoloop
+  common+=('makepkg pacman tar git gpg gpgflags config requestsplitn sudoloop nosudoloop
```
Jguer: The original version of this doesn't require array duplication, and it works, why change it?

Ashark: It is not "duplication", it is "addition". I made it to save comment in place. Alternatively, we can just remove comment line.
Currently, you can see clutter by doing this:
`$ yay -U -<tab><tab>`
you can see "--#yay" and "--stuff" in completion suggestions.

PR #1029, opened 2019-09-20
Changes -G in ABS to use git and symlink result.

### LINKED ISSUE #1027: -G/--getpkgbuild fails to download repo/ABS packages
#### Affected Versions
At least:
- `yay v9.2.1 - libalpm v11.0.3`
- `yay v9.2.1.r16.g1901138 - libalpm v11.0.3`
- `libarchive 3.4.0-2`

#### Issue

```bash
yay -G pacman
```

```
:: Querying AUR...
==> Failed to get pkgbuild: pacman: bsdtar: Error opening archive: Unrecognized archive format
```

It looks like it may not be querying the official repos at all, but I'm not sure. Same results for all official packages I've tried, but continues to work fine for the AUR.

```json
{
    "answerclean": "None",
    "answerdiff": "All",
    "buildDir": "/home/andy/Code/yaytmp/",
    "combinedupgrade": true,
    "makepkgbin": "/usr/bin/makepkg",
    "pacmanbin": "/usr/bin/pacman",
    "removemake": "yes",
    "tarbin": "/usr/bin/bsdtar"
}
```

Jguer:
Fixes #1027

Suppose to be temporary but maybe permanent solution will only come on `yay-next`.

### DIFF
--- a/download.go
+++ b/download.go
@@ -60,6 +60,33 @@ func gitHasDiff(path string, name string) (bool, error) {
 	return head != upstream, nil
 }
 
+// TODO: yay-next passes args through the header, use that to unify ABS and AUR
+func gitDownloadABS(url string, path string, name string) (bool, error) {
+	_, err := os.Stat(filepath.Join(path, name))
+	if os.IsNotExist(err) {
+		cmd := passToGit(path, "clone", "--no-progress", "--single-branch",
+			"-b", "packages/"+name, url, name)
+		cmd.Env = append(os.Environ(), "GIT_TERMINAL_PROMPT=0")
+		_, stderr, err := capture(cmd)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


Maryse47:
@Jguer could you set it to `buildDir` instead which defaults to same dir (`~/.cache/yay`) but it is configurable? My `~/.cache` is not persistent and I end up with broken symlinks after reboot. Using `buildDir` would solve the issue.

Jguer:
@Maryse47 You're right, I have `cacheHome` and `buildDir` set to the same but my intention was to keep it in the same place as AUR PKGBuilds

Maryse47:
@Jguer I opened PR which fixes this: https://github.com/Jguer/yay/pull/1035

PR #1035, opened 2019-09-29
Use BuildDir instead of cacheHome to store PKGBUILDS with -G

Maryse47:
`BuildDir` resolves to same place as `cacheHome` by default but it's configurable which helps in case user want to chose different directory to store PKGBUILDS.

### DIFF
--- a/download.go
+++ b/download.go
@@ -299,12 +299,12 @@ func getPkgbuildsfromABS(pkgs []string, path string) (bool, error) {
 
 	download := func(pkg string, url string) {
 		defer wg.Done()
-		if _, err := gitDownloadABS(url, cacheHome, pkg); err != nil {
+		if _, err := gitDownloadABS(url, config.BuildDir, pkg); err != nil {
 			errs.Add(fmt.Errorf("%s Failed to get pkgbuild: %s: %s", bold(red(arrow)), bold(cyan(pkg)), bold(red(err.Error()))))
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


Morganamilo:
What if I `-G foo` and `-S foo` would the paths conflict?

Jguer:
AUR has still not been ported to symlink, so as it only happens with repo packages it will never conflict. If this behaviour is ported to AUR packages though, it would for sure conflict

Morganamilo:
What about in the cause of packages moving between repos?

For example discord used to be in the AUR but is now in community. If I still have the discord files in cache and try to -G I would then cause a conflict.

Maryse47:
Why it doesn't conflict right now when both `BuildDir` and `cacheHome` default to `~/.cache/yay`?

Jguer:
> What about in the cause of packages moving between repos?
> 
> For example discord used to be in the AUR but is now in community. If I still have the discord files in cache and try to -G I would then cause a conflict.

Yes. That will be a problem, it would just symlink to a dir that does not exist in build. Fair enough, that's something that needs to be fixed in a more proper way ( the git -G patch was super rushed anyway).

I'll be merging this commit as the issue is already present by default in `cachedir=builddir`, at least like this all `PKGBUILD`s will be in the same place.

Either way, a new issue should be created to present the current issues with `-G` and a better solution present which is uniform between AUR and ABS and isn't broken by design.

PR #1041, opened 2019-10-03
Only remove VCS if the package was removed successfully

### LINKED ISSUE #1011: 'yay -R <package>' is canceled with Ctrl + c => entry is still removed from ~/.cache/yay/vcs.json
#### Affected Version
yay v9.2.1 - libalpm v11.0.3

#### Issue
Regarding -R I think it has a problem: if you cancel package removal with Ctrl + c, e.g.:
```
$ yay -R git-cola-git
[sudo] password for avangel: 
checking dependencies...

Packages (1) git-cola-git-v3.4.53.gf912364a-1

Total Removed Size:  3,15 MiB

:: Do you want to remove these packages? [Y/n] ^C
Interrupt signal received
```
then the entry for the package is deleted (too soon) from ~/.cache/yay/vcs.json.

cominixo01:
Only call removeVCSPackage if the package removal was successful. This closes #1011

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -281,7 +281,7 @@ func handleSync() error {
 }
 
 func handleRemove() error {
-	var err = show(passToPacman(cmdArgs))
+    err := show(passToPacman(cmdArgs))
 	if err == nil {
 		removeVCSPackage(cmdArgs.targets)
 	}
--- a/cmd.go
+++ b/cmd.go
@@ -281,7 +281,7 @@ func handleSync() error {
 }
 
 func handleRemove() error {
-    err := show(passToPacman(cmdArgs))
+	err := show(passToPacman(cmdArgs))
 	if err == nil {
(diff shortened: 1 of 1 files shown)


cmd.go
```diff
@@ -281,8 +281,12 @@ func handleSync() error {
 }
 
 func handleRemove() error {
-	removeVCSPackage(cmdArgs.targets)
-	return show(passToPacman(cmdArgs))
+	var err = show(passToPacman(cmdArgs))
```
Jguer: Hi, thanks for taking the time to fix this issue, Everything seems fine I would just like a small style change to `err :=` instead of explicitly declaring a var

cominixo01: Right, my bad on that one, thanks!

Jguer:
Approved, thanks for the time dispensed!

Issue #994, opened 2019-07-24
[Feature Request] Diff against last reviewed/used PKGBUILD, rather than last downloaded one


Nephyrin:
#### Affected Version

yay v9.2.1.r9.gf406866 - libalpm v11.0.3

#### Issue

When installing an AUR package, yay will offer to show you the diff of what was obtained.  If you bail out at *any point* after the download, however, it will still cache the downloaded package script.  This means that if I see something shady or (more often) incorrect in a PKGBUILD, I now have to go manually nuke it from `~/.cache/yay` to ensure it won't show up as something I've already reviewed.  Similarly, if I aborted the install for some reason without reviewing the diff, it will still appear as 'no changes'.

This makes the diff-review very non-useful as a security feature for me -- I have to assume any package that is not already installed I have never reviewed, because I have no way to discern why the previously cached files are there, or if I ever looked at them.

#### Steps to reproduce

1.  `yay -S <any aur package>`
2.  Control-C or select 'N' at any point after the download step, even before answering "Diffs to show"
3.  Repeat, and ask to see diffs - you will see 'no changes'.

#### Suggested fix/feature request

Either by default, or as an option, the diff step should should the difference between the last version I approved (by selecting "Proceed with build") and the present one, rather than against whatever was cached.

AladW:
You could do the same @rafasc did here: https://github.com/AladW/aurutils/commit/6ffd3958c3bf853904d88e1296080ed9a2ecdbd7

Jguer:
Closed with #1049

PR #1049, opened 2019-10-07
FR #994: Implement proper diff display

reyreaud-l:
As raised in #994, diff display was not proper when interrupting during
download or diff review.

This commit introduce:
- Add a ref in the git tree named "YAY_DIFF_REVIEW" which is set to the last
diff reviewed by the user.
- Change the function displaying diff to user to make proper use of this
new ref. If the ref does not exist in the git tree, it will display the
diff from origin to HEAD@{upstream}.
- Add a function updating the YAY_DIFF_REVIEW after it has been reviewed
by user. It uses the `toDiff` list in the `install` function.
- Add utils function related to handling this change in download.go.

This change only consider diff reviewed to update the YAY_DIFF_REVIEW ref.
If a user does not enable review, the ref will not be updated and the
user will be prompted for review when he enables diff review.

It is my first time submitting change to an open source project. If anything is wrong, feel free to correct me! :)

Signed-off-by: Loic Reyreaud <reyreaud.loic@gmail.com>

### DIFF
--- a/download.go
+++ b/download.go
@@ -48,17 +48,55 @@ func downloadFile(path string, url string) (err error) {
 	return err
 }
 
-func gitHasDiff(path string, name string) (bool, error) {
-	stdout, stderr, err := capture(passToGit(filepath.Join(path, name), "rev-parse", "HEAD", "HEAD@{upstream}"))
+// Update the YAY_DIFF_REVIEW ref to HEAD. We use this ref to determine which diff were
+// reviewed by the user
+func gitUpdateSeenRef(path string, name string) error {
+	_, stderr, err := capture(passToGit(filepath.Join(path, name), "update-ref", "YAY_DIFF_REVIEW", "HEAD"))
 	if err != nil {
-		return false, fmt.Errorf("%s%s", stderr, err)
+		return fmt.Errorf("%s%s", stderr, err)
 	}
+	return nil
+}
 
-	lines := strings.Split(stdout, "\n")
-	head := lines[0]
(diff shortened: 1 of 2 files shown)


Jguer:
Hey, thanks for the time put into this, I'll review it as soon as possible.

reyreaud-l:
Made a couple of force push to rebase the branch and resolve merge conflict, as well as GPG sign the commit.

Jguer:
Tested on package `yay-git`. Manually inserting ref showed it works as desired, clean install works.

Seems to be working as intended.

Jguer:
Seems to be working correctly still, let's merge it.
If you would like to leave some suggestions on how to unit test it or functionality test it I'm all ears.

Either way, thank you for your contribution, I hope to see more contributions from you in the future

Morganamilo:
I've missed my chance now but I'd much prefer the ref name be something generic like `AUR_DIFF` so that if other AUR helpers decide to use this method, cache dirs can be shared and this would apply to both helpers.

Jguer:
That's a cool idea actually. https://github.com/AladW/aurutils/commit/6ffd3958c3bf853904d88e1296080ed9a2ecdbd7 uses `AURUTILS_SEEN` it seems. Until the new version of yay is packaged we can change this with no problem.

reyreaud-l:
I agree, I will check other AUR helpers see if we can find a common pattern for naming these refs. If not I will use AURUTILS_SEEN. I'll open another PR soon.

PR #1053, opened 2019-10-10
Re-add --batchinstall and fix dep issues

Morganamilo:


Jguer:
Can you just point me out what was the issue with the deps?

Morganamilo:
DepOrder is built from DepPool. And as it builds it drains the dep pool. Meaning that `dp.repo` will be empty. It was literally just a typo of `dp` instead of `do` when looping over repo packages.

Jguer:
Fair enough, happens. 

What would you say are the biggest challenges in extracting it to its own package so it doesn't depend on yay runtime information? 
I would like to add some unit tests of the type: `We run a mock AUR server. We feed X package names, we expect to get these X installed as exp and Y installed as deps` to prevent this type of regression

Jguer:
Also missing is manpages and completion data for this

Morganamilo:
> Fair enough, happens.
> 
> What would you say are the biggest challenges in extracting it to its own package so it doesn't depend on yay runtime information?
> I would like to add some unit tests of the type: `We run a mock AUR server. We feed X package names, we expect to get these X installed as exp and Y installed as deps` to prevent this type of regression

I wouldn't really say there's any challenges, just a lot to write.

Jguer:
I know what you mean, just setting up all of the CI/"CD" pipeline has been hammering away at everything.

I would leave this PR unmerged until pacman 5.2 has passed (which I hope is soon). Meanwhile I've set go-alpm to check and build when pacman enters testing and also to compile against pacman-git once a week and on branches named `pacman*`. I'll set up the same but maybe a bit more lax on `yay`.

Jguer:
What did I say an hour ago :rofl: ?
If I make a release this week I'll make it without #1053

Morganamilo:
My message was the last time I checked the thread, whoops.

Morganamilo:
I just don't want to deal with the extra conflicts really.

Jguer:
Fine by me, no issue. I was planning on testing automatic releasing on master branch on these last commits but I'll have to implement release-branch releasing before, just moves the cycle a bit back.

PR #1054, opened 2019-10-10
Have "make build" build be the default again

Morganamilo:
It's been a while since I built yay manually so I'm not sure when this change was made.

But given how the default rule is still there I assume this change was accidental. The first rule is the default, the actual name `default` has no special meaning.


Jguer:
You're right, my bad when I was cleaning it up

PR #1056, opened 2019-10-10
Add --searchby

### LINKED ISSUE #936: [FR] Search in package names only

Morganamilo:
Allow using the RPC's &by= field during -Ss

Fixes #936.

### DIFF
--- a/depPool.go
+++ b/depPool.go
@@ -188,7 +188,7 @@ func (dp *depPool) findProvides(pkgs types.StringSet) error {
 		words := strings.Split(pkg, "-")
 
 		for i := range words {
-			results, err = rpc.SearchByNameDesc(strings.Join(words[:i+1], "-"))
+			results, err = rpc.Search(strings.Join(words[:i+1], "-"))
 			if err == nil {
 				break
 			}
--- a/config.go
+++ b/config.go
@@ -56,6 +56,7 @@ type Configuration struct {
 	GpgFlags           string `json:"gpgflags"`
 	MFlags             string `json:"mflags"`
 	SortBy             string `json:"sortby"`
+	SearchBy           string `json:"searchby"`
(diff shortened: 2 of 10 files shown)


Jguer:
Some mention that modules were updated would be nice in the log. Possibly separating `go-alpm`'s update from the rest. Apart from that looks fine

depPool.go
```diff
@@ -188,7 +188,7 @@ func (dp *depPool) findProvides(pkgs types.StringSet) error {
 		words := strings.Split(pkg, "-")
 
 		for i := range words {
-			results, err = rpc.SearchByNameDesc(strings.Join(words[:i+1], "-"))
+			results, err = rpc.Search(strings.Join(words[:i+1], "-"))
```
Jguer: In depPool wouldn't it be better to search only Name instead of NameDesc?

Morganamilo: I see no reason to. More results means a better chance of finding a provide. And seems as we can fit everything into a single into query there should be no speed difference.

Jguer: Ok, I can agree with that.

Jguer:
Also missing completions and man page update. I'll probably create an issue on it so it's not forgotten

boussou:
I can't believe it is impossible to limit search only to package name in AUR?
it tried using keyword "git", to many results !!

PR #1062, opened 2019-10-14
Change ref name used in git to compute diff to "AURUTILS_SEEN".

reyreaud-l:
As discussed in #1049, change the ref name we use in git to compute whether or not to display diff to AURUTILS_SEEN.
Also introduce a global constant in download.go to avoid re-typing the string every time.

This is made in an effort to use the same name as aurutils, so that if one user decide to share the cache directory between different aur helpers, the tools are used the same ref to display diffs.

### DIFF
--- a/download.go
+++ b/download.go
@@ -14,9 +14,7 @@ import (
 	"github.com/Jguer/yay/v9/pkg/types"
 )
 
-const (
-	gitDiffRefName = "AURUTILS_SEEN"
-)
+const gitDiffRefName = "AUR_SEEN"
 
 // Decide what download method to use:
 // Use the config option when the destination does not already exits
--- a/download.go
+++ b/download.go
@@ -14,6 +14,10 @@ import (
 	"github.com/Jguer/yay/v9/pkg/types"
 )
 
+const (
+	gitDiffRefName = "AURUTILS_SEEN"
+)
+
 // Decide what download method to use:
 // Use the config option when the destination does not already exits
 // If .git exists in the destination use git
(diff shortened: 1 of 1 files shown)


Morganamilo:
I'd prefer just `AUR_SEEN`. Ideally I could convince @AladW to change it ;)

reyreaud-l:
That would make more sense. Maybe merge it with AURUTILS_SEEN and wait until we agree with aurutils to change it ? Then we all change it at once ?

Otherwise I can change it to AUR_SEEN and update the PR.

AladW:
If more than 1 project uses `AUR_SEEN`, I don't mind changing it in aurutils as well.

Jguer:
`AUR_SEEN` or `AUR_REVIEWED` are equally good for me.

@AladW  Off-topic but it's already usual for yay topics: I understand that the default directory for repositories in `aurutils` is `~/.cache/aurutils/sync/` depending as well on`$XDG_CACHE_HOME` and `$HOME`.
If we standardized on a cache suffix (`${cache}/aur/sync`? `${cache}/aur`?) for AUR repos it could also benefit this change in diff ref names.
(Any other ideas on standardization of protocol or directories I'm interested)

AladW:
For cache directories, aurutils has used `AURDEST` since the beginning, because `pacaur` was using the same variable. So you could define XDG_CACHE_HOME and HOME, or just set `AURDEST=$cache/aur/`, for example.

Morganamilo:
Maybe we should respect AURDEST. not sure how I feel about using the same dir by default if that's what @Jguer is suggesting.

@E5ten

E5ten:
I'll implement whatever name gets decided here for increased compatibility, AUR_SEEN seems like a fine choice.

rafasc:
@Morganamilo Doesn't sound like a good idea at first. For example, `aurutils` will `reset --hard` repos in its operation and allows keeping local commits by pulling with `--rebase`. Having repos shared between tools kind of restricts a bit what you can do with the repos without stepping on each others toes.

Jguer:
> Maybe we should respect AURDEST. not sure how I feel about using the same dir by default if that's what @Jguer is suggesting.
> 
> @E5ten

I think we should respect `AURDEST`, possibly ignoring config file settings if `AURDEST` is set even, could be useful for scripting and whatnot.
 
Merging default paths, could be nice but I see the possible pains of forced migration to users. @Morganamilo thoughts on the cons of this?

download.go
```diff
@@ -14,6 +14,10 @@ import (
 	"github.com/Jguer/yay/v9/pkg/types"
 )
 
+const (
+	gitDiffRefName = "AURUTILS_SEEN"
```
Jguer: Just change this to `AUR_SEEN` and we're good

Morganamilo: And ditch the unneeded brackets.

PR #1064, opened 2019-10-16
Refactor files

Morganamilo:
This moves the stuff under pkg/types into separate modules. You once said `utils` is a bad module name, for the same reasons I think `types` is too.


Jguer:
You are right on that `types` is a bad module name, the only way I had of justifying it was that it was for types that didn't have AUR/package management context and I wanted to avoid `stringset.Stringset` but seeing `stringset.FromSlice` I like it

Morganamilo:
It's making me miss rust's import system.

But yeah I hate it too. We could do `import . stringset`. But then we would have to undo b60a888, otherwise that function would be in the global scope.

Morganamilo:
By the way, seems as I don't understand the go module system that well. Why is v9 in the import path? Surely what version we use should be managed by `go.mod ` or maybe it can just use the one in the repo as is. That may be annoying as the version get's bumped.

Jguer:
Better to have a `stringset.Stringset` than putting it in global scope I guess, `types` was not the best option and I think we can merge this one.

Jguer:
> By the way, seems as I don't understand the go module system that well. Why is v9 in the import path? Surely what version we use should be managed by `go.mod ` or maybe it can just use the one in the repo as is. That may be annoying as the version get's bumped.

Also another pain point but if the module is called `module github.com/Jguer/yay/v9` then it must all have that prefix. I'm researching now though to check if it's really idiomatic

Morganamilo:
By the way, I feel like I've totally broken #1010 now. But I would prefer if that got split into smaller PRs anyway.

Jguer:
That's actually my bad because I should have written there that I have given up on that branch. Was good to get an idea of the "finished" product, lessons learned but there are already things I would have done differently. One of them is doing it gradually and the other one is starting at the easiest and work the way up and not the reverse

Jguer:
My next objective would be the text package (for input and output of text) and passing `Base` into its own package (opens the way to moving `keys` to its own package)

Morganamilo:
Base should be part of the depsolver package really. Assuming I follow the model here: https://docs.rs/aur-depends/0.2.5/aur_depends/

Jguer:
And `keys`/`pgp` uses the `dep` package's Base for its operations? 

`dep.base.Base`  or `dep.Base`?

Morganamilo:
The fetch lib should all come under one module.

Jguer:
If you want to do the piece by piece module refactoring you can :
1. move base to dep/base
2. move keys functions to pgp
3. move depOrder and dep Pool to dep/<?>
4. move pgp into fetch/pgp

Keeping the smaller modules moving inwards to make it easier to extricate the more function/logic heavy modules

Morganamilo:
I consider the dep solving to be its own lib and should probably be totally separate from yay. Just like how `go-srcinfo` and `aur` are.

Jguer:
I agree, dep solving should be uncoupled as much possible, that was not my point. My only question was which could small steps are the closest ones we can do.

Morganamilo:
Well base should be part of the depsolver so it can't be moved into a different module (or rather it could, but then would have to be moved again later)

depOrder, depPool and depCheck will also be moved into the lib.

Jguer:
(or rather it could, but then would have to be moved again later) -> good enough for me, I'll move it to pkg/dep/base , extricate keys (and maybe vcs if already possible) and then when dep lib is done it's a matter of import paths.

Then keys and vcs can be moved again, it's just a question of untangling components one by one.

PR #1065, opened 2019-10-17
wip: move aur downloading to go-aurfetch

Morganamilo:
Adds an aur-sync like view to replace the diff+edit menu
Drops support for tar

### DIFF
--- a/clean.go
+++ b/clean.go
@@ -192,10 +192,8 @@ func cleanUntracked() error {
 		}
 
 		dir := filepath.Join(config.BuildDir, file.Name())
-		if shouldUseGit(dir) {
-			if err := show(passToGit(dir, "clean", "-fx")); err != nil {
-				return err
-			}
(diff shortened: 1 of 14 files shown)


Morganamilo:
I'd really like to know what people think of the new way of viewing diffs and files.

It's clearly more powerful but it may be so different people hate it. Especially the people who don't know vim. The finished PR will allow you to configure what program to use anyway. Seems as yay just calls `vifm <dir>` you could substitute that for any file manager or `ls` or something.

I did nuke the editmenu code to implement this just because It was easier to delete it. But I may be fine with supporting both. We need the menu code for `yay <target>` anyway so it's not that extra complex.

However I do wish to remove cleanmenu  (i'll make an issue on why). So if that is gone, then it would fit to have edit+diff menu gone and replace it with simpler code.

For example have it just print all diffs always. Then if the users want to do anything more complex, tell them to enable the new "view" feature.

PR #1070, opened 2019-10-19
more selfexplaining upgrade process

nicoduesing:
At least for me it was not very unintuitive what I have to do here. Maybe because I overread "not" and also because I simply likes to upgrade all packages. That was not explained ether, so I didn't consider to just press enter. This hopefully makes it much more understandable.

### DIFF
--- a/upgrade.go
+++ b/upgrade.go
@@ -356,7 +356,7 @@ func upgradePkgs(aurUp, repoUp upSlice) (stringset.StringSet, stringset.StringSe
 	fmt.Printf("%s"+bold(" %d ")+"%s\n", bold(cyan("::")), allUpLen, bold("Packages to upgrade."))
 	allUp.print()
 
-	fmt.Println(bold(green(arrow + " Packages to exclude: (eg: "1 2 3", "1-3", "^4" or repo name), press enter to upgrade all listed packages")))
+	fmt.Println(bold(green(arrow + " Packages to exclude: (eg: \"1 2 3\", \"1-3\", \"^4\" or repo name), press enter to upgrade all listed packages")))
 	fmt.Print(bold(green(arrow + " ")))
 
 	numbers, err := getInput(config.AnswerUpgrade)
--- a/upgrade.go
+++ b/upgrade.go
@@ -356,7 +356,7 @@ func upgradePkgs(aurUp, repoUp upSlice) (stringset.StringSet, stringset.StringSe
 	fmt.Printf("%s"+bold(" %d ")+"%s\n", bold(cyan("::")), allUpLen, bold("Packages to upgrade."))
 	allUp.print()
 
-	fmt.Println(bold(green(arrow + " Packages to not upgrade: (eg: 1 2 3, 1-3, ^4 or repo name)")))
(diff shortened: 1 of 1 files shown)


Morganamilo:
Exclude may better wording. I don't like the extra text on the end though. It's already long enough as is.

nicoduesing:
@Morganamilo Thanks for the feedback. That's a downside, yes. What solution would you prefer?
Maybe "Packages to exclude: (eg: \"1 2 3\", \"1-3\", \"^4\" or repo name) skip with enter"?
Or "Packages to exclude: (eg: \"1-3\" or name) enter to upgrade"?
Or "Type eg: \"1 2 3\", \"1-3\", \"^4\" or name to exclude or press enter to upgrade"?
Or "Exclude packages (eg: \"1 2 3\", \"1-3\", \"^4\" or repo name) or upgrade with enter"?
Or "Exclude packages with their number. Just enter starts upgrade"?
Maybe something else?

Morganamilo:
Honestly I don't really like all of them. I feel like the user should be able to figure stuff out just from reading it. Especially if using "exclude" makes it more clear.

There's also the option of disabling the menu by default. You'd then assume people enabling it would know what they are doing.

nicoduesing:
There is a lot that can be improved in many of the menus, I think. But I didn't expected to rebuild the complete user interaction. I just liked to improve it, where I was the most confused and liked to make a small change, so heavy users still have the expected, learned behaviour but it is more understandable for new users like me.

"I feel like the user should be able to figure stuff out just from reading it." Yes, that is the reason I changed it, to let the user understand it easily from reading it.
"Especially if using 'exclude' makes it more clear." That's the reason why I wrote "exclude", because it is more clear and because it can not be accidentally skipped while reading like the word "not".

As I have made a couple of suggestions and you like none of them, can you please explain what is the problem with them and/or make a better solution?

Morganamilo:
>I feel like the user should be able to figure stuff out just from reading it

By this I meant that I believe users should already be able to figure it out. And given how many users yay has it seems most do. I did agree "exclude" is better wording and I agreed with that change, that's why I said especially.

My problem with the others are either the length or how they no longer document the ranges and negation syntax.

nicoduesing:
Ah, "^4" is for negation? I've wondered what it could stand for. XD

Morganamilo:
Funny thing it's actually a pacman feature. ^4 works for negation in the group selection.

nicoduesing:
I've removed the part with "enter". Maybe exclude is enough. I think I would still be confused how I choose to "exclude nothing" so to upgrade all, but at least I know now and it is a small improvement. Maybe it gets improved further by someone else.

Morganamilo:
Type nothing to exclude nothing. Seems intuitive to me. And aligns with my view of minimal key presses.

I'm still considering disabling the menu by default but that depends on #1065 and how the ui changes.

These changes look good. But to keep the history clean there's no point in merging a commit then a fix up when you can just fix the original.

You can do that with:

`git rebase -i master` and change `pick` to `fixup` on the last commit.

nicoduesing:
So far I just have changed it with the github ui. As fixup changes my local git history and the changes are already on the github server, I would have to remove the branch there, create a new branch, fix the history and push it as new branch, right? 
I simply copied the text manually and created a new pull request: #1071

Morganamilo:
Ah, I forget you can make commits via the website. I guess that works instead of rebasing.

PR #1072, opened 2019-10-19
Move yay code into cmd/

Morganamilo:
This tidies up the repo quite a bit


Morganamilo:
No idea why this all of a sudden fails. The test fails for me locally in master, despite passing on github's CI.

For reference gpg says `gpg: keyserver receive failed: No keyserver available`

Jguer:
Does doing this break `go get github.com/Jguer/yay` moving it to `go get github.com/Jguer/yay/cmd`? I would prefer to keep this unchanged as the package itself is a cli application

Morganamilo:
Do we even support go get? People should really be getting this from the AUR. `cmd` is a go standard at least https://github.com/golang-standards/project-layout#cmd. I just hate the mess of files cluttering up the rood directory.

Jguer:
Would still break everything else related to the guide lines:

`The directory name for each application should match the name of the executable you want to have (e.g., /cmd/myapp).`

`It's common to have a small main function that imports and invokes the code from the /internal and /pkg directories and nothing else.`

I don't like the mess of files cluttering the root as well, but this breaks the small incremental changes. Better to have a mess as a motivation to do the uncoupling correctly instead of hiding the real issue into another directory.

Once everything is uncoupled we end up with a lonely main.go file in the root or a main.go file in `cmd/yay` if we decide to completely obsolete the current `go get`

Morganamilo:
Right, I'll just have to endure it for now :(

Jguer:
I'm delivering my thesis this week, so it's going to be a busy week, but after that I should be able to help accelerate the process.

PR #1073, opened 2019-10-20
Test failing CI

Morganamilo:


Morganamilo:
This passes here but fails when I run `make test` locally.

Morganamilo:
Honestly I'm tempted to delete the entire test file. It's feels like it's testing gpg itself rather than yay.

Jguer:
How does it fail?

Morganamilo:
It fails here https://github.com/Jguer/yay/pull/1072#issuecomment-544207058. I thought maybe it was something to do with a key being expired or a server being down so this was to test a PR that changed basically nothing.

Jguer:
https://github.com/Morganamilo/yay/blob/d699a85c1281f6bef3f127c56c50b9f257ac0d55/cmd/keys_test.go#L46

Morganamilo:
The error I get though is "no keyserver". Plus even though the source has moved. The tests are still ran from the repo's root. Unless `go test` `cd`'s into `cmd/`/

Jguer:
Which would make sense because the keyserver declared in `keys_test.go` doesn't return an answer unless it has its testdata (ran into this in `modules` branch`).

Morganamilo:
Right so go test does indeed `cd` into `cmd/`? Cargo does the opposite at least... which made that unexpected.

Jguer:
Apparently yes
https://brandur.org/fragments/testing-go-project-root

PR #1074, opened 2019-10-21
Remove delta support for pacman 5.2.0

rumpelsepp:
One `go get -u` is missing, but this depends on: https://github.com/Jguer/go-alpm/pull/37


Jguer:
This has already been prepared in `pacman-next`, thank you for the time spent on making sure `yay` stays working :+1:

PR #1078, opened 2019-10-21
Bump version for release

Jguer:


Morganamilo:
I think you forgot the makefile.

Jguer:
Very true, been using `make VERSION=x` so I kind of forgot that small issue. I can tag and compile 9.4.0 with the pacman-next merged and against pacman 5.2 so it's already done and release is one key push but maybe better to wait until it's in testing at least so `yay-git` is not broken for an unforseen time

doc/yay.8
```diff
@@ -1,5 +1,5 @@
 '\ t
-.TH "YAY" "8" "2018\-07\-01" "Yay v8.972+" "Yay Manual"
+.TH "YAY" "8" "2019\-10\-21" "Yay v9+" "Yay Manual"
```
Morganamilo: This should be the exact version number. Also I don't think this justifies three separate commits.

Jguer: I was going to squash it on merge, but I can squash it in branch.
I'll put `yay v9.4`

PR #1079, opened 2019-10-21
Update documentation

Morganamilo:


doc/yay.8
```diff
@@ -532,6 +552,16 @@ Shows statistics for installed packages and system health.
 pacman -Qmq | grep -Ee '-(cvs|svn|git|hg|bzr|darcs)$' | yay -S --needed -
 pacaur-like devel check.
 
+.SH ENVIRONMENT VARIABLES
+.TP
+.B AURDEST
+Can be set to configure the builddir.
```
Jguer:
```suggestion
Can be set to configure the build directory
Overrides --builddir.
```
Just a mention that it overrides the builddir would be nice

doc/yay.8
```diff
@@ -532,6 +552,16 @@ Shows statistics for installed packages and system health.
...
+
+.TP
+.B VISUAL EDITOR
```
Jguer:
```suggestion
.B VISUAL 
.B EDITOR
```
Disregard if it doesn't work well visually

doc/yay.8
Morganamilo: It doesn't override `--builddir`. The priority is `flags > environment > config`. I said `builddir` to match the flag name. But maybe I'll take this suggestion and change it to `Overridden by --builddir`.

doc/yay.8
Morganamilo: It does not work.

image


Maybe I should have but a comma though thinking about it.

Jguer: or a `/` , because I know there will be some who will try and set a `VISUAL EDITOR` env flag.
It will work though `EXPORT VISUAL EDITOR=vim` :sweat_smile:

doc/yay.8
Jguer: Fair enough, my confusion, just something in terms of priorities so it can be pointed to in order to shame users when they complain that their setting in `.config` didn't work :innocent:

PR #1082, opened 2019-10-23
Clarfiy Out Of Date Message

naclander:
The existing message was a bit confusing and made me think that the AUR package 
had an update pending which Yay refused to do. I realized instead that the package is
actually *flagged* out of date on the AUR.

I then realized there is a section in the README specifically about this confusion.

So, by prefixing the word "Flagged" I think the slight confusion can be eliminated, and the 
README section possibly removed.

### DIFF
--- a/print.go
+++ b/print.go
@@ -39,7 +39,7 @@ func (warnings *aurWarnings) print() {
 	}
 
 	if len(warnings.OutOfDate) > 0 {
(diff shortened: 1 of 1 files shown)


Jguer:
Thanks for your contribution.
I'll merge it in for this release and we'll see if it confuses people less.

PR #1099, opened 2019-10-25
completions/fish: inherit completions from pacman

akiirui:
Inherit completions from [pacman](https://github.com/fish-shell/fish-shell/blob/master/share/completions/pacman.fish), can made yay completions file easier to read and edit.
Do not need update this completions when only pacman completions changed.

This completions file is refine from `yay 9.4.2` man pages.
But `-P --fish` output `invalid option 'fish'`, So I mark it as comment.


akiirui:
Inherit completions have many weird issues. close this pull request.

Issue #492, opened 2018-06-13
Feature request: show AUR comments before install


palb91:
#### Affected Version
```console
$ yay -V
yay v6.786.r90.g1002a4f
```

#### Issue

I compare some AUR helpers since few days – since I don't know what is happening with `yaourt` – and I choose `yay` for some intersting features (reverse ordering when searching). One feature is missing that I appreciated in yaourt, the possibility to see **AUR comments**.

Some packages are broken, and sometimes a quick fix is provided in comments, that's really help…

I imagine a config parameter such as:
```json
{ "aurcomments": true }
```

What do you think about that ?

palb91

Jguer:
Seems somewhat out of scope for `yay`, news is already on the borderline but if a quickfix is available in the AUR wouldn't the first thing to do be going there and check, it's not that much of a time saver

Jguer:
Depends, are the comments parseable or by XML? Because I think I saw it once implemented by page parsing and that's something I would like to avoid

Morganamilo:
Nope pretty sure you have to parse the comments of the AUR page. Not that I've really looked into it.

Jguer:
Zawinski's law at work I guess.

I'll try and look into it but although go offers great facilities in web parsing, I would really, really like to avoid turning `yay` into a web browser.

palb91:
Hi,
> That's why I initially suggested using -W/--web for news.

Yep, I spend few time in open/close issues searching for "comments" and similar word… I haven't found this one :/ Sorry for duplicate…
At least #384 provides a workaround for this.

> Zawinski's law at work I guess.

Maybe, not sure where are the limits… I asked for that because yay is « based on the design of yaourt, apacman and pacaur » according to you presentation, and AUR comments was one of the reasons I used yaourt and never changed.

I understand your reluctance for what could imply this functionnality but eh… it doesn't hurt to ask :)

sandwm:
I don't know the AurJson RPC interface, so I can't tell if that would do, but if web parsing would be necessary, pakku seems to implement it [here in downloadAurComments](https://github.com/kitsunyan/pakku/blob/master/src/aur.nim#L147). It's not too complicated, though I understand the reluctance to crawl the web instead of a well defined RPC interface.

When installing new AUR packages I usually go to their web page anyway, but when updating I like to see the (new) comments about the current version, so this feature would be very much appreciated. In case you decide not to implement it, I would like to know if pull requests for this feature would be accepted (although I can't promise if / when I would find the time).

[...]

PR #1101, opened 2019-10-28
Add SudoBin and SudoFlags

### LINKED ISSUE #1031: [Feature]- Can we get a terminal bell before sudo prompt on long builds.
#### Affected Version
yay v9.3.1 - libalpm v11.0.3
#### Steps to reproduce

1. build large package such as linux-ck
2. Miss sudo prompt 
3. have to redo the intial build process
FEATURE Request

Cam we get a terminal bell before the sudo prompt shows up (especially after a long build) to prompt us that our password has to be entered.

Morganamilo:
Same as #492 but with https://github.com/Jguer/yay/pull/942#discussion_r294338344.

Fixes #1031

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -111,6 +111,8 @@ Permanent configuration options:
     --batchinstall        Build multiple AUR packages then install them together
     --nobatchinstall      Build and install each AUR package one by one
 
+    --sudo                <file>  sudo command to use
+    --sudoflags           <flags> Pass arguments to sudo
     --sudoloop            Loop sudo calls in the background to avoid timeout
     --nosudoloop          Do not loop sudo calls in the background
 
--- a/config.go
+++ b/config.go
@@ -61,6 +61,8 @@ type Configuration struct {
 	SearchBy           string `json:"searchby"`
 	GitFlags           string `json:"gitflags"`
 	RemoveMake         string `json:"removemake"`
+	SudoBin            string `json:"sudobin"`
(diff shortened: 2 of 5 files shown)


Jguer:
I would have added a warning pre sudo call detailing the location of `sudobin` if it's different from `/usr/bin/sudo`

Morganamilo:
Why? We don't do it with any other commands.

Jguer:
Because changing yay's sudo bin and flags can potentially allow the execution of other commands as root. 
If we can't prevent it, at least a small warning would be nice.

Morganamilo:
I guess. But a malicious program could instead change pacmanbin to a script that prompts you for your password or makes a call to sudo. It could also place a sudo binary somewhere in home and add it to path. Hell it could probably redirect all your input to a file or log your keys.

So I don't really think sudo specifically makes it any less secure.

Morganamilo:
Surely you would have to impede the user though by adding a prompt. Otherwise you're adding a warning but still calling the potentially malicious script right after.

Jguer:
In case of sudo commands (without `NOPASSWD`) or malicious sudo asking for a password, it will be prompting right below for said password. 

In this scenario you're assuming user level is already compromised so executing a malicious script just in the user directories would be out of scope already.

Morganamilo:
Isn't the entire point of the warning for the case there's a malicious script? Anyway my point is the entire thing is out of scope.

PR #1106, opened 2019-10-30
Remove -F in fish completion

Morganamilo:
This flag is only in fish-git, making it broken for most people.


KokaKiwi:
Thanks :heart:

PR #1102, opened 2019-10-28
Update completions for zsh and bash to pacman 5.2

Jguer:
Also adds missing options for `yay 9.4`

PR #1111, opened 2019-11-04
Remove tar support

Morganamilo:
tar is slower and harder to diff. I doubt any one still uses it intentially.

There's a chance some people have not cleared their cache in two years
and still have tar based packages around. But forcing them to clear
their cache is not a big deal at all.

### DIFF
--- a/config.go
+++ b/config.go
@@ -45,7 +45,6 @@ type Configuration struct {
 	MakepkgConf        string `json:"makepkgconf"`
 	PacmanBin          string `json:"pacmanbin"`
 	PacmanConf         string `json:"pacmanconf"`
-	TarBin             string `json:"tarbin"`
 	ReDownload         string `json:"redownload"`
 	ReBuild            string `json:"rebuild"`
 	BatchInstall       bool   `json:"batchinstall"`
@@ -72,7 +71,6 @@ type Configuration struct {
 	NoConfirm          bool   `json:"-"`
 	Devel              bool   `json:"devel"`
 	CleanAfter         bool   `json:"cleanAfter"`
-	GitClone           bool   `json:"gitclone"`
 	Provides           bool   `json:"provides"`
 	PGPFetch           bool   `json:"pgpfetch"`
 	UpgradeMenu        bool   `json:"upgrademenu"`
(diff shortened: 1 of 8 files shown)


Morganamilo:
I ended up removing `-f` but looks like we still need it for ABS packages. Even though we use git now we still delete the dir and copy it over.

Morganamilo:
Also reading through this, do we really need to shell out to `cp`?

Jguer:
> Also reading through the ABS code, do we really need to shell out to `cp`?

A inheritance from the old inline tar.gz code, can be removed.
https://stackoverflow.com/questions/51779243/copy-a-folder-in-go

Jguer:
#1102 now implements the changes to bash and zsh completion for this

parser.go
```diff
@@ -528,10 +526,6 @@ func handleConfig(option, value string) bool {
 		config.AnswerUpgrade = value
 	case "noanswerupgrade":
 		config.AnswerUpgrade = ""
-	case "gitclone":
-		config.GitClone = true
-	case "nogitclone":
```
Jguer: Also missing removal of case "tar" in both attribution and valid param

config.go
```diff
@@ -72,7 +72,6 @@ type Configuration struct {
 	NoConfirm          bool   `json:"-"`
 	Devel              bool   `json:"devel"`
 	CleanAfter         bool   `json:"cleanAfter"`
-	GitClone           bool   `json:"gitclone"`
```
Jguer: Missing removal of `config.Tarbin`

Morganamilo:
Annoyingly when searching for `tar` it also pulls in `nextArg` and `target`. So I couldn't really just do a find all. I was kind of relying on the CI to catch the unused for me.

bundi78:
Hi,

sorry to be late, but for me, that feature was useful. I have to work behind an awkward proxy and --nogitclone was the only working solution to use AUR.

With 9.4.3, I have no (easy) access to AUR.

mjruschak:
Oh boy... Any way you could add this back.  Found this option useful for behind proxys.  Much appreciated if we could get it back.

PR #1122, opened 2019-11-21
Add low priority option [v3]

tinti:
Low priority will make yay and all child process use the lowest possible
CPU and IO priority (nice +19, ionice class idle).

Useful for background build updates.

Signed-off-by: Vinicius Tinti <viniciustinti@gmail.com>

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -116,6 +116,9 @@ Permanent configuration options:
     --timeupdate          Check packages' AUR page for changes during sysupgrade
     --notimeupdate        Do not check packages' AUR page for changes
 
+    --lowpriority         Use lowest CPU and IO priorities
+    --nolowpriority       Do not use lowest CPU and IO priorities
+
 show specific options:
     -c --complete         Used for completions
     -d --defaultconfig    Print default yay configuration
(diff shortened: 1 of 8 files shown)


Jguer:
Before getting into the code, I have some questions:
- If I'm someone that wants to touch on priorities wouldn't I be better served with using ananicy?
- If I have selinux installed, won't this mean creating a policy for it?
- is the alternative running `yay` as normal vs `nice -n  19 ionice -c 3 yay`?

tinti:
> Before getting into the code, I have some questions:
> 
>     * If I'm someone that wants to touch on priorities wouldn't I be better served with using ananicy?

I have never heard about it. Nice.
My argument to use this patch instead of Ananicy would be that Ananicy is too much for a simple task.
 
>     * If I have selinux installed, won't this mean creating a policy for it?

Yes. I believe so. Either that or not use this option at all. Maybe would be better to make it clear in the documentation.

>     * is the alternative running `yay` as normal vs `nice -n  19 ionice -c 3 yay`?

Yes. I have been using `nice -n 19 ionice -c 3 yay` for a long time. It works really well for me specially when I update packages like [https://github.com/root-project/cling](https://github.com/root-project/cling) while working in my workstation.

I was thinking in creating an *alias* but making it builtin sounds safer. Specially because I can do `yay save` and guarantee that this option will be used always.

Morganamilo:
I'm also very against importing syscall.

tinti:
How about **V4** here? https://github.com/tinti/yay/tree/feature/add_low_priority_option_v4

This uses `exec.Command` instead of `syscall`.

Jguer:
On the code side, I don't like `syscall`, but I like `exec.Command` even less. 

I just don't like this integration at all when using the proper unix tools for it gives you more choice and more freedom. If we wanted to replicate that we'd end up reimplementing nice and ionice inside of yay and that includes adding 7 command line options for something that is essentially around 20 characters of text before `yay` or a command statically put in a background update script.

The implementation would be good but the feature itself drifts from the simplicity and the integration with other tooling yay aims for. Don't take this a discouragement, I hope to see more pull requests from you but this one is a pass for me.

PR #1147, opened 2020-01-10
Add error handler on saveConfig

msAlcantara:
I think that should be good to add error handler on `saveConfig()` function


Jguer:
You are indeed right, thanks for your contribution

Issue #1137, opened 2019-12-31
unable to skip dependency with --assume-installed / --nodeps


irgendwr:
#### Affected Version

yay v9.4.2 - libalpm v12.0.1

#### Issue

I am unable to skip a given dependency using yay with the `--assume-installed` flag.
In #722 it was mentioned that yay doesn't understand the flag. It looks like that's still the case.

> yay does not actually understand this flag.

Similarly the `--nodeps` flag is also not understood by yay.

If there is another way to skip dependencies then please let me know, thanks.

#### Steps to reproduce

1. `yay --assume-installed pyqt5-common multibootusb`
2. Choose `1`

Note: This example no longer produces the same output because the `pyqt5-common` dependency was dropped. I don't have another example at the moment but the issue still exists of course.

#### Output

```
~ >>> yay --assume-installed pyqt5-common multibootusb
2 aur/multibootusb-git 20181002.dec46ef-1 (+13 2.05%) (Out-of-date 2019-12-23) 
    Boot multiple live Linux distros from a usb flash drive.
1 aur/multibootusb 9.2.0-3 (+104 1.04%) (Out-of-date 2019-12-16) 
    Boot multiple live Linux distros from a usb flash drive.
==> Packages to install (eg: 1 2 3, 1-3 or ^4)
==> 1
==> Error: Could not find all required packages:
    pyqt5-common (Wanted by: multibootusb)
```

ftsell:
I currently have the same problem in the following constellation:
(I know it requires a whole bunch of dependencies, sorry about that)

`yay -S --assume-installed ros-melodic-rqt-image-view ros-melodic-desktop-full`

the package `ros-melodic-rqt-image-view` is currently not compiling so I wish to skip it until a change is released.

Jguer:
I can confirm this issue

YourPsychiatrist:
Any updates here? :) Would really need this

jclsn:
I read somehwere that yay is abandoned and you should use paru instead, which is a maintained fork

irgendwr:
> I read somehwere that yay is abandoned and you should use paru instead, which is a maintained fork

\**checks commits\**  Last commit three days ago... yep looks totally abandoned :p
Also, paru is not a fork of yay. But this is all off-topic anyway.
*(Edit: I think you are referring to [this](https://www.reddit.com/r/archlinux/comments/jjn1c1/paru_v100_and_stepping_away_from_yay/))*

[...]

PR #1164, opened 2020-01-20
ignore dependency checks on double --nodeps

irgendwr:
This skips the dependency checks when `-d / --nodeps` is provided at least twice.

As mentioned in the [pacman(8) documentation](https://www.archlinux.org/pacman/pacman.8.html#_transaction_options_apply_to_em_s_em_em_r_em_and_em_u_em):

> -d, --nodeps
> Skips dependency version checks. Package names are still checked. Normally, pacman will always check a package’s dependency fields to ensure that all dependencies are installed and there are no package conflicts in the system. **Specify this option twice to skip all dependency checks.**

I mentioned this in #1137, however this pull-request does not solve the issue since `--assume-installed` is still not yet implemented.
I have a question regarding this: Where would it be best to skip dependencies? After `getDepPool(...)`? Or modify `getDepPool` to accept targets that should be assumed as installed?
I could also try to implement `--assume-installed` if you like.


irgendwr:
Hi @Jguer, could you please take a look at this?

Jguer:
Sorry it has taken a while, there was some refactoring I wanted to do, specially on argument parsing. I will be removing the cmdArgs/parser from the globals now so let's get this merged in and see if it's needs a bit of tweaking, 

`--assume-installed` I would prefer to mark it as satisfied in the depPool but I think it will require some trial and refactor.

PR #1174, opened 2020-01-27
Add missing diff completions for zsh

patatahooligan:


Jguer:
Thanks for your contribution

Issue #1165, opened 2020-01-20
`yay --devel` doesn't do anything


dschrempf:
#### Affected Version
yay v9.4.4 - libalpm v12.0.1
(sorry not the git version)

#### Issue
`yay --devel` doesn't do anything; no output; return value 0.

I don't have a custom yay config.

`yay -Syu --devel` works as expected.

(maybe this is not a bug, but since I am always using `yay` without command arguments, I expected that it should work with the `--devel` flag alone!)

Edit: I was too fast. I executed `yay -Y --gendb`, and now the updates are not shown anymore. So `yay -Syu --devel` does also not work as expected. Maybe related to #902 or #1016 ?

Jguer:
`yay` is `yay -Syu`, it's a shortcut but `yay --option` is `yay -Y --option`

dschrempf:
OK, thank you. This is totally confusing. You/we may think of changing this behavior.

Jguer:
`yay <package1> <package2> ` is yay/yogurt mode `-Y`, meaning `yay -Y <package1> <package2> `. 

Yay without any Operation specified was always meant to be the `-Y` operation (like yaourt did for example)

Then `yay` for `yay -Syu` was added as an inheritance for pacaur users. Let's just say the conjugation of both has never been harmonious. 

What I've thought of doing is continue to follow `-Syu` if no targets are specified as I think not many people use `yay --gendb` or `yay -c` as replacements for `yay -Y --gendb` or `yay -Yc`

Jguer:
Just pushed a branch that implements this proposed behaviour

Morganamilo:
I don't like this at all. I frequently use `yay --foo --save` and I also see it recommended/recommended it myself online.

Jguer:
`yay --foo --save` will still work this way, it will just trigger an update after.

But I do understand it's a breaking change for current behavior but as well, I understand the point of this issue. The compromise seems even worse which would be reverting to `-Y` if `--save` is present :sweat_smile:

Morganamilo:
I think how it works now is perfectly fine. I've never seen anybody complain about it before. I don't want to make this change just because it confused one person at first.

dschrempf:
Hi, I understand that the current behavior is, nearly always, the most user-friendly. I use just `yay` most of the time, and I am done with it. On the other hand, it is also highly inconsistent. I am sure that I am not the only user who was hit by this inconsistency. Maybe, I am the only one who filed an issue.

[...]

PR #1175, opened 2020-01-30
Move note `yay --option /= yay -Syu --option`.

dschrempf:
See issue #1165


Jguer:
Seems fine by me, thanks

PR #1191, opened 2020-02-16
Also install `*-debug` packages if built by makepkg

### LINKED ISSUE #1190: yay doesn't install *-debug packages created by makepkg
#### Affected Version
v9.4.4

#### Issue
When the `debug` and `strip` options are enabled in `makepkg.conf`, then `makepkg` also produces a `*-debug` package, which will automatically be installed by makepkg when given the `-i` option. When the same package is built using yay, the `*-debug` package is not installed.

#### Steps to reproduce
1. enable `OPTIONS=(strip [...] debug)` in `makepkg.conf`
2. build an install a package from AUR with `makepkg -i`
3. compare with `yay -S name`
4. observe `name-debug` package being built, but not installed

#### Notes
This might be interpreted as a 'feature request' of sorts, since many people don't actually use this feature of `makepkg`.

Ferdi265:
Make `yay` also install the separate `*-debug` packages which are created if both the `debug` and `strip` options are set in `makepkg.conf`.

This fixes #1190.

Also note: I don't usually code go, and just tried to follow the conventions of elsewhere in the file.

### DIFF
--- a/install.go
+++ b/install.go
@@ -1109,19 +1109,39 @@ func buildInstallPkgbuilds(dp *depPool, do *depOrder, srcinfos map[string]*gosrc
 			}
 		}
 
-		for _, split := range base {
-			pkgdest, ok := pkgdests[split.Name]
+		doAddTarget := func(name string, optional bool) error {
+			pkgdest, ok := pkgdests[name]
 			if !ok {
-				return fmt.Errorf("Could not find PKGDEST for: %s", split.Name)
+				if !optional {
+					return fmt.Errorf("Could not find PKGDEST for: %s", name)
+				} else {
+					return nil
+				}
 			}
 
 			arguments.addTarget(pkgdest)
 			if parser.existsArg("asdeps", "asdep") {
-				deps = append(deps, split.Name)
+				deps = append(deps, name)
 			} else if parser.existsArg("asexplicit", "asexp") {
-				exp = append(exp, split.Name)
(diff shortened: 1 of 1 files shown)


Ferdi265:
I found a bug in this PR and fixed in the latest commit:

Sometimes `*-debug` packages generated by makepkg are listed in `makepkg --packagelist`, but are not built when running `makepkg`.

This is the case when no debug info sections could be found in the package, which is often the case for `python-*` packages, like `python-bashlex`.

before this commit, `yay` would fail to install `python-bashlex` if the `strip` and `debug` option were both set in `/etc/makepkg.conf`. This fixes `yay` to ignore missing `*-debug` packages.

Ferdi265:
(rebased onto latest master)

Ferdi265:
(another rebase)

Ferdi265:
(added commit signature)

Ferdi265:
(added signature for all commits in branch)

Jguer:
Hey, I'm sorry it's taken me so long to get to your MR, I would like one or two code style changes like reducing the scope of that err but since we've taken so long to review it I'll let it pass. 

```go
if err := doAddTarget(split.Name, false);err != nil {
  return err
}
```

Thanks for your contribution and as well your patience :sweat_smile: , this will be available on `-next` for a bit so we iron out any kinks

Ferdi265:
Thanks! I don't usually code in go, so I didn't know about that if with declaration syntax yet.

Thanks for merging :)

PR #1197, opened 2020-02-24
add dependencies for manual installation, fixes #1196

### LINKED ISSUE #1196: add `pacman -S fakeroot binutils` to installation
#### Issue
When installing yay on a fresh arch install it throws errors about missing packages. This can be fixed by installing fakeroot and binutils

#### Steps to reproduce
1. install the latest arch linux
2. follow the steps provided in the installation section of _yay_

#### Output
```
git clone https://aur.archlinux.org/yay.git
cd yay
makepkg -si
==> ERROR: Cannot find the fakeroot binary.
==> ERROR: Cannot find the strip binary required for object file stripping.
```

RubenKelevra:


Morganamilo:
```
morganamilo@Octavia ~ % pacman -Si fakeroot | grep Group
Groups          : base-devel
```

Morganamilo:
Aside from that, I don't really like hold handing the users too much. Especially when people *should* only use AUR helpers after they understand the manual process and how the AUR works.

I don't really think we need to tell users to install git either. I think a user should know that they need to install a program to be able to use it. And I especially object to the use of noconfrim.

What I'd prefer is to just have step 0 be: read the arch wiki page for the aur, and understand the manual process.

RubenKelevra:
> ```
> morganamilo@Octavia ~ % pacman -Si fakeroot | grep Group
> Groups          : base-devel
> ```

I might have been too tired ;)

---

> Aside from that, I don't really like hold handing the users too much. Especially when people _should_ only use AUR helpers after they understand the manual process and how the AUR works.
> 
> I don't really think we need to tell users to install git either. I think a user should know that they need to install a program to be able to use it. And I especially object to the use of noconfrim.
> 
> What I'd prefer is to just have step 0 be: read the arch wiki page for the aur, and understand the manual process.

I do completely agree, especially since everything is already provided by base-devel.

I don't think there's *any* need for additional documentation then.

Issue #1209, opened 2020-03-13
Custom GOPROXY issue


zpdcodes:
#### Issue

The Makefile contains `export GOPROXY=https://proxy.golang.org`, which overrides `GOPROXY` in `go env` settings. However the `proxy.golang.org` connection is unstable and sometime unreachable in my network. It cause some trouble when trying to run `makepkg -si`. Can we remove it from Makefile and let users choose whatever proxy they prefer?

zpdcodes:
The latest `next` branch would prefer user-set GOPROXY over default one. Close this.

PR #1210, opened 2020-03-13
Prefer user-defined GOPROXY in Makefile

zpdcodes:
Related issue: #1209 

Use `GOPROXY` if it's set, otherwise default to `https://proxy.golang.org`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 export GO111MODULE=on
-export GOPROXY=https://proxy.golang.org
+GOPROXY ?= https://proxy.golang.org
+export GOPROXY
 
 BUILD_TAG = devel
 ARCH ?= $(shell uname -m)


Makefile
```diff
@@ -1,5 +1,5 @@
 export GO111MODULE=on
-export GOPROXY=https://proxy.golang.org
+export GOPROXY="${CUSTOM_GOPROXY:-https://proxy.golang.org}"
```
Jguer: Can you test if
 
```
GOPROXY ?= https://proxy.golang.org
export GOPROXY=$(GOPROXY)
```
would work, then you wouldn't need to set a `CUSTOM_GOPROXY` but instead use the one you desire

zpdcodes: @Jguer Thanks for the hint. I just updated the code to remove unnecessary `CUSTOM_GOPROXY`.

Jguer:
Thanks for your contribution!

PR #1212, opened 2020-03-14
Treat whole split-package as devel when pkgbase matches too

Prototik:
I got false-positive reports for `linux-git` pkgbase, it provides package `linux-git-headers` and yay doesn't honor that fact that whole pkgbase is devel:

```
:: Checking development packages...
 -> linux-git-headers: local (5.6rc5.r241.g69a4d0baeeb1-1) is newer than AUR (5.6rc1.r5.g0a679e13ea30-1)
```

### DIFF
--- a/dep.go
+++ b/dep.go
@@ -172,3 +172,7 @@ func isDevelName(name string) bool {
 
 	return strings.Contains(name, "-always-")
 }
+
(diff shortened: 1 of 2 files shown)


Jguer:
Thanks for your contribution!

PR #1217, opened 2020-03-19
Use golang.org's terminal package to determine isTty()

invisiblek:
* Don't trust that the user doesn't have a "test" binary in their
  PATH somewhere, and instead use the official golang library to
  determine whether or not we're running in a terminal

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"fmt"
+	"golang.org/x/crypto/ssh/terminal"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


Jguer:
from an official library I may consider adding it, can you give a size delta on the static binaries?

invisiblek:
I can't seem to get this thing compiled statically, complains of alpm things. A resulting, dynamically-linked, binary is a few bytes bigger.

Jguer:
Sorry, should have meant everything static except libalpm
`next`: 6513520
`istty_fix`: 6517936

delta:
+0.06%

Jguer:
Seems fine by me, I'm taking it in, thanks for your help

Issue #1207, opened 2020-03-11
panic if vim not in $PATH


sosiska:
#### Affected Version
yay v9.4.6 - libalpm v12.0.1
#### Issue
panic if `vim` not in $PATH

1. Remove vim from $PATH
2. yay -S oracle-instantclient-sqlplus --editmenu
3. Select edit PKGBUILD
#### Output
`yay -S oracle-instantclient-sqlplus --editmenu`                                                                                                                                
```
:: Checking for conflicts...
:: Checking for inner conflicts...
[Aur: 2]  oracle-instantclient-basic-19.5.0.0.0-1  oracle-instantclient-sqlplus-19.5.0.0.0-1

  2 oracle-instantclient-basic       (Build Files Exist)
  1 oracle-instantclient-sqlplus     (Build Files Exist)
==> Packages to cleanBuild?
==> [N]one [A]ll [Ab]ort [I]nstalled [No]tInstalled or (1 2 3, 1-3, ^4)
==> n
:: PKGBUILD up to date, Skipping (1/2): oracle-instantclient-sqlplus
:: PKGBUILD up to date, Skipping (2/2): oracle-instantclient-basic
  2 oracle-instantclient-basic       (Build Files Exist)
  1 oracle-instantclient-sqlplus     (Build Files Exist)
==> Diffs to show?
==> [N]one [A]ll [Ab]ort [I]nstalled [No]tInstalled or (1 2 3, 1-3, ^4)
==> n
:: Parsing SRCINFO (1/2): oracle-instantclient-basic
:: Parsing SRCINFO (2/2): oracle-instantclient-sqlplus
  2 oracle-instantclient-basic       (Build Files Exist)
  1 oracle-instantclient-sqlplus     (Build Files Exist)
==> PKGBUILDs to edit?
==> [N]one [A]ll [Ab]ort [I]nstalled [No]tInstalled or (1 2 3, 1-3, ^4)
==> 1
exec: "vim": executable file not found in $PATH
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
main.editor(0xc0004762c0, 0xc00014e090, 0x1, 0x1, 0xc000436a50)
	config.go:247 +0x1172
main.editPkgbuilds(0xc0003f2da0, 0x1, 0x1, 0xc00039c3c0, 0xc00039c300, 0xc0003f2da0)
	install.go:784 +0x476
main.install(0xc000012380, 0x0, 0x0)
	install.go:292 +0x1e60
[...]

PR #1239, opened 2020-04-12
check len before accessing element

sosiska:
This PR closes https://github.com/Jguer/yay/issues/1207
In some cases, we can get into the `case:` without checking the conditions through the `fallthrough`. Then there is an appeal to a nonexistent element, which leads to panic. If check for availability before applying, we can safely `fallthrough` to `default:` block.


Jguer:
Indeed, I would prefer to get this code refactored, but this seems like a simple bugfix for now, thanks

PR #1260, opened 2020-05-04
Fix segfault from 9fccdcb

themadprofessor:
9fccdcb renamed a variable, but didn't update all uses of the variable.


Jguer:
Indeed, good catch, thanks

PR #1262, opened 2020-05-09
implement yay -Pp option

### LINKED ISSUE #1223: Print PKGBUILD to stdout
yay v9.4.6 - libalpm v12.0.1
#### Feature Request
It'd be great if yay had the possibility to not just download the PKGBUILD to the cwd, but also to just display it to stdout. Often i want to look at how a package is built and find myself looking it up in the browser because i don't want to have the pkgbuild file saved, but just shown.

nycex:
Fixes #1223 

This pull request adds the `yay -Pp <packages>` or in long `yay -P --pkgbuild <packages>`
It prints the content of the PKGBUILD of the package to stdout.

I went for the approach to make http requests to the raw PKGBUILDS directly instead of downloading the git repos since that doesn't need any temp files and doesn't download unused data. It won't work for packages in repos that are non-official though, but I don't think that would be possible even with a different approach.

Note: This is more or less my first real programming in golang so there may be some bad/unidiomatic code in it.

Edit: I'm not actually sure if the zsh, bash and fish completions work, I'll test it. 
Edit2: completions are now fixed


completions/zsh
```diff
@@ -171,6 +171,7 @@ _pacman_opts_print_modifiers=(
 		{-s,--stats}'[Display system package statistics]'
 		{-u,--upgrades}'[Print update list]'
 		{-w,--news}'[Print arch news]'
+		{-p,--pkgbuild}'[Print arch news]:package:_pacman_completions_all_packages'
```
Jguer: Needs change of `Print arch news` to `Print PKGBUILDs`

query.go
```diff
@@ -574,3 +577,140 @@ func aurInfoPrint(names []string) ([]*rpc.Pkg, error) {
...
+
+	makeRequest := func(name string) {
+		defer wg.Done()
+
+		values := url.Values{}
+		values.Set("h", name)
+
+		url := "https://aur.archlinux.org/cgit/aur.git/plain/PKGBUILD?"
+
+		resp, err := http.Get(url + values.Encode())
```
Jguer: use https://golang.org/pkg/net/url/#Parse to make urls and join them

query.go
```diff
@@ -574,3 +577,140 @@ func aurInfoPrint(names []string) ([]*rpc.Pkg, error) {
...
+			errs.Add(fmt.Errorf("package \"%s\" was not found", name))
+			return
+		}
+
+		mux.Lock()
+		pkgbuilds = append(pkgbuilds, pkgbuild)
+		mux.Unlock()
+	}
+
+	for _, name := range names {
```
Jguer: Not blocking: I would prefer it to be done in chunks of 20? let's say because if someone adds 100 targets it will be a bit of a mess, but it's a lot of work for a niche edge case.

query.go
```diff
@@ -574,3 +577,140 @@ func aurInfoPrint(names []string) ([]*rpc.Pkg, error) {
...
+func repoPkgbuilds(names []string) ([]string, error) {
+	pkgbuilds := make([]string, 0, len(names))
+	var mux sync.Mutex
+	var wg sync.WaitGroup
+	var errs multierror.MultiError
+
+	makeRequest := func(full string) {
+		defer wg.Done()
+
+		dbList, dbErr := alpmHandle.SyncDBs()
```
Jguer: initialize dbList outside makeRequest

query.go
```diff
@@ -574,3 +577,140 @@ func aurInfoPrint(names []string) ([]*rpc.Pkg, error) {
...
+		if dbErr != nil {
+			errs.Add(dbErr)
+			return
+		}
+
+		db, name := splitDBFromName(full)
+
+		if db == "" {
+			var pkg *alpm.Package
+			for _, alpmDB := range dbList.Slice() {
```
Jguer: Fixing the array should also be done out of `makeRequest` so you pass through dbList only once (it's way bigger than targets)

query.go
```diff
@@ -574,3 +577,140 @@ func aurInfoPrint(names []string) ([]*rpc.Pkg, error) {
...
+		go makeRequest(full)
+	}
+
+	wg.Wait()
+
+	if err := errs.Return(); err != nil {
+		return pkgbuilds, err
+	}
+
+	return pkgbuilds, nil
```
Jguer: `return pkgbuilds, errs.Return()` instead

query.go
nycex: why should I use that there? the values I encode there aren't a url of themselves and http.Get accepts a string and not a URL

query.go
nycex: What do you mean by "mess"? The PKGBUILDs are downloaded concurrently but written in sync.

query.go
nycex: I'm not sure about this one, shouldn't dbList be only like `core`, `community` and such?

query.go
Jguer: Sometimes when we open too many concurrent connections against the AUR we get rate limit banned, which is why yay chunks request making, just warning that this seems the same unlimited pattern that gave us that problem,

query.go
Jguer: My bad, I was distracted, I thought you were going through the `PKGCache`. the Slice() conversion could be done outside of the anonymous function.

query.go
```diff
@@ -574,3 +577,132 @@ func aurInfoPrint(names []string) ([]*rpc.Pkg, error) {
...
+		body, readErr := ioutil.ReadAll(resp.Body)
+		pkgbuild := string(body)
+
+		if readErr != nil {
+			errs.Add(readErr)
+			return
+		}
+
+		if strings.Contains(pkgbuild,
+			"<div class='content'><div class='error'>Invalid branch: "+name+"</div>") {
```
Jguer: `https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ewemwe` seems to return 404, why not check `resp.StatusCode` instead, and add err after `http.Get` if `StatusCode != 200`

query.go
```diff
@@ -574,3 +577,132 @@ func aurInfoPrint(names []string) ([]*rpc.Pkg, error) {
...
+		case "core", "extra", "testing":
+			url = "https://git.archlinux.org/svntogit/packages.git/plain/trunk/PKGBUILD?"
+		case "community", "multilib", "community-testing", "multilib-testing":
+			url = "https://git.archlinux.org/svntogit/community.git/plain/trunk/PKGBUILD?"
+		default:
+			errs.Add(fmt.Errorf("unable to get PKGBUILD from repo \"%s\"", db))
+			return
+		}
+
+		resp, err := http.Get(url + values.Encode())
```
Jguer, line 484: resp status code != 200 not checked here

nycex:
Is there still something blocking this PR?

Morganamilo:
In paru I named this `-Gp` or `--getpkgbuild --print`. If you wanna be consistent with it :P

Jguer:
> Is there still something blocking this PR?

@nycex 

There is not, I've rebased and squashed it on https://github.com/Jguer/yay/commits/jguer/-/pkgbuild-print , I'll just add a commit to move it to `-Gp` and I'll merge it into next.

Jguer:
@nycex I've extracted -Gp into smaller manageable parts and added a few tests. Once again, thank you for your contribution and I'm sorry it's taken this long to get it into mainline.


I've also renamed it to `-G --print` for consistency

PR #1271, opened 2020-05-20
adds option for fuzzy finder instead of just number menu

conradludgate:
Using --fuzzy, packages are presented using https://github.com/ktr0731/go-fuzzyfinder (a library that is similar to fzf). Multiple packages can be selected using tab.

String formatting could use some improvement. The fuzzy finder doesn't support multiline so descriptions are on the same line. Also could use better documentation, I'm unsure of what to put.

Fuzzy finder also can present extra information about the current hover. Perhaps it could be used to show aur comments, or some other information

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -112,9 +112,9 @@ Permanent configuration options:
     --combinedupgrade     Refresh then perform the repo and AUR upgrade together
     --nocombinedupgrade   Perform the repo upgrade and AUR upgrade separately
     --batchinstall        Build multiple AUR packages then install them together
-	--nobatchinstall      Build and install each AUR package one by one
-	--fuzzy            Use fuzzy finding instead of number selection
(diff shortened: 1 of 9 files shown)


JojiiOfficial:
The fuzzy menu doesn't display colors correctly. The color codes are displayed as text in the fuzzy menu

conradludgate:
Yes you are right, I didn't ever use the color option so I guess I missed that. Not sure how best to fix that since that seems to be a problem with the fuzzy finder library. Perhaps disabling color in fuzzy as a simple solution. 

I guess what we really need is a fuzzy finder that support color, perhaps uses underlines to highlight which characters match the typed input and supports multiline fields so that descriptions can be put on the line below and not be so cluttered.

If I have time I might look into implementing that but I don't have much of it at the moment

PR #1281, opened 2020-06-03
Add Simplified Chinese Translation

Hill-98:
Add Simplified Chinese Translation


Jguer:
Checked some translations using google translate, seems fine. Anyone that can verify please do :sweat_smile:

PR #1285, opened 2020-06-07
remove percent sign from popularity in `yay -Ss`

nycex:
The popularity in the AUR is no percentage, therefore the percent sign is rather confusing there.
In `yay -Si`, the popularity is shown correctly as a normal number.
I have looked for other occurences of the percent sign, but there still may be one.


Jguer:
Seems fine by me, thank you @nycex , I'll get back on our other MR once things have calmed down a bit with i18n

PR #1288, opened 2020-06-08
Update zh_CN.po

Hill-98:
Improve Simplified Chinese translation


Jguer:
Thanks for your contribution :smile:

PR #1292, opened 2020-06-14
Add Spanish translation

MarioJim:


Jguer:
Thank you for your contribution @MarioJim

MarioJim:
Glad I could help!

PR #1294, opened 2020-06-14
fix(locale): fix errors in the Simplified Chinese locale

ouuan:
Fix some errors in the Simplified Chinese locale.


Jguer:
@Hill-98 You were the first one to add the translation so your input would be nice, I may move translation to transifex or another utility to make it easier to review

Hill-98:
I didn't actually test where each text appeared, so it was a bit inaccurate, and this fix improved a lot.

Jguer:
Fine by me then :+1: , We'll always hear if there's an issue with it after. Thanks for your contribution

PR #1302, opened 2020-06-16
Improving translation for PT

guiananias:
Hello!
I made some improvements regarding the translation into Portuguese, among them:

- Translation correction;
- Ortography correction;

### DIFF
--- a/po/pt.po
+++ b/po/pt.po
@@ -23,7 +23,7 @@ msgstr " (Alvo"
 
 #: depCheck.go:281
 msgid " (Wanted by: "
-msgstr " (Querido por: "
+msgstr " (Requerido por: "
 
 #: callbacks.go:72
 msgid " Input too long"
@@ -48,7 +48,7 @@ msgstr ""
 
 #: download.go:273
 msgid "%s already downloaded -- use -f to overwrite"
-msgstr "%s já descarregado -- utilize -f para forçar"
+msgstr "%s já baixado -- utilize -f para forçar"
 
 #: install.go:1084
 msgid "%s already made -- skipping build"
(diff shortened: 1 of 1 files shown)


po/pt.po
```diff
@@ -182,11 +182,11 @@ msgstr "Diffs a mostrar?"
 
 #: clean.go:89
 msgid "Do you want to remove ALL AUR packages from cache?"
-msgstr "Quer remover todos os pacotes AUR da cache?"
+msgstr "Quer remover todos os pacotes AUR do cache?"
```
Jguer, line 185: I follow the priberam dictionary on this so cache should be feminine

https://dicionario.priberam.org/cache

guiananias: Yes Yes. If it is the case to follow the dictionary, cache is a feminine word, you are right.

Is it that, generally speaking, we tend to bring the "cache" to the male? Now why is a good question. :joy:

I will make the correction, thank you.

agendator:
Hi! I'd like to point out that these changes made by @guiananias are meant for a Brazilian Portuguese translation, which is not the case. This input will break what was meant to be a Portugal Portuguese translation. The best solution is to make a separate ptBR translation, since ptPT and ptBR have so many differences.

guiananias:
@agendator yeah, good point i forgot to mention. 

Indeed, in this case, the use of Portuguese is actually located in Brazilian Portuguese, which, although not different in writing, are languages that are quite different.

Jguer:
I use `pt_PT` and I can understand these fixes to the translation ( crear and analizar :man_facepalming: ). 

The most shocking differences for me is `usuario` versus `utilizador` and `baixado` vs `descarregado` but even for the `pt_PT` translations I would keep some of these changes.

I propose we do like [pacman](https://git.archlinux.org/pacman.git/tree/src/pacman/po) and use `pt.po` for pt_PT and `pt_BR.po` for pt_BR.

On this MR:
- Keep the changes to the `pt.po` as they are, I'll cherrypick the most appropriate
- Add a new commit with a `pt_BR.po` (don't forget to add pt_BR to the makefile `LANGS`) 

Would that work for you @guiananias ?

guiananias:
@Jguer it's fine by me. I will add the new file.

PR #1303, opened 2020-06-16
Russian translation

dvdesolve:
The very first Russian translation


Jguer:
Thank you @dvdesolve , I'm sure users will give feedback on it once they start to use it

PR #1310, opened 2020-06-19
Improve print format under non-English localization

Hill-98:
Under the localization of Chinese, the current print format of `yay -Si` is a little strange, and I tried to improve it.
Perhaps this is not the best solution, but I am not very familiar with the `printf` format.

### DIFF
--- a/pkg/text/print.go
+++ b/pkg/text/print.go
@@ -50,5 +50,5 @@ func PrintInfoValue(str, value string) {
 		value = gotext.Get("None")
 	}
 
-	fmt.Fprintf(os.Stdout, bold("%-16s%s")+" %s\n", str, ":", value)
+	fmt.Fprintf(os.Stdout, bold("%-16s\t%s")+" %s\n", str, ":", value)
 }


Jguer:
Does `pacman -Si` look equally strange?

Hill-98:
`pacman -Si` looks normal

Hill-98:
Even if it is localized in English, yay print format is not as good as pacman, I think this is a place that can be improved.

Jguer:
Inded, we should improve this by learning upstream
https://git.archlinux.org/pacman.git/tree/src/pacman/package.c#n192

https://git.archlinux.org/pacman.git/tree/src/pacman/util.c <- most of the string formatting is here

Hill-98:
I have seen upstream code, but C code is a little difficult for me to understand.

Jguer:
Hey @Hill-98 , I'm closing this until we can figure out how to make it closer to the original pacman code, if you could turn it into an issue that would be great.

PR #1316, opened 2020-06-26
Fix #1279: Error not git repository on yay -Sc

### LINKED ISSUE #1279: yay -Sc bug?
#### Affected Version

```
tata@floki ~ % yay -V
yay v9.4.7 - libalpm v12.0.1
```

#### Issue

`yay -Sc`

#### Output

```
tata@floki ~ % yay -Sc        
Packages to keep:
  All locally installed packages

:: Do you want to remove all other packages from cache? [Y/n] 
removing old packages from cache...

Database directory: /var/lib/pacman/
:: Do you want to remove unused repositories? [Y/n] 
removing unused sync repositories...

Build directory: /home/tata/.cache/yay
==> Do you want to remove all other AUR packages from cache? [Y/n] 
removing AUR packages from cache...
==> Do you want to remove ALL untracked AUR files? [Y/n] 
removing Untracked AUR files from cache...
fatal: not a git repository (or any of the parent directories): .git
```

msAlcantara:
This PR include a simple validation to check if a folder is a git repository before to clean.

This validation make sense to all scenarios of yay? Sorry if something here make no sense, I'm starting with yay source code.

### DIFF
--- a/clean.go
+++ b/clean.go
@@ -196,12 +196,18 @@ func cleanUntracked() error {
 		}
 
 		dir := filepath.Join(config.BuildDir, file.Name())
-		if err := show(passToGit(dir, "clean", "-fx")); err != nil {
-			return err
+		if isGitRepository(dir) {
+			if err := show(passToGit(dir, "clean", "-fx")); err != nil {
+				return err
+			}
 		}
 	}
-
 	return nil
+}
+
+func isGitRepository(dir string) bool {
+	_, err := os.Stat(filepath.Join(dir, ".git"))
+	return !os.IsNotExist(err)
 }
 
 func cleanAfter(bases []Base) {


clean.go
```diff
@@ -196,14 +196,20 @@ func cleanUntracked() error {
...
+				return err
+			}
 		}
 	}
-
 	return nil
 }
 
+func isGitRepository(dir string) bool {
+	cmd := passToGit(dir, "rev-parse")
```
Jguer, line 209: Couldn't we just do a 
```go
_, err := os.Stat(filepath.Join(dir, ".git"))
return  !os.IsNotExist(err)
```

To check the existence of a `.git`? It's not as perfect as letting git check but it would avoid shelling out for it

msAlcantara:
Thanks for reviews, commit updated

Jguer:
Thanks for your contribution, I think it should be ok, let's see it in `yay-git`!

PR #1319, opened 2020-07-01
Fix "%s is up to date -- skipping" message formatting

Giftpflanze:
%s is not substituted because the argument is not in the argument list.


Jguer:
My bad :smile: , thanks for your contribution

PR #1324, opened 2020-07-13
Add Japanese translation

kusakata:


Jguer:
Looks mostly fine here, merging and expecting feedback from other users.

Thanks for you contribution!

PR #1333, opened 2020-07-23
Turn config parsing into a struct

Morganamilo:
I finally got around to finishing the reflect-config branch. This is an alternative to #717 and it's missing the same things at that PR, docs, completions, etc. Also the warning about the old config needs to be added.

This is not fully tested but it works Feel free to give it a shot. Yay will work with no config file provided, but will read the config file from /etc/yay.conf or ~/.cache/yay/yay.conf. The repo has a template config under conf/yay.conf. the pkgbuild should move it into /etc, but you can just copy it yourself for testing.

Also this is less code than before :)


grandchild:
Why is the config file in `~/.cache/`? Wouldn't `~/.config/` be the more logical location?

Morganamilo:
> Why is the config file in `~/.cache/`? Wouldn't `~/.config/` be the more logical location?

typo

Jguer:
Right now my main two objectives are the interfacing of alpm and the shell executors to provide strong basis for testing, after that I'm interested in pulling this MR in (as well as sifting through others) and ensure proper tests are included in each

Issue #1335, opened 2020-08-02
Failure to build a package prints incorrect package name


kraxarn:
#### Affected Version
yay v10.0.3 - libalpm v12.0.2
yay v10.0.3.r18.g8074179 - libalpm v12.0.2

**Describe the bug**
When failing to build a package from AUR, the package name when reporting the name of the package in the error message is incorrectly printed, not showing the package name at all.

**To Reproduce**
* Build package from AUR that fails.
* After failure, package name is printed incorrectly.

**Expected behavior**
Expected print of actual package name.
```
error making: <package-name>
```

**Output**
Incorrect package name
```
yay -S <package-name>
...
error making: %!s(func() string=0xaaa5bdf832d1)
```
Same happens on a clean install without any configuration changes.

kraxarn:
Confirmed fixed in #1341

Issue #1325, opened 2020-07-14
`yay --save` doesn't insert newline at end of file, and removes it if it exists


lberrymage:
#### Affected Version

```
$ yay -V
yay v10.0.2.r38.gfd7934d - libalpm v12.0.2
```

**Describe the bug**
`yay --save` doesn't insert a newline at the end of the config file, and removes it if it exists.

**To Reproduce**
```
$ yay -Pg | grep devel
	"devel": false,
$ tail -c1 ~/.config/yay/config.json | wc -l
1
$ yay --save --devel
$ tail -c1 ~/.config/yay/config.json | wc -l
0
```

**Expected behavior**
The tail command run should always return 1, since this indicates a newline at the end of the file. This is required according to the [POSIX definition of a line](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)

lberrymage:
Closing as resolved due to #1341

PR #1341, opened 2020-08-07
Fix issues

Feresey:
<https://github.com/Jguer/yay/issues/1335>
<https://github.com/Jguer/yay/issues/1325>


Jguer:
Thanks for your contribution :smile:

PR #1344, opened 2020-08-10
fix(translate) : Japanese

Hayao0819:
Improved some unnatural Japanese translations.


Jguer:
@kusakata care to review?

Issue #1142, opened 2020-01-05
Add dependencies for installation


hofmannu:
Actually not an issue of `yay` itself but more of the documentation here. I think there are some dependencies to install `yay`. Maybe we should add them into the documentation. I just tried to install it on a fresh system and got the following error when I tried to run `makepkg -si`:

    ==> ERROR: Cannot find the fakeroot binary.
    ==> ERROR: Cannot find the strip binary required for object file stripping.

which indicates that `fakeroot` is one dependency. I just installed the whole `base-devel` group and then it ran through, so I am not sure what the core dependencies are but it would be good to have them listed here in the documentation within the section Installation. Also good to know for people who use scripts to set up their systems since it defines the installation order.

Jguer:
https://wiki.archlinux.org/index.php/Arch_User_Repository#Getting_started
https://wiki.archlinux.org/index.php/PKGBUILD#makedepends
> Note: The group base-devel is assumed to be already installed when building with makepkg. Members of this group should not be included in makedepends array.
`base-devel` is a per-requesite

UnlimitedCookies:
Sorry to bring up this stale issue again.

Why don't we add `fakeroot` to the dependencies? It is required for the core functionality of yay.

Jguer:
https://wiki.archlinux.org/index.php/Arch_User_Repository#Getting_started
https://wiki.archlinux.org/index.php/PKGBUILD#makedepends

As written on the wiki:

> Note: The group base-devel is assumed to be already installed when building with makepkg. Members of this group should not be included in makedepends array.

It's required for a core functionality of makepkg and the group base-devel is assumed to be already installed when building with makepkg. I'm ok with adding a note to the `README.md` reminding users to have `base-devel` installed

PR #1345, opened 2020-08-11
Updated Readme to include note about dependencies

AdamGEmerson:
Per issue #1142 I've added a note to the readme to let users know they'll need to install `base-devel`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,15 +41,13 @@ If you are migrating from another AUR helper, you can simply install Yay with th
 Alternatively, the initial installation of Yay can be done by cloning the PKGBUILD and
 building with makepkg:
 
+Make sure you have the `base-devel` package group installed
+
 ```sh
 git clone https://aur.archlinux.org/yay.git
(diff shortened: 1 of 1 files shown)


grandchild:
I would write plainly "Make sure you have the `base-devel` package group installed" -- since it's not just people who "have an issue with makepkg", everyone must have it installed.

People might overlook it initially, because... well, they don't have a problem with makepkg... yet. ;)

AdamGEmerson:
I have updated the readme per @grandchild notes.

Let me know if you think there should be any addtional formatting tweaks.

Jguer:
Thanks for your contribution, it is true that this should have been fixed a long time ago :sweat_smile:

AdamGEmerson:
> Thanks for your contribution, it is true that this should have been fixed a long time ago 😅

No problem!  I'm just getting started trying to contribute on GitHub.  I'll keep a look out for issues I can tackle.

Issue #1352, opened 2020-08-27
yay-git saves settings failed


arenekosreal:
#### Affected Version

yay v10.0.4.r24.g3b2bb28 - libalpm v12.0.2

**Describe the bug**
Execute `yay --aururl "https://aur.tuna.tsinghua.edu.cn" --save` and terminal returns "\<nil\>". If I replace `yay-git` with `yay`, this problem can be solved. I am using Manjaro on RaspberryPi 4.
**To Reproduce**
Execute `yay --aururl "https://aur.tuna.tsinghua.edu.cn" --save`  in terminal.
**Expected behavior**
Replace AUR URL with mirror site and save changes to file to make it also be applied next time.
**Output**
Terminal output:
```
# zhanghua @ raspberrypi in ~ [17:04:46]
$ yay --aururl "https://aur.tuna.tsinghua.edu.cn" --save
<nil>
```
yay config:
```
# zhanghua @ raspberrypi in ~ [17:05:47]
$ yay -Pg
{
        "aururl": "https://aur.archlinux.org",
        "buildDir": "/home/zhanghua/.cache/yay",
        "absdir": "/home/zhanghua/.cache/yay/abs",
        "editor": "",
        "editorflags": "",
        "makepkgbin": "makepkg",
        "makepkgconf": "",
        "pacmanbin": "pacman",
        "pacmanconf": "/etc/pacman.conf",
        "redownload": "no",
        "rebuild": "no",
        "answerclean": "",
        "answerdiff": "",
        "answeredit": "",
        "answerupgrade": "",
        "gitbin": "git",
        "gpgbin": "gpg",
        "gpgflags": "",
        "mflags": "",
        "sortby": "votes",
        "searchby": "name-desc",
        "gitflags": "",
        "removemake": "ask",
        "sudobin": "sudo",
        "sudoflags": "",
        "requestsplitn": 150,
        "sortmode": 0,
        "completionrefreshtime": 7,
        "sudoloop": false,
        "timeupdate": false,
        "devel": false,
        "cleanAfter": false,
        "provides": true,
        "pgpfetch": true,
        "upgrademenu": true,
        "cleanmenu": true,
        "diffmenu": true,
        "editmenu": false,
        "combinedupgrade": false,
        "useask": false,
        "batchinstall": false
}
```

mattheusv:
Seems that runtime.ConfigPath is not filled on next branch

[...]

PR #1353, opened 2020-08-27
Fix(config): Error to save config #1352

mattheusv:
runtime.Config Path is not set on config when create Runtime struct, and the config path is not created if not exists, so this pr fix save config operation.

### DIFF
--- a/pkg/settings/config.go
+++ b/pkg/settings/config.go
@@ -183,7 +183,11 @@ func NewConfig() (*Configuration, error) {
 	cacheHome := getCacheHome()
 	newConfig.BuildDir = cacheHome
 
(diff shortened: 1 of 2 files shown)


Jguer:
My objective is to allow yay to work even without a config file, so even if there's an error finding the config path, execution should continue

mattheusv:
Even when we try to save config?

Jguer:
> Even when we try to save config?

Yes, I wish I could find the issue for it but it's a breaking point for running as `nobody`

mattheusv:
Okay, I understand, if you can share the next steps for setting up yay later, maybe I can help you. Thanks.
So, I'm closing the pr.

Issue #1332, opened 2020-07-23
`yay -Si` print format should be like pacman


Hill-98:
The current print format of `yay -Si` is bad in a non-English environment (such as the following), while the print format of pacman is very good. Even in the English environment, the print format is still not as good as pacman.

![1.png](https://i.loli.net/2020/06/28/mDIOBSdN1ho3k7u.png)
![2.png](https://i.loli.net/2020/06/28/wRXIAOfxNdcG5vp.png)

We should be able to learn from upstream to make the printing behavior of yay and pacman consistent.

grandchild:
For reference, this issue was pruned from #1310

lwlf:
Has this issue been fixed? If fixed, how will I configure and enable it? 
@Jguer

PR #1356, opened 2020-08-30
fix(display): adapt padding to match pacman's.

ghost:
This change fixes part of #1332 related to splitting information into
multiple lines.
Change is implemented based on current pacman's code.


Jguer:
Tested, seems to work fine, would have liked some snapshot tests on it but since it's a poorly tested area I won't put the burden on you. Thanks for your contribution

PR #1362, opened 2020-09-07
Add dependencies to the installation instructions.

ghost:
Expected: installation instructions explain how to install yay.
Actual: `makepkg -si` fails unless the dependencies are installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ building with makepkg:
 Before you begin, make sure you have the `base-devel` package group installed.
 
 ```sh
+pacman -S git fakeroot binutils gcc make
 git clone https://aur.archlinux.org/yay.git
 cd yay
 makepkg -si


Morganamilo:
> Before you begin, make sure you have the `base-devel` package group installed.

README.md
```diff
@@ -44,6 +44,7 @@ building with makepkg:
 Before you begin, make sure you have the `base-devel` package group installed.
 
 ```sh
+pacman -S git fakeroot binutils gcc make
```
Jguer, line 47:
```suggestion
pacman -S --needed git base-devel
```

Morganamilo:
If we're going to go that route and add it as an explicit command then you should probably add `--needed` to it.

README.md
ghost: You can clean up build artifacts by using a multi-stage Dockerfile, however downloading packages that are never used goes against [Lean Flow](https://www.google.com/search?q=lean+flow) principles.  Your batch size and process time will not be as small as possible.  Determine that `git fakeroot binutils gcc make` are the smallest number of dependencies takes time too.

grandchild: If having the smallest subset of packages to build yay and _only_ yay is your goal, your argument is reasonable.

However, I would consider another goal, namely that _after_ installing yay, you want to install other AUR packages. And many of those have the implicit assumption that _all_ of base-devel is installed (and won't list them in their dependencies). Failure to instruct users to install base-devel completely will probably result in people coming here or elsewhere to complain about missing package dependencies.

Jguer: https://wiki.archlinux.org/index.php/Arch_User_Repository#Getting_started

`Ensure the base-devel package group is installed in full (pacman -S --needed base-devel).`

it's a "minimum requirement" for using the AUR. 

A more experienced user will know how to install with the bare minimum of dependencies. If you want as a compromise you may add a note that to build yay itself only x,y,z are used from base-devel

grandchild: > you may add a note that to build yay itself only x,y,z are used from base-devel

But please in an extra section, the installation instructions should be as brief and precise as possible, as not to confuse newcomers.

grandchild:
I'm not so sure about `--needed` after all. Random flags from readme-example lines often wind up copied along into the future because people don't understand why it's there (how could they, they're only just reading on how to get started with this "yay" thing). And yes, omitting `--needed` might re-install `gcc` here and there, but it's a small price to pay for having a really simple installation line, that only has the minimum flags. At least the latter is an appealing feature to me.

Morganamilo:
To be fair there's not really a downside to using `--needed` in any case really.

>Random flags from readme-example lines often wind up copied along into the future because people don't understand why it's there

You're probably right on this but if people don't understand how to use the package manager their distro uses I consider it their fault.

grandchild:
> You're probably right on this but if people don't understand how to use the package manager their distro uses I consider it their fault.

Fair point. And it's not an issue I wanted to press too much. Don't care enough either way :)

Jguer:
Closing due to inactivity

PR #1378, opened 2020-10-01
Translated the help display for 'yay'. And I modified the PO file.

FuRuYa7:
Fixed the 'cmd.go' file. In addition, the translated part of the help display was added to all PO files, and only 'en.po' and 'ja.po' were translated in the help display part.


cmd.go
```diff
@@ -20,124 +20,124 @@ import (
...
+	fmt.Println(gotext.Get("    -g --currentconfig    Print current yay configuration"))
+	fmt.Println(gotext.Get("    -s --stats            Display system package statistics"))
+	fmt.Println(gotext.Get("    -w --news             Print arch news"))
+	fmt.Println()
+	fmt.Println(gotext.Get("yay specific options:"))
+	fmt.Println(gotext.Get("    -c --clean            Remove unneeded dependencies"))
+	fmt.Println(gotext.Get("       --gendb            Generates development package DB used for updating"))
+	fmt.Println()
+	fmt.Println(gotext.Get("getpkgbuild specific options:"))
+	fmt.Println(gotext.Get("    -f --force            Force download for existing ABS packages"))
```
grandchild: Genuine question: Is this the way other projects translate their usage sections?

Why can't it remain one string, which, as a whole, should be translated?

FuRuYa7: Hello

I am using "yay" conveniently. Thanks.
I will explain the reason why it is divided into lines.

Creating a POT file in page-by-page display is incompatible and makes the "xgettext" tool unusable. For example, "syntax error" and "keyword unknown" occur frequently in "msginit" and "msgfmt". I think the advantage of using POT, PO, and MO files when translating is that you can use the "xgettext" tool.



## Operation example:

- I created a POT file using "xgotext". It is the operation after that.

```
$ cd ~/go/src/yay-help/out/

$ cat yay-hlp3.pot
msgid ""
msgstr ""
"Plural-Forms: nplurals=2; plural=(n != 1);\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=UTF-8\n"
"Content-Transfer-Encoding: 8bit\n"
"Language: \n"
"X-Generator: xgotext\n"

#: cmd.go:23
msgid `Usage:
    yay
    yay <operation> [...]
    yay <package(s)>

: (Omitted)

getpkgbuild specific options:
    -f --force            Force download for existing ABS packages`
msgstr ""
```



### Try using "msginit":

```
$ LANG=C msginit --locale=en --input=yay-hlp3.pot
out/yay-hlp3.pot:11:8: syntax error
out/yay-hlp3.pot:11: keyword "Usage" unknown
out/yay-hlp3.pot:12: keyword "yay" unknown
out/yay-hlp3.pot:13: keyword "yay" unknown

: (Omitted)

msginit: too many errors, aborting
```



### Try using "msgfmt":

```
$ cp yay-hlp3.pot yay-hlp3.po

$ LANG=C msgfmt -o yay-hlp3.mo yay-hlp3.po
yay-hlp3.po:11:8: syntax error
yay-hlp3.po:11: keyword "Usage" unknown
yay-hlp3.po:12: keyword "yay" unknown

: (Omitted)

msgfmt: too many errors, aborting
```

Jguer:
Hey @FuRuYa7 , thanks for the time spent on this, I know it's hard to get this working properly, 
I do have some changes I would like to request in order to get this merged. 

I'm trying to avoid placement information getting embedded in the translations

As examples:
- On operations

```
     yay {-F --files}       [options] [package(s)] 
```

Of this statement the translatable strings are `options` and `package(s)`

> `fmt.Println(gotext.Get("    yay {-F --files}       [options] [package(s)]")) ` has way to much context information, this would need to split using printf  (I don't mind helper functions to determine placement) 

- On flags

``` 
--sortby    <field>   Sort AUR results by a specific field during search 
```

Of this statement what can actually be translated is `field` and  `Sort AUR results by a specific field during search`

> Again printf statements and keeping track of the spacing for column alignment.

This will allow a lot of translations to be reused across concepts and if we migrate to a config system where we can keep the var declaration and help message together free of context.

Jguer (changes requested):
https://github.com/Jguer/yay/pull/1378#issuecomment-702671505

cmd.go
FuRuYa7: Hey @Jguer , thank you for your hard work.

I think it's okay if the placement information is embedded in the translation. If you change it too much, the source program will be difficult to see. Translation is only difficult at first. It is not updated every time. Instead, it is more efficient to focus on other things.

I'm editing a PO file directly in a text editor, so it's easier to translate if I can see the whole thing rather than just a few strings.

If you paste the "help display part" added to "en.po" into a PO file in another language, the "msgid" will be copied to the "msgstr", making it easier to understand the translation position. It might have been better to do so.

FuRuYa7:
<Fixed>

I fixed it because there was a request. The translation range has been narrowed down.

FuRuYa7:
(#1378) <Fixed> Translated the help display for 'yay'. And I modified the PO … 

It seems that sending a modified version of the source failed. Resend the PR from another clone. I think "diff" will also be easier to see.

PR #1384, opened 2020-10-08
<Fixed 2> Translated the help display for 'yay'. And I modified the P…

FuRuYa7:
Fixed an error in "golangci-lint"

Fixed the 'cmd.go' file. In addition, the translated part of the help display was added to all PO files, and only 'en.po' and 'ja.po' were translated in the help display part.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 
 	alpm "github.com/Jguer/go-alpm/v2"
 	"github.com/leonelquinteros/gotext"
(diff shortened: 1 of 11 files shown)


Jguer:
Hey @FuRuYa7,  I was thinking of using Printf formatted strings, if you're not accustomed with it I can take it from here and clean up the MR

FuRuYa7:
In "Help" of "yay", only fixed character strings are displayed without using variables. You don't have to use "Printf()" to add unnecessary "format parameters" to make it complicated.

#### For example:

```
	fmt.Print("(1). ")
	fmt.Print(gotext.Get("Usage"))
	fmt.Print(":\n\n")

	fmt.Printf("%v", "(2). ")
	fmt.Printf("%v", gotext.Get("Usage"))
	fmt.Printf("%v\n\n", ":")

	fmt.Print("END\n")
```

--> Using "Print ()" makes it easier to read, but is "Printf ()" better? Is there any reason?

FuRuYa7:
- When I tried again with "printf()" I was able to reduce the total number of lines. So I will try this.

When I send it again, the "diff" display becomes hard to see. Therefore, I will send it in another pull request.

grandchild:
@FuRuYa7 do you know about `git commit --amend` and `git push --force-with-lease`? With these commands you can change and push the same commit(s) over and over again. This is usually the way changes are applied in PRs, because it's going to be a very scattered conversation if you create a new pull request for every little change that is requested.

What I mean is basically option 2 from this guide: https://www.burntfen.com/2015-10-30/how-to-amend-a-commit-on-a-github-pull-request
(But instead of `-f`, or `--force`, you should *always* use **`--force-with-lease`** -- details in [the `git push` man page](https://www.git-scm.com/docs/git-push#Documentation/git-push.txt---force-with-leaseltrefnamegt).)

FuRuYa7:
I didn't know about those commands. Thank you for your advice.

FuRuYa7:
<Fixed 3> Translated the help display for 'yay'. And I modified the PO file.

I tried again with "printf()".

Fixed the 'cmd.go' file. In addition, the translated part of the help display was added to all PO files, and only 'en.po' and 'ja.po' were translated in the help display part.

FuRuYa7:
#### <Fixed 4> Translated the help display for 'yay'. And I modified the PO file.

With the addition of the help display, the order of PO files has become unnatural. I recreated the PO file with "xgotext". And corrected the translation of **"ja.po"** .

By the way, some items have been added to the PO file by other program updates.

#### Added translation items:

```
#: download.go:200
msgid "%s already exists. Use -f/--force to overwrite"
msgstr ""

#: clean.go:177
msgid "Unable to clean:"
msgstr ""
```

Jguer:
Please check https://github.com/Jguer/yay/blob/FuRuYa7-develop/cmd.go for how I think it should be, my only issue on that branch is that the alignment hasn't been corrected (the Print package functions has a similar code that would need adaptation)

FuRuYa7:
<Fixed 5> FuRuYa7-develop/cmd.go

I checked and fixed "https://github.com/Jguer/yay/blob/FuRuYa7-develop/cmd.go".

I also changed the PO file at the same time. In particular, the translation of **"ja.po"** was incorrect, so I corrected it.

FuRuYa7:
<Fixed 5> FuRuYa7-develop/cmd.go -- Fixed '<package (s)>'

Fixed "yay {-R --remove} [options]    <package (s)>".

Issue #1386, opened 2020-10-09
Downloading with self-signed certificate


Sohimaster:
program does not working at corps with self-signed certificates, add ssl verify option to avoid this

Sohimaster:
https://github.com/Jguer/yay/issues/137 link to this ticket, I use pacman

grandchild:
Hi there, I see you found the original ticket. So it seems there is really no need for this duplicate ticket then. The maintainers are clearly aware of the issue already, since they even commented on #137 too.

In general it's not much use having multiple tickets for the same issue, it only creates confusion.

One option you have to signal your need for a solution is to :+1: the original issue via the reaction button at the top right of the original issue message -> https://github.com/Jguer/yay/issues/137#issue-294622355.

I get that it can be frustrating seeing an issue that's giving you problems not resolved for a long time, but remember, the people who are writing this software are doing this in their spare time!

grandchild:
On the other hand, the other isse is closed, so maybe this issue does have merit...

Sohimaster:
@grandchild, the issue you are talking about is marked as bug however it's common network-paired software functionality. Furthermore, this issue is closed

Sohimaster:
that's why I created this issue.

Sohimaster:
I have no expertise in GoLang, but I can try to implement this option, if u can support me somehow if i need

grandchild:
I already started implementing it. The PR is linked above :)

Issue #137, opened 2018-02-06
Custom Certificate Authority


ibrokemypie:
Sometimes traffic must go through networks which require a custom ssl certificate authority as they resign all pages. How can yay be used in such an environment, as it currently complains such as the following
```
Get https://api.github.com/repos/jaagr/polybar/branches: x509: certificate signed by unknown authority

```

Jguer:
I would get guess golang uses the cacertificates of the system through a not very deep research I find this:
[https://golang.org/src/crypto/x509/root_linux.go](https://golang.org/src/crypto/x509/root_linux.go)

```
// Possible certificate files; stop after finding one.

var certFiles = []string{

	"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.

	"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6

	"/etc/ssl/ca-bundle.pem",                            // OpenSUSE

	"/etc/pki/tls/cacert.pem",                           // OpenELEC

	"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7

}
```

Guess you have to add the CA to the CA-trust of the system ( I've had this issue with pacman due to work proxies and it involved messing with the system)

nifisher:
If I'm not mistaken, that would set a universal acceptance of the unkown authority for every ssl request being made on my system as opposed to it being allowed in just this case, correct?

Why would I want to do that?   In my environment, I know who the unknown authority is (my employer's man-in-the-middle) and am accepting of the matter in this case, but would not necessarily want it in all cases on this system.  Telling yay to ignore the cert error but it still flagging for everything else would be a superior approach i feel.

grandchild:
@nifisher: But the authority is not unknown, is it? You do trust your employer's certificate. (Or at least your system is expected to.)

I found this though: https://forfuncsake.github.io/post/2017/08/trust-extra-ca-cert-in-go-app/ , which boils down to, I think:
```go
rootCAs, err := x509.SystemCertPool()
if err != nil || rootCAs == nil {
	rootCAs = x509.NewCertPool()
}
certs, err := ioutil.ReadFile(localCertFile)
rootCAs.AppendCertsFromPEM(certs)
```

That could do what the people above want...

(I'd still argue that if you have an employer that does stuff like SSL resigning and you install packages with it, then there's not much to gain by limiting other apps on your system, but :shrug:)

[...]

PR #1387, opened 2020-10-09
Add --mitmsslcert option

grandchild:
This is a draft for implementing a fix for jguer/yay#137 and jguer/yay#1386 ...

Two remarks:

* I'd appreciate feedback on code style and location. (Not sure where everything is supposed to go, this option should take effect pretty early in the code, so it's not buried very much, but maybe it's too central?)
* I have not tested this yet, I will probably do that on the weekend, using this approach: https://docs.mitmproxy.org/stable/howto-transparent-vms/ to simulate a corporate network HTTPS MITM.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -149,6 +149,10 @@ func handleCmd(cmdArgs *settings.Arguments, dbExecutor db.Executor) error {
 		sudoLoopBackground()
 	}
 
+	if config.MitmSSLCert != "" {
+		addMitmSSLCert(config.MitmSSLCert)
+	}
+
 	switch cmdArgs.Op {
 	case "V", "version":
 		handleVersion()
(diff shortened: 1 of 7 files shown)


grandchild:
Translations are obviously missing as well, but that's for later.

grandchild:
The name of the flag is also very much up for discussion.

Jguer:
@grandchild I can review it and I can even accept it but how will you make sure that `PKGBUILDs` will actually download https content if the cert is only in store for `yay`? 

As far as I see this fix moves the issues to 2 separate points:
- File downloading in makepkg
- Database refresh on pacman

Sohimaster:
@grandchild why should I provide sslCert file? Can I just disable mimt security check?

Sohimaster:
@grandchild @Jguer for example in Postman I have option SSL certificate verification (on/off)

grandchild:
@Sohimaster yes, you and @Jguer are both right, it doesn't make much sense this way.

It should be a mix of both "disable SSL verification" and editing `/etc/makepkg.conf` to add the necessary `curl` or `wget` options to `DLAGENTS`... I might look into this next week -- this week I probably won't have time anymore.

Issue #1399, opened 2020-10-26
When yay's path does not exist, config is not save


lwlf:
**Affected Version**
``` 
yay v10.1.0 - libalpm v12.0.2
```

**Describe the bug**
When the the path ```~/.config/yay/``` does not exist, use the command ```yay --save``` to save the configuration, which return <nil>.

**To Reproduce**

If the path is exist, delete it: 
```
rm -rf ~/.config/yay
```
Save any configure:
```
yay --save --editmenu
```

**Expected behavior**
Save the configuration and create the profile ```~/.config/yay/config.json``` and the path ```~/.config/yay/``` .

**Output**
```
$ yay --save --editmenu
<nil>
```

jugendhacker:
Same problem for me, creating the dir worked for me

Jguer:
Fixed on https://github.com/Jguer/yay/commit/497f94ca9f4c784917522cfedad764267c8e66d9

PR #1401, opened 2020-10-26
unsaved config when yay's config path does not exist

lwlf:
issue: [When yay's path does not exist, config is not save](https://github.com/Jguer/yay/issues/1399)
fix: unsaved config when yay's config path does not exist


lwlf:
I gave up.

PR #1398, opened 2020-10-26
Jguer/ /fix mflags

Jguer:
fixes #1397

PR #1412, opened 2020-11-07
Fill fields of CmdBuilder with options specified in commandline

### LINKED ISSUE #1409: makepkgconfig has no effect
#### Affected Version
yay v10.1.0.r9.gde7373d - libalpm v12.0.2

**Describe the bug**
`yay --makepkgconfig alternative.conf` uses standard `/etc/makepkg.conf` instead of provided alternative.
I specifically noticed this with different CFLAGS/CXXFLAGS used.
`makepkg --config alternative.conf` works as expected.

**To Reproduce**
create /etc/alternatemakepkg.conf with different CFLAGS/CXXFLAGS
yay --makepkgconf /etc/alternatemakepkg.conf
makepkg still uses CFLAGS/CXXFLAGS from /etc/makepkg.conf

**Expected behavior**
yay passes makepkgconfig onto makepkg.

kaz:
Command line options such like `--git` and `--makepkgconf` seem not to be applied since `exe.CmdBuilder` was introduced.
This is partially fixed by #1398, but rest of fields have not been passed to git/makepkg yet.

This PR fixes problems above.
Fixes #1409

### DIFF
--- a/pkg/settings/parser.go
+++ b/pkg/settings/parser.go
@@ -817,8 +817,14 @@ func (a *Arguments) ParseCommandLine(config *Configuration) error {
 	}
 
 	a.extractYayOptions(config)
-	config.Runtime.CmdBuilder.MakepkgFlags = strings.Fields(config.MFlags)
-	config.Runtime.CmdBuilder.GitFlags = strings.Fields(config.GitFlags)
+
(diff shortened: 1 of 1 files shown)


Jguer:
Ah, oops, my bad, thanks for the fix

Issue #1416, opened 2020-11-11
If the local language is of a type of hieroglyph, "yay -Si" format output is not aligned. Like, Chinese.


lwlf:
#### Affected Version

yay v10.1.0 - libalpm v12.0.2

**Describe the bug**
If user's local language is hieroglyphics, "yay -Si" format output is not aligned. Like, Chinese.
![Screenshot_20201111_173515](https://user-images.githubusercontent.com/35271500/98794841-7d177f00-2444-11eb-8012-4df8b3a11b12.png)

**To Reproduce**
1. Install zh_CN font(s) or other.
``` 
yay -S wqy-zenhei
```
2. Print package's info
```
env LANG=zh_CN.UTF-8 yay -Si zsh-git
```

**Expected behavior**
Like this:
![Screenshot_20201111_174415](https://user-images.githubusercontent.com/35271500/98795788-910fb080-2445-11eb-8a13-d4b9aeb70d9d.png)

**Output**
![Screenshot_20201111_174505](https://user-images.githubusercontent.com/35271500/98795835-a84e9e00-2445-11eb-8f35-727ccdf3bca4.png)

lwlf:
Some hieroglyphics also output not aligned, like Japanese, Korean, etc.

PR #1423, opened 2020-11-15
Fix: Aligned the string, when using 'yay -aSi'.

lwlf:
Fix: Aligned the string, when using 'yay -aSi'. 
For more information: #1416 
* Fixes the 'yay -aSi' output text alignment issue.
* I tested Chinese, English and Japanese, and they had no problem aligning.
* Where the maximum key value in Japanese exceeds the default length, the width of the maximum key value is calculated as the output length.

Note: Hopefully this fix can be done, and the problem of incorrect alignment is very troubling.


lwlf:
This is the test pictures: 

English: 
Screenshot_20201115_165702

Japanese: 
Screenshot_20201115_165749

Chinese:
Screenshot_20201115_165824

PR #1430, opened 2020-12-05
Add "--noconfirm" flag to usage output

lBeJIuk:
When solving one problem, I was surprised that the yay does not have a flag to not ask the user any questions. After looking at the code, I realised that this flag is simply undocumented.


Morganamilo:
Pacman flags are not documented. Otherwise out man page would be 4 times as large.

lBeJIuk:
Okay. Thank you for the answer!

grandchild:
Someone should close this PR then. @lBeJIuk

PR #1437, opened 2020-12-23
Print error if PATH is empty.

### LINKED ISSUE #881: yay fails silently if PATH is not set
#### Affected Version

`yay v9.1.0 - libalpm v11.0.2`

#### Issue

Running yay without a PATH environment variable makes it fail very silently. I'd expect it to print some message before failing.

#### Steps to reproduce

normal operation
```
$ yay -Qu
…show some packages…
$ echo $?
0
```

Buggy operation, no packages are listed:
```sh
$ PATH= /usr/bin/yay -Qu
$ echo $?
1
```

ghost:
This fixes #881.

Before:
```bash
$ PATH= /usr/bin/yay -Qu
$ echo $?
1
```

After:
```bash
$ PATH= ./yay -Qu
exec: "pacman-conf": executable file not found in $PATH
$ echo $?
1
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -30,7 +30,11 @@ func initAlpm(cmdArgs *settings.Arguments, pacmanConfigPath string) (*pacmanconf
 
 	pacmanConf, stderr, err := pacmanconf.PacmanConf("--config", pacmanConfigPath, "--root", root)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


Jguer:
Merged, thanks for your contribution

PR #1441, opened 2020-12-29
Add Swedish translation

riiga:
Added Swedish translation for Yay.


Jguer:
I guess we'll soon know if there's any issue need fixing, thanks for your contribution

PR #1459, opened 2021-02-13
fix(yay): mirror pacman return codes on run error

### LINKED ISSUE #1455: exit code 127 on yay -T, same as pacman -T
#1444 hins on how to install local packages, and #1351 details a little more on one challenge with exit code 1 if a dependency is not there.

please give same exist code on same parameter as pacman. here, for yay -T, give 127 instead of 1.

127 error code is defined here:
https://git.archlinux.org/pacman.git/tree/lib/libalpm/po/en_GB.po#n579

julixau:
makepkg expects pacman-alikes to return 127 (missing dependency)
when called with -T as argument. yay does not behave like this since
it overwrite the return value with 1 in every case.

This commit will test for ExitCode errors and return the called
applications error code instead of 1, when possible. This gets yay
closer to be used as pacman alternative for makepkg with

```
PACMAN=yay makepkg -s
```

It is however still not possible for yay to be used this way, since makepkg will call 
yay as root. This prevents yay from installing packages from aur (and rightfully so). 

fixes  Jguer/yay#1455

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@ package main // import "github.com/Jguer/yay"
 import (
 	"fmt"
 	"os"
+	"os/exec"
 
 	pacmanconf "github.com/Morganamilo/go-pacmanconf"
 	"github.com/leonelquinteros/gotext"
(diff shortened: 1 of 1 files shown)


julixau:
i suspect the test failure to be related to these lines at the start of logging

```
error: failed to initialize alpm library
(could not find or read directory: /var/lib/pacman/)
error: failed to initialize alpm library
(could not find or read directory: /var/lib/pacman/)
```

local testing passes, let me know if changes are required

Jguer:
Nice MR @julixau , It does seem to be related to the CI, I'll take a look at it and try the patch locally

Jguer (approved):

PR #1462, opened 2021-02-18
Loosen Package Dependencies in YAY

HoffmannJM:
Hi, I had some spare time during the lock down and have taken a deeper look at yay
with the intent to contribute and help you maintain it.
My primary goal is to improve the architecture of yay making it easier to maintain.
This is just the first step of my plan to tidy up yay.

Here is a rough sketch of the architectural changes I have in mind for that plan.

Layers Now (internal dependencies)

    multierror: -
    stringset: -
    text: -

    intrange: stringset
    exe: text

    vcs: exe, text

    settings: exe, vcs
    upgrade: text, vcs, intrange

    db: upgrade
    news: settings, text

    completion: db
    query: intrange, multierror, text, stringset, db, settings

    dep: db, text, stringset, query, settings

    pgp: dep, text

    main: *

Layers Goal (internal dependencies)

    db: -
    intrange: -
    multierror: -
    stringset: -
    text: -

    completion: db, text
    exe: text
    settings: text

    vcs: exe, text
    query: intrange, multierror, text, stringset, db, settings
    news: settings, text

    dep: query, text, db, stringset, settings
    upgrade: db, query, text, vcs, intrange

    pgp: dep, text

    main: *

This would reduce the number of layers from 9 to 6.
I don't know yet how every change would look in detail to accomplish this
or if I'd be able to reach that in the end.

I currently do not plan on adding anything just moving things around
loosening and inverting dependencies.

-------------------------

With this first change I hid all alpm functionality behind the pkg/db package,
reduced the interface required by the completion package,
inverted the dependency from db to upgrade and
made the query package reexport the rpc types.

What you will see as result is that many external imports have been replaced
by module local ones.

My next step would be to apply this to the main package and make the
pkg/text package enclose the gotext Get functionality.

I hope you find this useful and thanks for creating yay.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -60,7 +60,6 @@ linters:
     - gosimple
     - govet
     - ineffassign
-    - interfacer
     - lll
     - misspell
     - nakedret
@@ -85,7 +84,8 @@ linters:
     #- funlen
     #- dogsled
     # disabled for now
-    # - godox
+    #- godox
+    #- interfacer
 
 issues:
   # Excluding configuration per-path, per-linter, per-text and per-source
--- a/pkg/completion/completion.go
+++ b/pkg/completion/completion.go
@@ -15,8 +15,12 @@ import (
 	"github.com/Jguer/yay/v10/pkg/db"
 )
 
+type PkgSynchronizer interface {
+	SyncPackages(...string) []db.IPackage
+}
+
 // Show provides completion info for shells
-func Show(dbExecutor db.Executor, aurURL, completionPath string, interval int, force bool) error {
+func Show(dbExecutor PkgSynchronizer, aurURL, completionPath string, interval int, force bool) error {
 	err := Update(dbExecutor, aurURL, completionPath, interval, force)
 	if err != nil {
 		return err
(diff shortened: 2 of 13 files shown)


HoffmannJM:
I can also break this up more if you'd prefer that.

Jguer:
Hey @D1CED , this is looking really good. Let me get the CI fixed and I think I'll get this merged without much alteration

Jguer (approved):

HoffmannJM:
Hi, thanks for taking a look.

Take your time, no need to rush.

Jguer:
@D1CED you may want to take a look at the action output, the linter does find some valid warnings for the patch

HoffmannJM:
The first and the fourth lint issues are of stylistic nature. I have no opinion on this.

The second and third are caused by the linter strangely making a distinction between a method and a function
while the semantics should be identical.

    func (a *A) Func()
    func Func(a *A)

    a.Func()  // implicit referencing happens
    Func(&a) // linter complains here because of explicitly taking a reference while being semantically equivalent

Non the less the function `StylizedNameWithRepository` can be made taking its argument by value (its fully immutable).

grandchild:
Well that is to be expected of a linter. It is precisely concerned with coding style, which _must_ be semantically equivalent, or else it wouldn't be a style change.

HoffmannJM:
As you can see I adjusted to three of the four suggestions. I did not change the interface type in query as it would break symmetry with the other functions in this file.

Jguer:
Seems fine by me, as well now the CI bug should have been fixed upstream, I'm merging this, thanks @D1CED

PR #1484, opened 2021-04-03
List `-debug` packages missing from AUR separately (Split Debug Package Support)

Ferdi265:
This PR makes it so `-debug` packages created via the `(debug strip)` makepkg option combination are listed separately from regular packages in the "Missing AUR Packages" warning.

The rationale for this is, that if someone uses this option combination, every AUR package will have a `$pkgbase-debug` counterpart containing source files and debug symbols. These `$pkgbase-debug` packages obviously do not exist in the AUR, so are always listed in the "Missing AUR Packages" warning, making legitimately missing packages hard to spot.

Since there is no way (that I know of) to identify these debug packages other than by their name, this invariably will lead to false-positives (e.g. `python-flask-debug`, `dosbox-debug`, or `godot-debug`), which is why I opted to show them as a separate warning rather than dropping them entirely from the list.

As for the implementation, this small patch just separates the packages directly in the print function, which works well, but admittedly is a bit ugly.

I appreciate discussion on whether `yay` actually wants to handle this, and/or if `-debug` packages should get some way to identify them properly (this would require upstream changes to `makepkg` and possibly `pacman`). Split Debug Packages are currently not very well supported in general, and also not documented in many places. I have marked the PR as [RFC] until the design is clear.

(As an added note, yay grew support for installing these `-debug` packages in PR #1191, also by me)

### DIFF
--- a/pkg/query/aur_warnings.go
+++ b/pkg/query/aur_warnings.go
@@ -46,8 +46,8 @@ func (warnings *AURWarnings) Print() {
 }
 
 func filterDebugPkgs(names []string) (normal, debug []string) {
-	normal = make([]string, 0)
-	debug = make([]string, 0)
+	normal = make([]string, 0, len(names))
+	debug = make([]string, 0, len(names))
 
 	for _, name := range names {
 		if strings.HasSuffix(name, "-debug") {
--- a/pkg/query/aur_warnings.go
+++ b/pkg/query/aur_warnings.go
@@ -2,6 +2,7 @@ package query
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/leonelquinteros/gotext"
 
@@ -21,9 +22,16 @@ func NewWarnings() *AURWarnings {
 }
 
 func (warnings *AURWarnings) Print() {
-	if len(warnings.Missing) > 0 {
+	normalMissing, debugMissing := filterDebugPkgs(warnings.Missing)
+
+	if len(normalMissing) > 0 {
 		text.Warn(gotext.Get("Missing AUR Packages:"))
-		printRange(warnings.Missing)
+		printRange(normalMissing)
+	}
+
+	if len(debugMissing) > 0 {
+		text.Warn(gotext.Get("Missing AUR Debug Packages:"))
+		printRange(debugMissing)
 	}
 
 	if len(warnings.Orphans) > 0 {
@@ -35,6 +43,21 @@ func (warnings *AURWarnings) Print() {
 		text.Warn(gotext.Get("Flagged Out Of Date AUR Packages:"))
 		printRange(warnings.OutOfDate)
 	}
+}
+
+func filterDebugPkgs(names []string) (normal, debug []string) {
+	normal = make([]string, 0)
+	debug = make([]string, 0)
+
+	for _, name := range names {
+		if strings.HasSuffix(name, "-debug") {
+			debug = append(debug, name)
+		} else {
+			normal = append(normal, name)
+		}
+	}
+
+	return
 }
 
 func printRange(names []string) {


pkg/query/aur_warnings.go
```diff
@@ -37,6 +45,21 @@ func (warnings *AURWarnings) Print() {
 	}
 }
 
+func filterDebugPkgs(names []string) (normal, debug []string) {
```
Jguer, line 48:
```suggestion
func filterDebugPkgs(names []string) ([]string, []string) {
```

pkg/query/aur_warnings.go
```diff
@@ -37,6 +45,21 @@ func (warnings *AURWarnings) Print() {
...
+	normal = make([]string, 0)
+	debug = make([]string, 0)
```
Jguer:
```suggestion
	debug = make([]string, 0, len(names))
```

pkg/query/aur_warnings.go
```diff
@@ -37,6 +45,21 @@ func (warnings *AURWarnings) Print() {
 	}
 }
 
+func filterDebugPkgs(names []string) (normal, debug []string) {
+	normal = make([]string, 0)
```
Jguer:
```suggestion
	normal = make([]string, 0, len(names))
```

pkg/query/aur_warnings.go
```diff
@@ -37,6 +45,21 @@ func (warnings *AURWarnings) Print() {
...
+
+	for _, name := range names {
+		if strings.HasSuffix(name, "-debug") {
+			debug = append(debug, name)
+		} else {
+			normal = append(normal, name)
+		}
+	}
+
+	return
```
Jguer, line 60:
```suggestion
	return normal, debug
```

Jguer (changes requested):
I know it's less clear for the returns like this but it avoids reallocating the slice at the beginning after it's already been declared and it allocates the max capacity of those slices

Jguer:
> I appreciate discussion on whether yay actually wants to handle this, and/or if -debug packages should get some way to identify them properly (this would require upstream changes to makepkg and possibly pacman). Split Debug Packages are currently not very well supported in general, and also not documented in many places. I have marked the PR as [RFC] until the design is clear.

This ambiguity in them is the reason why I feel hackish solutions are fine for them :sweat_smile:

pkg/query/aur_warnings.go
```diff
@@ -21,9 +22,16 @@ func NewWarnings() *AURWarnings {
...
+	normalMissing, debugMissing := filterDebugPkgs(warnings.Missing)
+
+	if len(normalMissing) > 0 {
 		text.Warn(gotext.Get("Missing AUR Packages:"))
-		printRange(warnings.Missing)
+		printRange(normalMissing)
+	}
+
+	if len(debugMissing) > 0 {
+		text.Warn(gotext.Get("Missing AUR Debug Packages:"))
```
Jguer, line 33: You might have to regenerate the english .po file for this, if you need help with it just ping me and I'll add it after

Ferdi265: I have looked in the repo to see if there are simple instructions on how to regenerate the .po file, but didn't find a lot.
I'd appreciate some help with that, yes

Ferdi265:
Thanks for the review, I'll address the changes to the filter function in a commit shortly

pkg/query/aur_warnings.go
Ferdi265: addressed in c23def7

pkg/query/aur_warnings.go
Ferdi265: addressed in c23def7

pkg/query/aur_warnings.go
Ferdi265: addressed in c23def7

pkg/query/aur_warnings.go
Ferdi265: addressed in c23def7

Ferdi265:
The latest push should have addressed your change requests (minus the .po changes).
The lint check failed because of missing `diff` command in the container somehow, haven't seen that happen before.

Ferdi265:
> This ambiguity in them is the reason why I feel hackish solutions are fine for them :sweat_smile:

That is a valid way to do it IMO. Better deal with the common cases in a nice way and make it possible to deal with the odd ones out, but don't pessimize the common case for them.

Removing the [RFC] label then :)

Jguer:
Broken lint stage is on me, I'll have it fixed

Ferdi265:
sorry for the push clutter; I just wanted to check if the linting worked already.

Anyway, the branch is pretty much ready, except that I don't know how to regenerate the `.po` translation files (I presume there's a better way than to manually add the entry)

Jguer:
@Ferdi265 the CI should be working now as expected

Ferdi265:
Rebased; the new linter seems to like the named return style more, so I reverted it back to that.

(Link to lint result: https://github.com/Jguer/yay/runs/2334122378)

Jguer:
Let's take it for a spin, thanks for the contribution @Ferdi265

PR #1485, opened 2021-04-12
XDG compatibility

keeganjk:
This makes it so that someone with a different cache configuration than ~/.cache can have things cached in their custom directory.

### DIFF
--- a/pkg/settings/config.go
+++ b/pkg/settings/config.go
@@ -141,8 +141,8 @@ func (c *Configuration) String() string {
 func DefaultConfig() *Configuration {
 	return &Configuration{
 		AURURL:             "https://aur.archlinux.org",
-		BuildDir:           os.ExpandEnv("$HOME/.cache/yay"),
-		ABSDir:             os.ExpandEnv("$HOME/.cache/yay/abs"),
+		BuildDir:           os.ExpandEnv("$XDG_CACHE_HOME/yay"),
+		ABSDir:             os.ExpandEnv("$XDG_CACHE_HOME/yay/abs"),
 		CleanAfter:         false,
 		Editor:             "",
 		EditorFlags:        "",


Jguer (changes requested):
If `$XDG_CACHE_HOME` is not set, build dir and ABSdir seem to get set to root `/yay` and `/yay/abs`

keeganjk:
I was thinking of possibly also adding a part that would set `$XDG_CACHE_HOME` to `~/.cache` if it was not set.

tengwar:
> part that would set `$XDG_CACHE_HOME` to `~/.cache` if it was not set

According to the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables), if `$XDG_CACHE_HOME` is not set (or empty), a default value of `$HOME/.cache` should be used. I believe you meant just that, but it kind of sounds like you want to set the envvar instead of using a default, so I'm clarifying just in case.

keeganjk:
> > part that would set `$XDG_CACHE_HOME` to `~/.cache` if it was not set
> 
> According to the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables), if `$XDG_CACHE_HOME` is not set (or empty), a default value of `$HOME/.cache` should be used. I believe you meant just that, but it kind of sounds like you want to set the envvar instead of using a default, so I'm clarifying just in case.

Yes, if `$XDG_CACHE_HOME` is not set, it should use the default, `~/.cache`.

Jguer:
@keeganjk are you still working on this?

keeganjk:
> @keeganjk are you still working on this?

No, I am not currently working on this.

Jguer:
Closing as this is already implemented by `NewConfig` , `getCacheHome`

PR #1500, opened 2021-05-06
Sync with go-alpm master

x1b6e6:


Jguer:
@x1b6e6 linter seems to not like your commit

As well if you could rebase you'll get `pacman-git` tests

x1b6e6:
> linter seems to not like your commit

Yes. This caused by renamed function (fixed in Jguer/go-alpm#47). My bad.

x1b6e6:
@Jguer rebased. Test failing because library `Morganamilo/go-pacmanconf` returning incorrect conf. Maybe this caused by [removing `TotalDownload` option from pacman](https://git.archlinux.org/pacman.git/commit/?id=41f9c50abf08e71aa5fd03568934ab80abc7715c).

Morganamilo:
go-pacmanconf uses pacman-conf behind the scenes. Because total download is removed from pacman-git, pacman-conf will not include it in the generated config even though go-pacmanconf has support for it.

My advice is to just have it set to false in your tests.

Morganamilo:
> @Jguer rebased. Test failing because library `Morganamilo/go-pacmanconf` returning incorrect conf. Maybe this caused by [removing `TotalDownload` option from pacman](https://git.archlinux.org/pacman.git/commit/?id=41f9c50abf08e71aa5fd03568934ab80abc7715c).


Also I wonder who it was that made that commit hmmmmm....

Jguer:
All good, let's merge this :+1:

Issue #1448, opened 2021-01-20
install dependencies before donwloading sources


LuckyTurtleDev:
### Is your feature request related to a problem? Please describe.
If a dependency needed by an dlagent the installation failed. Because the yay only ~check~ install dependency after downloading the source.

### Describe the solution you'd like
Yay should check and install dependency after cloning the AUR repo and before downloading the other source to satisfy also the dependency for the dlagents.

### Describe alternatives you've considered
Installing dependency manual.

### Additional context
DLAGENTS allow you fetch source from additional protocols.
As example this is used by many gog packages.

Morganamilo:
Yay resolves dependencies right at the start, before downloading sources and before cloning the aur package.

LuckyTurtleDev:
> Yay resolves dependencies right at the start, before downloading sources and before cloning the aur package.

If I try to install `gog-ftl` yay recognize the `lgogodownloader` dependency. But it try to download the sources of  `gog-ftl` before compiling/installing `lgogodownloader`.

Morganamilo:
I guess yay should not download all sources then start the building.

LuckyTurtleDev:
> I guess yay should not download all sources then start the building.

This should be a better behavior, which should fix the problem.

x1b6e6:
@Lukas1818 I tried implement needed behavior with new flag `--one-by-one`

[demo on asciinema](https://asciinema.org/a/nDQccDGAfMnbOQO1umCNnxIyz)
You can see code at [here](https://github.com/x1b6e6/yay/tree/feat-one-by-one).

This behavior is what you want? If yes i will create PR.

x1b6e6:
> I guess yay should not download all sources then start the building.

I think this made for possibility disconnect from internet after download all sources and building packages offline. 
This useful for systems what work in offline and connecting to internet only for updates (i haven't more ideas). After approving #1505 this behavior will be more comfortable.

LuckyTurtleDev:
> @Lukas1818 I tried implement needed behavior with new flag `--one-by-one`
> 
> [demo on asciinema](https://asciinema.org/a/nDQccDGAfMnbOQO1umCNnxIyz)
> You can see code at [here](https://github.com/x1b6e6/yay/tree/feat-one-by-one).
> 
> This behavior is what you want? If yes i will create PR.

yes, thanks

But I think this may should the default behavior, because many package hit this problem, as example every -git package.
[...]

PR #1505, opened 2021-05-14
Fix downloadonly flag for AUR packages

### LINKED ISSUE #1373: build without install - yay -Sw does not work
**Is your feature request related to a problem? Please describe.**
I need to download a package with it's dependencies without installing them.
For example downloading vmware-workstation with its dependency vmware-keymap

**Describe the solution you'd like**
yay -Sw vmware-workstation which will download and build vmware-workstation and vmware-keymap packages.

x1b6e6:
Fix #1373.

<details>
<summary>demo</summary>

[demo](https://asciinema.org/a/MSzgzX3DwGZBk7bMeU0R6KJhN?autoplay=1)

</details>

### DIFF
--- a/install.go
+++ b/install.go
@@ -373,13 +373,8 @@ func install(cmdArgs *settings.Arguments, dbExecutor db.Executor, ignoreProvider
 		_ = completion.Update(dbExecutor, config.AURURL, config.Runtime.CompletionPath, config.CompletionInterval, false)
 	}()
 
-	err = downloadPkgbuildsSources(do.Aur, incompatible)
-	if err != nil {
-		return err
-	}
-
 	if downloadOnly {
-		return nil
+		return downloadPkgbuildsSources(do.Aur, incompatible)
 	}
 
 	err = buildInstallPkgbuilds(cmdArgs, dbExecutor, dp, do, srcinfos, incompatible, conflicts, noDeps, noCheck)
--- a/install.go
+++ b/install.go
@@ -70,6 +70,7 @@ func install(cmdArgs *settings.Arguments, dbExecutor db.Executor, ignoreProvider
 	noCheck := strings.Contains(config.MFlags, "--nocheck")
 	sysupgradeArg := cmdArgs.ExistsArg("u", "sysupgrade")
 	refreshArg := cmdArgs.ExistsArg("y", "refresh")
+	downloadOnly := cmdArgs.ExistsArg("w", "downloadonly")
 	warnings := query.NewWarnings()
 
 	if noDeps {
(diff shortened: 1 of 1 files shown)


Morganamilo:
It looks like you're just not installing the final package, Meaning all the deps still get installed. This is why I've never attempted to implemented for AUR packages, there does not seem to be a sane way to handle it.

x1b6e6:
Look again. This downloading all dependencies from `pacman` and downloading sources of all dependencies in AUR but don't installing any of that.

`dkms` not installed. look at pacman output.
`vmware-keymap` not installed. pacman not called.
`vmware-workstation` not installed. I show it at the end.

All sources downloaded and checked by checksum.

Morganamilo:
Well shoulda put that in the commit message :P

If you're downloading sources though doesn't that mean we need the deps to be installed? e.g. svn packages.

Plus I don't really see a use to this. If the issue is -Sw going all funky we could just have yay error out saying it doesnt work for aur packages.

x1b6e6:
> we could just have yay error out saying it doesnt work for aur packages.

Why if this working in 90% situations? I understand what this will not work with `svn`, `git` or something else (but if it provided as dependency and it's exists in pacman repo then this [will work too](https://github.com/Jguer/yay/issues/1448#issuecomment-846024638)).

Morganamilo:
> Why if this working in 90% situations? 

Event if it works, is it actually useful to anybody? What does the user expect it to do?

There's three ways -Sw could work:

1. Download the pkgbuilds into cache
2. Download the pkgbuilds and sources into cache
2. Build the packages but don't install

1 Is pretty harmless and works in 100% of situations but would any one ever use it?
2 Won't work because deps are needed to download sources
3 Is kind of insane but it's what pacaur does so there's some prior art there

I don't see a clean way to do implement it so I feel it shouldn't be implemented.

> but if it provided as dependency and it's exists in pacman repo then this will work too

But the whole point is to not install anything.

x1b6e6:
> but if it provided as dependency and it's exists in pacman repo then this will work too

Sorry doesn't work. Because it will be downloaded but not installed.

x1b6e6:
> 1 Download the pkgbuilds into cache

Does it really useful downloading only pkgbuilds?

> 3 Build the packages but don't install

But flag named `--downloadonly`. I think this feature should have another name for flag. Maybe `--buildonly`? I think this flag should be only long for prevent conflicts with pacman flags in the future. I can try implement this.

> 2 Won't work because deps are needed to download sources

At now bug only if all of this points are met:

- user using '-w' flag
- user installing package what require specified downloader
- this downloader not installed

I have idea with using hardcoded list of downloaders but I think it's too wild.

Jguer:
@x1b6e6 seems to be working fine, sorry for slow review on this, I'll still merge it this cycle and try to refactor it a bit so `install` doesn't keep growing and growing

install.go
```diff
@@ -354,9 +373,8 @@ func install(cmdArgs *settings.Arguments, dbExecutor db.Executor, ignoreProvider
 		_ = completion.Update(dbExecutor, config.AURURL, config.Runtime.CompletionPath, config.CompletionInterval, false)
 	}()
 
-	err = downloadPkgbuildsSources(do.Aur, incompatible)
-	if err != nil {
-		return err
+	if downloadOnly {
```
Jguer, line 376: seems that this will stop `downloadPkgbuildsSources` from running when it's in normal mode

x1b6e6: Yes, this for #1448.

PR #1533, opened 2021-06-12
Add Korean translation

adenosie:


Jguer:
Seems to be fine, missing a small tweak of the makefile but I'll add that myself, 감사합니다

PR #1535, opened 2021-06-22
Add go to list of dependencies in install instructions

evanfarrar:
I just installed arch onto a new machine today; I wanted yay to install my current editor of choice and noticed that I needed `go` to run makepkg.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ building with makepkg:
 Before you begin, make sure you have the `base-devel` package group installed.
 
 ```sh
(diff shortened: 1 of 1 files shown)


Jguer:
Installing this will make `go` and explicit install and not recognized as a orphan dependency by `yay -Yc` after. 

This is why we use the `makepkg -si`, `-s` being `--syncdeps` to install missing dependencies with the correct type.

If you don't use go for development you can safely remove it

PR #1534, opened 2021-06-13
Don't use GOFLAGS for tags

OneOfOne:
go gets confused when it has `"-tags six"` in the go env `GOFLAGS`, so move this to `EXTRA_GOFLAGS`.

Fixes #1525 .

PR #1537, opened 2021-06-27
Makefile: GOFLAGS -> FLAGS since it's a reserved kw

OneOfOne:
A proper rewrite of #1534.


OneOfOne:
@Jguer I just renamed the flag, GOFLAGS is used by go and for some reason Makefile breaks it.

Jguer:
Thanks for the contribution @OneOfOne , I'm merging this and then tweaking it slightly for release

Issue #1541, opened 2021-07-02
Unable to download PKGBUILD from ABS


zhmars:
### Affected Version

`yay v10.2.3 - libalpm v13.0.0`

### Describe the bug

Yay fail to download PKGBUILD from ABS due to `git.archlinux.org` retired.

Ref: https://lists.archlinux.org/pipermail/arch-devops/2021-July/000527.html

### Reproduction Steps
1. `yay -G wget`

### Expected behavior

 Downloads PKGBUILD from ABS

### Output

```sh
:: Querying AUR...
failed to get pkgbuild: wget: error fetching wget: fatal: unable to access 'https://git.archlinux.org/svntogit/packages.git/': Could not resolve host: git.archlinux.org
```

edacval:
https://lists.archlinux.org/pipermail/arch-dev-public/2021-June/030466.html

omgold:
The location of the sources repo has changed. They can now be found at:

https://github.com/archlinux/svntogit-packages.git
https://github.com/archlinux/svntogit-community.git

zhmars:
The error has been fixed in v10.3.0. Thanks!

I just found the message `Querying AUR...` seems wrong here since the package was downloaded from ABS. Am I supposed to create a new issue for that?

k0ste:
Still not works

```python
[k0ste@k0ste tmp]$ yay -V
yay v10.3.0 - libalpm v13.0.0
[k0ste@k0ste tmp]$ yay -G netdata
:: Querying AUR...
failed to get pkgbuild: netdata: error fetching netdata: fatal: unable to update url base from redirection:
  asked for: https://git.archlinux.org/svntogit/community.git/info/refs?service=git-upload-pack
   redirect: https://gitlab.archlinux.org/explore/groups
```

yochananmarqos:
> The error has been fixed in v10.3.0.

Oh really? 🤔

```
❯ yay -G chromium
:: Querying AUR...
failed to get pkgbuild: chromium: error fetching chromium: fatal: unable to update url base from redirection:
  asked for: https://git.archlinux.org/svntogit/packages.git/info/refs?service=git-upload-pack
   redirect: https://gitlab.archlinux.org/explore/groups
```

zhmars:
I'm not sure why you still get the error. But it works for me at least.

```sh
$ yay -V
yay v10.3.0 - libalpm v13.0.0
$ yay -G netdata
:: Querying AUR...
[...]

PR #1546, opened 2021-07-09
fixing upstream urls from git.archlinux.org to github

omgold:
The ABS repo is no longer found at git.archlinux.org but (to my knowledge) only on github. See #1541. This patches makes `yay -G` work again for me.


Morganamilo:
This could also be a sign to depend on `asp` instead of rolling out own thing.

Jguer:
@omgold I confirm the fix does work, thanks for fixing this

PR #1549, opened 2021-07-10
Support download only mode for AUR packages

idanyadgar:
This fixes issue #1373.


Jguer:
Hey @idanyadgar , would you be more inclined into reviewing https://github.com/Jguer/yay/pull/1505 (or presenting a compelling case why your implementation would be better)?

PR #1559, opened 2021-07-27
fix: skip --assume-installed packages

### LINKED ISSUE #1137: unable to skip dependency with --assume-installed / --nodeps

irgendwr:
This fixes #1137 by skipping yay's dependency checks for packages provided as parameter of `--assume-installed`.

[pacman(8) documentation](https://www.archlinux.org/pacman/pacman.8.html#_transaction_options_apply_to_em_s_em_em_r_em_and_em_u_em):

> --assume-installed <package=version>
> Add a virtual package "package" with version "version" to the transaction to satisfy dependencies. This allows to disable specific dependency checks without affecting all dependency checks.

Not sure if there are other checks that need to be skipped. Let me know if I missed something.

One thing to note is that one still needs to edit [PKGBUILD files](https://wiki.archlinux.org/title/PKGBUILD#Dependencies) to remove/skip dependencies afaik.

### DIFF
--- a/install.go
+++ b/install.go
@@ -68,6 +68,7 @@ func install(cmdArgs *settings.Arguments, dbExecutor db.Executor, ignoreProvider
 	var srcinfos map[string]*gosrc.Srcinfo
 	noDeps := cmdArgs.ExistsDouble("d", "nodeps")
 	noCheck := strings.Contains(config.MFlags, "--nocheck")
+	assumeInstalled := cmdArgs.GetArgs("assume-installed")
 	sysupgradeArg := cmdArgs.ExistsArg("u", "sysupgrade")
 	refreshArg := cmdArgs.ExistsArg("y", "refresh")
 	warnings := query.NewWarnings()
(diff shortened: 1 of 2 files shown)


Jguer (approved):

Jguer:
Seems to be ok @irgendwr , I have to try it out a bit but I think we can merge this :+1: , thanks for the work on this, it's the place of yay most needing a refactor though

Jguer:
Hey @irgendwr I've merged your patch on another branch https://github.com/Jguer/yay/commit/7dda0fa2f84cd1e32568ec583c3dca4502f0f8df to also include a small fix to take into account depstrings and use the builtin separator. Thanks for the patch :+1:

PR #1562, opened 2021-07-30
Update yay.8 to supress troff warning

peterwu:
When running `man yay > /dev/null`, the system reports 
```
troff: <standard input>:1: name expected (got '\ '): treated as missing
```
The culprit appears to be the first line in the yay.8 file. This PR fixes this problem.


Jguer (approved):

Jguer:
:+1: thanks @peterwu , I can confirm the error

Issue #915, opened 2019-03-20
Parallelize downloading multiple packages sources


kangalioo:
Would it be possible to add support for downloading and building in parallel? I had in mind that when installing multiple packages, yay could already start building the first package while the second package is downloading. Then the second package is built while the third is downloaded and so on.

I think that would be a nice feature to have because downloading and building can be parallelized so well. It would make updating the system with `yay -Syu` up to twice as fast.

Jguer:
Parallel building is not a great advantage because then you have competition for cycles (and packages with dependencies will have to be treated). Concurrent download is something that has been thought about and experimented with (it messes a bit with the outputs though).

kangalioo:
True, parallel building wouldn't be beneficial, but that's not what I meant. I meant having the downloading process running in parallel to the building process. Those process need completely different parts of the system, so parallelizing those two would provide a big speed improvement.

Morganamilo:
I can see it working but I can't see any way to implement it while also keeping any sort of readable output.

kangalioo:
It could be a command-line option that's turned off by default, so yay has readable output by default. When enabled, yay could display a status message in between the compilation output giving information about the downloads, maybe something like this:
```
...
[ 41%] Building CXX object ...
[yay] Download of ___.tar.gz finished, starting download ___.tar.gz...
[ 43%] Building CXX object ...
[ 45%] Building CXX object ...
...
```

BrendanBall:
Seen as parallelizing builds isn't an option, trying to download and build in parallel seems like overkill. I would go with just parallelizing multiple downloads at the same time.

dsdante:
@Jguer why not pass -j to make and let it do parallelization?

@BrendanBall what's the point in dividing the bandwidth among packages?

Jguer:
@dsdante That's the job of `/etc/makepkg.conf` not yay, you already have the parallelization of a single build and that is effective (everything C++ for example ...). Parallelization of multiple builds would be like trying to shove 20 clowns by a door at a single time, probably slower than having them line up. 

Dividing the bandwidth among packages could be beneficial for multiple AUR builds as the sources are different and there could be slow servers in some of them. 
On repo packages where all the packages have the same source your connection will typically max out before you're able to max out a single repository, so less effective.

dsdante:
@Jguer thanks for the makepkg tip!

I think the easiest way to optimize it is to make a pipeline:
[...]

PR #1573, opened 2021-08-13
Concurrent PKGBUILD source download

Jguer:
Implements #915

### DIFF
--- a/aur_source.go
+++ b/aur_source.go
@@ -1,0 +1,121 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"runtime"
+	"sync"
+
+	"github.com/leonelquinteros/gotext"
+
+	"github.com/Jguer/yay/v10/pkg/dep"
+	"github.com/Jguer/yay/v10/pkg/multierror"
+	"github.com/Jguer/yay/v10/pkg/settings/exe"
+	"github.com/Jguer/yay/v10/pkg/stringset"
+	"github.com/Jguer/yay/v10/pkg/text"
+)
+
+type ErrDownloadSource struct {
+	inner   error
+	pkgName string
+	errOut  string
+}
+
+func (e ErrDownloadSource) Error() string {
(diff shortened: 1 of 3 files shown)


Morganamilo:
Does this ensure deps are installed first?

Jguer:
> Does this ensure deps are installed first?

This only changes the way `makepkg --verifysource -Ccf` is run, not the install logic itself so it should be ok for now

Morganamilo:
No, deps are needed before sources are downloaded. Git is a good example, custom dlanglets also pop up sometimes.

Jguer:
> No, deps are needed before sources are downloaded. Git is a good example, custom dlanglets also pop up sometimes.

This patch only replaces an existing for cycle with no return for a concurrent for cycle, unless the `bases []dep.Base` order is important I see no issue with it even in testing. Is `bases` order sensitive?

Morganamilo:
Ah I remember now. Yay already does the wrong thing. Now it just does the wrong thing in parallel :P

See https://github.com/Jguer/yay/issues/1448

SampsonCrowley:
Don't be such a sourpuss. this is a big performance improvement for the current process. *when* you submit a PR for your suggested change to the build process you can adapt it to be able to download deps first then main packages. Concurrent source downloading is a feature that the majority have been asking for and this is compatible with how things are already handled. Your issues are with edge cases and custom downloaders

Morganamilo:
Anyway the proper solution is to queue only if deps are satisfied

so for packages to install -> pick out ones with all deps satisfied -> push to download queue

if download queue is empty then build and install packages in queue.

This is similar to how batch install works so the core logic should already be there.

Jguer:
> Ah I remember now. Yay already does the wrong thing. Now it just does the wrong thing in parallel :P
> 
> See #1448

:) Exactly, the block is tested and will just need to be moved once the "install" is replacing with a pulling system instead of a pushing system

PR #1608, opened 2021-10-04
Fix caps in search leading to no results

Morganamilo:


Jguer (approved):

Jguer:
Seems fine, merging in

PR #1642, opened 2021-11-23
Prefer VISUAL over EDITOR for editmenu

godille:
VISUAL should take precedence since EDITOR is intended for line editors like ed or ex.


Jguer:
Makes sense @godille , thanks for the contribution

PR #1649, opened 2021-12-02
Add config option to disable package installation

aequabit:
I was surprised this didn't already exist, so I decided to add it myself. I never actually looked into the yay source and haven't written a single like of Go ever, so forgive me if I have no clue what I'm doing.

The only sync flag that's allowed is `-s`, because I still like to use it for searching instead of visiting the AUR website.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -318,6 +318,12 @@ func handleYogurt(ctx context.Context, cmdArgs *parser.Arguments, dbExecutor db.
 func handleSync(ctx context.Context, cmdArgs *parser.Arguments, dbExecutor db.Executor) error {
 	targets := cmdArgs.Targets
 
+	if config.DisableInstall {
+		if !cmdArgs.ExistsArg("s", "search") {
+			return fmt.Errorf(gotext.Get("installing packages is disabled"))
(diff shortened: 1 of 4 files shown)


Morganamilo:
I don't really get what the point of this is.

aequabit:
I want it to tell me "No!" when I try it, especially after I got used to it in the beginning but switched to installing AUR packages manually. That's pretty much it.

Morganamilo:
I don't really think there's any reason yay should include this. Yay prompts you before installing anyway so you shouldn't be able to accidentally install.

aequabit:
Well, that's up to you. I'm happy to compile it myself, just thought it might be nice to have for others.

Jguer:
@aequabit could you check out https://github.com/Jguer/aur and its included cmd? if it's only to search AUR packages it might be better, if you have suggestions/contributions on how to make it better and relevant it might even be quality enough to package/publish it

Jguer:
Closing as outside the scope of yay, some contributions for `https://github.com/Jguer/aur/cmd/aur-cli` would be great though

PR #1651, opened 2021-12-08
Update ru_RU.po to match pacman translation

JohnTheCoolingFan:
Changed "Optional Deps" and "Provides" to match with pacman translation.


Jguer:
Hey @JohnTheCoolingFan , I'm merging this but I'm not sure if it will play well with the transifex integration :sweat_smile: , I hope it does, thanks for the contribution

PR #1654, opened 2021-12-15
add --quiet to systemd-run

gdamjan:
this will avoid the `Press ^] three times within 1s to disconnect TTY.` message.

it's mentioned in the systemd-run man page too:
```
--quiet, -q
    Suppresses additional informational output while running. This is particularly useful in combination with --pty when it will suppress the
    initial message explaining how to terminate the TTY connection.
```

### DIFF
--- a/pkg/download/abs_test.go
+++ b/pkg/download/abs_test.go
@@ -206,7 +206,7 @@ func TestABSPKGBUILDRepo(t *testing.T) {
 		if path, _ := exec.LookPath(ld); path != "" {
 			ld = path
 		}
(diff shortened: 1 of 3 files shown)


Jguer:
Good idea @gdamjan , I think you only need to fix a test or two and it's good to merge

gdamjan:
not sure why the tests fail again, `make test` locally passes. Any ideas?

grandchild:
Looks to me like you also need to change the `TestABSPKGBUILDRepoExistsPerms` test.

Jguer (approved):

PR #1710, opened 2022-02-11
`go` package is not in the `base-devel` package group

dkuzmenko-sunpower:
A minor README update.
I tried to build `yay` on a fresh system from source using this README information and got this:
```
==> Missing dependencies:
  -> go>=1.17
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,10 @@ If you are migrating from another AUR helper, you can simply install Yay with th
 The initial installation of Yay can be done by cloning the PKGBUILD and
 building with makepkg:
 
-Before you begin, make sure you have the `base-devel` package group installed.
+Before you begin, make sure you have the `base-devel` package group and the `go` package installed.
 
 ```sh
-pacman -S --needed git base-devel
+pacman -S --needed git base-devel go
 git clone https://aur.archlinux.org/yay.git
 cd yay
 makepkg -si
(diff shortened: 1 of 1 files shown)


Jguer:
Hey @dkuzmenko-sunpower , did you run `makepkg -si`, the -s option normally installs missing dependencies from the repositories

README.md
```diff
@@ -33,10 +33,10 @@ If you are migrating from another AUR helper, you can simply install Yay with th
 The initial installation of Yay can be done by cloning the PKGBUILD and
 building with makepkg:
 
-Before you begin, make sure you have the `base-devel` package group installed.
+Before you begin, make sure you have the `base-devel` package group and the `go` package installed.
 
 ```sh
-pacman -S --needed git base-devel
+pacman -S --needed git base-devel go
```
Jguer:
```suggestion
pacman -S --needed --asdeps git base-devel go
```

Both `go` and `git` should be installed with `--asdeps` actually.

`base-devel` should be kept as explicit I think (at least I think it's better) but for the purpose of installing yay they are indeed make deps

Jguer (changes requested):

README.md
dkuzmenko-sunpower: Updated as suggested

dkuzmenko-sunpower: Updated as suggested

dkuzmenko-sunpower:
> Hey @dkuzmenko-sunpower , did you run `makepkg -si`, the -s option normally installs missing dependencies from the repositories

Oh really? I probably missed the `-s` param. Closing this.

PR #1716, opened 2022-02-18
fix(aur): -Ss fails on multi arg if no results.

### LINKED ISSUE #1674: -Ss option with two word argument fails with `no query was executed`
yay v11.1.0.r4.g09695c6 - libalpm v13.0.1:
```sh
$ yay -Ss cargo watch
 -> no query was executed
```
```sh
$ pacman -Ss cargo watch
community/cargo-watch 8.1.2-1
    Utility for Cargo to compile projects when sources change
```

ghost:
This fixes #1674.

If search command included multiple packages and no results were found
in AUR, response from `query.queryAUR` was returned
as nil slice instead of non-nil empty slice and as nil slice is treated
during subsequent validation in `query.SourceQueryBuilder.Results` as
meaning that searching in AUR was not executed, command was failing.
Now empty non-nil slice is returned if results were not found.

### DIFF
--- a/pkg/query/source.go
+++ b/pkg/query/source.go
@@ -188,10 +188,7 @@ func queryAUR(ctx context.Context, aurClient *aur.Client, pkgS []string, searchB
 		return r, err
 	}
 
-	var (
-		aq aurQuery
-		n  int
-	)
+	aq := make(aurQuery, 0, len(r))
 
 	for i := range r {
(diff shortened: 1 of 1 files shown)


Jguer:
Seems to be fine, need to see what's wrong with those pacman-git tests. Thanks a lot :smile:

Issue #1667, opened 2022-01-06
--editmenu option keeps you from starting the installation, it will say that the user aborted it


41zu:
### Affected Version

yay --version -> yay v11.1.0 - libalpm v13.0.1

### Describe the bug

--editmenu option keeps you from starting the installation, it will say that the user aborted it

### Reproduction Steps

1. yay -S linux-amd --editmenu --sudoloop
2. Confirm cleaning build dir
3. Show no diffs
4. Confirm editing all PKGBUILDs
5. Confirm to continue the installation
6. Confirm dialog says the user aborted it

### Expected behavior

with the previous version it continued the installation

### Output

:: Prüfe auf mögliche Konflikte...
:: Prüfe auf mögliche interne Konflikte...
[Repo Make:3]  xmlto-0.0.28-4  docbook-xml-4.5-9  docbook-xsl-1.79.2-7
[Aur:1]  linux-amd-5.15.v.13-1

==> Abhängigkeiten nach der Installation entfernen? [J/N] J
  1 linux-amd                       (Installiert)(Build-Dateien sind vorhanden)
==> Pakete für neuen Build?
==> [N] Keine [A]lle [Ab]brechen [I]nstalliert [No]nicht installiert oder (1 2 3, 1-3, ^4)
==> A
:: Lösche (1/1): /home/dominik/.cache/yay/linux-amd
:: (1/1) PKGBUILD heruntergeladen: linux-amd
  1 linux-amd                       (Installiert)(Build-Dateien sind vorhanden)
==> Diffs zeigen?
==> [N] Keine [A]lle [Ab]brechen [I]nstalliert [No]nicht installiert oder (1 2 3, 1-3, ^4)
==> N
:: (1/1) SRCINFO geparst: linux-amd
  1 linux-amd                       (Installiert)(Build-Dateien sind vorhanden)
==> Welche PKGBUILDs editieren?
==> [N] Keine [A]lle [Ab]brechen [I]nstalliert [No]nicht installiert oder (1 2 3, 1-3, ^4)
==> A

==> Mit der Installation fortfahren? [J/N] J
 -> Abbruch durch den Benutzer

My config output with yay -Pg:
{
	"aururl": "https://aur.archlinux.org",
	"buildDir": "/home/dominik/.cache/yay",
	"editor": "",
	"editorflags": "",
	"makepkgbin": "makepkg",
	"makepkgconf": "",
	"pacmanbin": "pacman",
[...]

PR #1735, opened 2022-03-27
Fix ContinueTask user prompt for locales with uppercase yes

### LINKED ISSUE #1717: Answering any Question isnt possible anymore -> everytime abort
Hello,

since few Releases, there is a Problem where Answering a Question from yay isnt possible (expect to enter nothing, only press enter).

For exception:

Import PGP? Y/n > y > abort
Import PGP? Y/n > n > abort
Import PGP? Y/n > *just hit enter* > imports

or

Edit PKGBUILD? Y/n > y > abort
Edit PKGBUILD? Y/n > n > abort
Import PGP? Y/n > *just hit enter* > continues installation 

and so on. 

Every Question, only enter is working. Else yay aborts. Except at the Point Pacman gets started to install packages, there everything works.

Drumstickx:
This pull request fixes the issue outlined in #1667. May also fix #1717

### DIFF
--- a/pkg/text/text.go
+++ b/pkg/text/text.go
@@ -74,7 +74,5 @@ func ContinueTask(s string, cont, noConfirm bool) bool {
 		return cont
 	}
 
-	response = strings.ToLower(response)
-
 	return strings.EqualFold(response, yes) || strings.EqualFold(response, y)
 }
--- a/pkg/text/text.go
+++ b/pkg/text/text.go
@@ -76,5 +76,5 @@ func ContinueTask(s string, cont, noConfirm bool) bool {
 
 	response = strings.ToLower(response)
 
-	return response == yes || response == y
+	return response == strings.ToLower(yes) || response == strings.ToLower(y)
 }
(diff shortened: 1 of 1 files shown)


Drumstickx:
Also fixes #1737

Jguer:
well spotted, that seems to be the issue indeed, thanks, I'll probably make a release soon, just unsure between making a patch only for this or a major including the new sort algorithm

pkg/text/text.go
```diff
@@ -76,5 +76,5 @@ func ContinueTask(s string, cont, noConfirm bool) bool {
 
 	response = strings.ToLower(response)
```
Jguer: In this case you should not need the ToLower if you're using `strings.EqualFold`, at least from checking the documentation

Jguer (changes requested):

pkg/text/text.go
Drumstickx: Correct, I missed that when fixing the linter violation. Thanks for the hint.

Jguer (approved):

PR #1777, opened 2022-07-24
add go dependency to README.md

kaue:
in a standard arch linux installation, the `go` binary seems to be a dependency to run `makepkg -i`


Morganamilo:
> in a standard arch linux installation, the go binary seems to be a dependency to run makepkg -i

Why not just pass -s to tell makepkg to install deps?

kaue:
That explains it @Morganamilo i didn't notice the `-s` argument in the README example. 
I guess we can close this if this argument already makes sure the `go` dependency will be installed with `makepkg`

PR #1802, opened 2022-08-24
Remove package from VCS on failed install

### LINKED ISSUE #1676: AUR devel packages get added to vcs.json even when the package fails to build/install
### Affected Version

Tested:
`yay v11.1.0 - libalpm v13.0.1`
`yay v11.1.0.r4.g09695c6 - libalpm v13.0.1`
The issue likely goes back for many versions.

### Describe the bug

If you attempt to install an AUR devel package and it fails for whatever reason, the package still gets added to vcs.json. This means that the git repository for the package will be checked on any system upgrade with the devel option, even though the package is not actually installed. The package also seems to be stuck in vcs.json unless the user manually removes the entry. (There may be a way to remove it I don't know, but I do know that `yay -R <package>` doesn't work since the package isn't installed).

### Reproduction Steps

For the reproduction steps, I'm going to use the AUR package [`run_scaled-git`](https://aur.archlinux.org/packages/run_scaled-git/). This package fails to install because one of its dependencies (`xpra`) provides the `run_scaled` script already.

1. `yay -S run_scaled-git`
2. Confirm that `run_scaled-git` was not installed (eg, looking at the output of the previous call, or running `yay -Qi run_scaled-git`)
3. Look in vcs.json for an entry for `run_scaled-git`

### Expected behavior

Since the package was not installed, it should not be added to vcs.json.

### Output

I truncated the build output for brevity.

```sh
$ yay -S run_scaled-git
...
Packages (1) run_scaled-git-r20.6699934-1
 
Total Installed Size:  0.01 MiB
 
:: Proceed with installation? [Y/n] 
(1/1) checking keys in keyring                               [#################################] 100%
(1/1) checking package integrity                             [#################################] 100%
(1/1) loading package files                                  [#################################] 100%
(1/1) checking for file conflicts                            [#################################] 100%
error: failed to commit transaction (conflicting files)
run_scaled-git: /usr/bin/run_scaled exists in filesystem (owned by xpra)
Errors occurred, no packages were upgraded.
 -> exit status 1
$ yay -Qi run_scaled-git
error: package 'run_scaled-git' was not found
$ grep run_scaled-git ~/.cache/yay/vcs.json
        "run_scaled-git": {
$
```

davejab:
Fixes #1676

### DIFF
--- a/install.go
+++ b/install.go
@@ -626,7 +626,7 @@ func buildInstallPkgbuilds(
 	remoteNamesCache := stringset.FromSlice(remoteNames)
 	localNamesCache := stringset.FromSlice(localNames)
 
-	for _, base := range do.Aur {
+	for i, base := range do.Aur {
 		pkg := base.Pkgbase()
 		dir := filepath.Join(config.BuildDir, pkg)
 		built := true
@@ -654,6 +654,9 @@ func buildInstallPkgbuilds(
 			exp = make([]string, 0)
 
 			if err != nil {
+				if i != 0 {
+					go config.Runtime.VCSStore.RemovePackage([]string{do.Aur[i-1].String()})
+				}
 				return err
 			}
 		}
@@ -784,6 +787,9 @@ func buildInstallPkgbuilds(
 	}
 
 	err = doInstall(ctx, arguments, cmdArgs, deps, exp)
+	if err != nil {
+		go config.Runtime.VCSStore.RemovePackage([]string{do.Aur[len(do.Aur)-1].String()})
+	}
 	settings.NoConfirm = oldConfirm
 
 	return err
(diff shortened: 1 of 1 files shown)


Jguer:
Hey @davejab , that part of the code is not very healthy so I'm ok to accept this patch, I can't think of a better way to solve it in the current architecture ( would prefer a single point where this would be evaluated). If you can just fix up the linting, I'll merge it, thanks :+1:

davejab:
@Jguer thanks for reviewing, I have fixed the linting errors, if there is anything else let me know :+1:

Jguer (approved):

Jguer:
Thanks @davejab for the patch

Issue #1799, opened 2022-08-22
Chinese characters are not aligned when using yay -Si


QiE2035:
### Affected Version

yay v11.3.0 - libalpm v13.0.1.

### Describe the bug

Chinese characters are not aligned when using `yay -Si`.

### Reproduction Steps

1. Change the language to Chinese.
2. `yay -Si aur/yay`

### Expected behavior

Chinese characters are aligned like this:

![image](https://user-images.githubusercontent.com/18079122/185847082-14e30894-9eb6-4559-abfa-99e851feb19b.png)

### Output

![image](https://user-images.githubusercontent.com/18079122/185847202-8b2635e4-0c39-456e-b83d-978a72222495.png)

```json
{
        "aururl": "https://aur.archlinux.org",
        "buildDir": "/tmp/yay/",
        "editor": "",
        "editorflags": "",
        "makepkgbin": "makepkg",
        "makepkgconf": "",
        "pacmanbin": "powerpill",
        "pacmanconf": "/etc/pacman.conf",
        "redownload": "no",
        "rebuild": "no",
        "answerclean": "",
        "answerdiff": "",
        "answeredit": "",
        "answerupgrade": "",
        "gitbin": "git",
        "gpgbin": "gpg",
        "gpgflags": "",
        "mflags": "",
        "sortby": "votes",
        "searchby": "name-desc",
        "gitflags": "",
        "removemake": "ask",
        "sudobin": "sudo",
        "sudoflags": "",
        "requestsplitn": 150,
        "completionrefreshtime": 7,
        "maxconcurrentdownloads": 0,
        "bottomup": true,
        "sudoloop": false,
        "timeupdate": false,
        "devel": false,
        "cleanAfter": false,
        "provides": false,
        "pgpfetch": true,
        "upgrademenu": true,
        "cleanmenu": true,
        "diffmenu": true,
        "editmenu": false,
        "combinedupgrade": false,
        "useask": false,
        "batchinstall": false,
        "singlelineresults": false,
        "separatesources": true,
        "version": "11.3.0"
}

```

### How to solve

I think this is because In a monospaced font, the width of a Chinese character should be equal to the width of **two** English characters (full-width and half-width), but yay may simply judge the number of characters without paying attention to this problem, so This leads to unaligned output, so properly judging the character width problem when formatting the output may help to solve this problem.

Jguer:
Closed by https://github.com/Jguer/yay/pull/1803 , thanks @linweiyuan

PR #1803, opened 2022-08-25
fix(yay): fix -Si alignment when language is Chinese (#1799)

linweiyuan:
before fix:
image

after fix:
image

### DIFF
--- a/pkg/text/print.go
+++ b/pkg/text/print.go
@@ -92,24 +92,24 @@ func PrintInfoValue(key string, values ...string) {
 	)
 
 	var (
-		str          string
 		language     = gotext.GetLanguage()
 		keyTextCount = keyLength - delimCount
 	)
 
 	if strings.HasPrefix(language, "zh") {
 		keyRuneCount := utf8.RuneCountInString(key)
-		if len(key) == keyRuneCount {
-			// these keys are all English words, do nothing, use the existing logic
-		} else if strings.HasSuffix(language, "TW") && key == "AUR 網址" {
+		switch {
+		case len(key) == keyRuneCount:
+		// these keys are all English words, do nothing, use the existing logic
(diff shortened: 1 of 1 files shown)


pkg/text/print.go
```diff
@@ -90,7 +91,26 @@ func PrintInfoValue(key string, values ...string) {
...
+		language     = gotext.GetLanguage()
+		keyTextCount = keyLength - delimCount
+	)
+
+	if strings.HasPrefix(language, "zh") {
+		keyRuneCount := utf8.RuneCountInString(key)
+		switch {
+		case len(key) == keyRuneCount:
+		// these keys are all English words, do nothing, use the existing logic
+		case strings.HasSuffix(language, "TW") && key == "AUR 網址":
```
Jguer: This is liable to change and may be broken :frowning:

Jguer (changes requested):
I think this approach might be error prone and go out of sync easily.

Alternative approaches:

- Go allows us to get the real length in bytes instead of in characters, maybe this can be used for a generic solution
https://www.golangprograms.com/golang-get-number-of-bytes-and-runes-in-a-string.html 
https://stackoverflow.com/questions/49397913/difference-between-length-of-rune-slice-and-runecountinstring

- or, I think you might be able to use https://siongui.github.io/2018/10/11/go-check-if-rune-value-is-chinese-character/ instead of grabbing the key, but then it will be only chinese symbolic specific (which for now seems to be ok)

pkg/text/print.go
linweiyuan: You are right, let me update this.

Jguer:
@linweiyuan if you're having issues with getting the linter running locally, I can handle touching up the linting before merging if that helps :+1:

Jguer (approved):
Thanks for your contribution and changing it up @linweiyuan :slightly_smiling_face: , I'm merging this

PR #1806, opened 2022-09-07
Fix environment variable name in man-page

Mudskipper875:
https://github.com/Jguer/yay/blob/0d1e339fda19fed487cb912b9b122d27a4db2556/pkg/settings/config.go#L266


Jguer (approved):
Thanks for noticing and fixing this @Mudskipper875 :slightly_smiling_face:

PR #1826, opened 2022-11-08
yay v12: new AUR query engine, Local PKGBuild install, debug mode, new install engine + bugs (+bugs)

Jguer:
- Add AUR cache engine (from metadata endpoint)
- Add local PKGBUILD install (`-Bi`)
  - -Bi can be used with several directories as targets 
- Add new basic install engine  `	"newinstallengine": true,` in `config.json`
- Add debug mode `	"debug": true,`

--- 

TODO:

- [x] port missing menus to install flow (new pluggable menu framework)
- [x] missing options (ex: dep ignore) 
- [x] pkg skip pull
- [x] group install
- [x] VCS info write


Morganamilo:
> Adds local PKGBUILD install (-Ui)

This is the way I went with paru but decided it was a mistake to have current dir semantics like makepkg. Instead I'm moving it to `-Bi <dir>`.

Jguer:
> > Adds local PKGBUILD install (-Ui)
> 
> This is the way I went with paru but decided it was a mistake to have current dir semantics like makepkg. Instead I'm moving it to `-Bi <dir>`.

:nod: makes sense, what does `-B` stand for? `--build`?

Morganamilo:
Yep --build.

Jguer:
@Morganamilo confirm me just one thing please, for building in the working directory, paru would need `paru -Bi .` ? 

From the man page of paru I don't find any support for `-Gi`, is that planned?

Morganamilo:
> paru would need paru -Bi .

Yep

> From the man page of paru I don't find any support for -Gi, is that planned?

Nope

Issue #1758, opened 2022-06-09
yay -Sc does not clean the .cache/yay folder


arch-user-france1:
### Affected Version

yay v11.1.2 - libalpm v13.0.1

### Reproduction Steps

1. `yay -Sc`
2. I chose yes for every question, make sure that you at least say yes if it asks you to clean the yay cache.
3. See result: `du -hc ~/.cache/yay | tail -1` said me 37 GB.

### Expected behavior

Yay would clean the cache .cache/yay

### Output

```sh
{
        "aururl": "https://aur.archlinux.org",
        "buildDir": "/home/france1/.cache/yay",
        "editor": "",
        "editorflags": "",
        "makepkgbin": "makepkg",
        "makepkgconf": "",
        "pacmanbin": "pacman",
        "pacmanconf": "/etc/pacman.conf",
        "redownload": "no",
        "rebuild": "no",
        "answerclean": "",
        "answerdiff": "",
        "answeredit": "",
        "answerupgrade": "",
        "gitbin": "git",
        "gpgbin": "gpg",
        "gpgflags": "",
        "mflags": "",
        "sortby": "votes",
        "searchby": "name-desc",
        "gitflags": "",
        "removemake": "ask",
        "sudobin": "sudo",
        "sudoflags": "",
        "requestsplitn": 150,
        "completionrefreshtime": 7,
        "bottomup": true,
        "sudoloop": false,
        "timeupdate": false,
        "devel": false,
        "cleanAfter": false,
        "provides": true,
        "pgpfetch": true,
        "upgrademenu": true,
        "cleanmenu": true,
        "diffmenu": true,
        "editmenu": false,
        "combinedupgrade": false,
        "useask": false,
        "batchinstall": false,
        "singlelineresults": false,
        "version": "11.1.2"
}
```

Note: There was no feedback at all, it just asked for all the file locations and then closed. What did I answer wrong?
And this thing was here: `Fehler: Konnte Datei /var/cache/pacman/pkg/gedit-plugins-41.0-1-x86_64.pkg.tar.zst nicht öffnen: Truncated zstd input`

pulga10:
See my comment here: https://github.com/Jguer/yay/issues/772#issuecomment-1152890843
 
[...]

PR #1863, opened 2022-12-20
fix(clean): notify RemoveAll error when cleaning AUR

joanbm:
Otherwise if any hard-to-delete file lands in the AUR cache folder, running `yay -Scc` will appear to succeed to the user, but actually abort midway.

*Practical relevance*: Aside from the obvious ways to create a file you can't delete inside `.cache/yay`, if you build a Go package without passing the `-modcacherw` flag to `go build`, it creates files inside `.cache/yay` that can't be deleted with just a `os.RemoveAll()` call (or just using the `rm` shell command).  (See also the Go package guidelines: https://wiki.archlinux.org/title/Go_package_guidelines#Flags_and_build_options)

With this change, at least yay tells you that it can't delete the those files instead of failing without notice. For example:

```shell
$ # Let's create a hard-to-delete folder like Go
$ mkdir -p .cache/yay/annoying_folder/inner
$ chmod 555 .cache/yay/annoying_folder -R
$ rm -rf .cache/yay/annoying_folder
rm: cannot remove '.cache/yay/annoying_folder/inner': Permission denied
$ rm -rf .cache/yay/annoying_folder/inner/
rm: cannot remove '.cache/yay/annoying_folder/inner/': Permission denied
$ # -------------------------------
$ # Let's try to delete it with yay
$ yay -Scc --noconfirm

:: Do you want to remove ALL files from cache? [y/N] 

Database directory: /var/lib/pacman/
:: Do you want to remove unused repositories? [Y/n] 
removing unused sync repositories...

Build directory: /home/sol/.cache/yay
removing AUR packages from cache...
 -> Unable to remove /home/sol/.cache/yay/annoying_folder: unlinkat /home/sol/.cache/yay/annoying_folder/inner: permission denied
$ # !!! Before this PR, the line above didn't appear and yay appeared to silently succeed!
```

Possibly related to #1758.

### DIFF
--- a/clean.go
+++ b/clean.go
@@ -154,9 +154,10 @@ func cleanAUR(ctx context.Context, keepInstalled, keepCurrent, removeAll bool, d
 			}
 		}
 
-		err = os.RemoveAll(filepath.Join(config.BuildDir, file.Name()))
+		dir := filepath.Join(config.BuildDir, file.Name())
+		err = os.RemoveAll(dir)
 		if err != nil {
-			return nil
+			text.Warnln(gotext.Get("Unable to remove %s: %s", dir, err))
 		}
 	}
 


joanbm:
Example AUR package that creates files yay can't delete inside the cache: https://aur.archlinux.org/packages/ryzen-stabilizator-git

clean.go
```diff
@@ -156,7 +156,7 @@ func cleanAUR(ctx context.Context, keepInstalled, keepCurrent, removeAll bool, d
 
 		err = os.RemoveAll(filepath.Join(config.BuildDir, file.Name()))
 		if err != nil {
-			return nil
+			return err
```
Jguer: Hey @joanbm , I would change this to a `text.Warnln` so the removal action can continue and is not blocked.

Jguer (changes requested):
I would change it so it's non blocking but still displays messages. `text.Warnln` or `text.Errorln` would probably be the best to display this

clean.go
joanbm: Updated. I've maintained the `return err` inside `cleanAUR` (since logically the operation fails), and added the `text.Warnln`+continue logic at the caller (`syncClean`) instead.

Note that this also makes the other possible errors from `cleanAUR` (e.g.  fetching from AUR failed) non-blocking. Let me know if that sounds good to you.

joanbm: Sounds good to me, done. This way yay will still clean all top-level entries in the cache it can, and notify the user of those it can't.

Jguer (approved):

PR #1865, opened 2022-12-27
fix(pre-commit): update URL and versions in the pre-commit configuration

jdholtz:
There was a dead link in the `pre-commit-golang` URL, which caused `pre-commit install` to fail. Also, `Woile/commitizen` redirects to `commitizen-tools/commitizen`, so change that URL as well. Last, update all the versions using `pre-commit autoupdate`. All hooks still pass.

This will make it easier for new contributors to use the pre-commit hooks when contributing.

Last, I noticed that the `news_test.go` unit test doesn't work in some timezones (e.g. it works in `Europe/Paris` but not in `America/Denver`). This causes the pre-commit hook to fail. Wanted to mention this in case other contributors from different timezones run into this issue (or this is an actual issue within the news code).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,8 +1,8 @@
 default_stages: [commit]
 
 repos:
-  - repo: git://github.com/dnephin/pre-commit-golang
-    rev: v0.3.5
+  - repo: https://github.com/dnephin/pre-commit-golang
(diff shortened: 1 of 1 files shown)


Jguer (approved):

Jguer:
I see, weird that it's working in the CI, I'll investigate

PR #1893, opened 2023-01-13
Fix zsh completion for yay specific flags

balki:
Example: 
Before fix:
```
❯ yay -P<TAB>
Completing no arguments
```
After fix:
```
❯ yay -P<TAB>
Completing option
-c  -- Used for completions
-d  -- Print default yay configuration
-g  -- Print current yay configuration
-n  -- Print number of updates
-s  -- Display system package statistics
-u  -- Print update list
-w  -- Print arch news
```
---
Other minor changes
  * Remove duplicate entry for flag `T`
  * Remove vim modeline option `fdm=marker foldlevel=0`.
    There are no fold markers (a fold marker is `{{{` in vim) in the file. So these options don't do anything.

### DIFF
--- a/completions/zsh
+++ b/completions/zsh
@@ -1,5 +1,5 @@
 #compdef yay
-# vim:fdm=marker foldlevel=0 tabstop=2 shiftwidth=2 filetype=zsh
+# vim:tabstop=2 shiftwidth=2 filetype=zsh
 
 typeset -A opt_args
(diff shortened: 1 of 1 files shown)


Jguer (approved):

Jguer:
Thanks for the patch @balki, feels good to see the zsh completions getting some maintenance

PR #1940, opened 2023-02-26
Respect global arguments when removing make dependencies

### LINKED ISSUE #1869: Yay doesn't respect --root argument when removing make dependencies
### Affected Version
yay v11.3.1.r73.g2bda76e - libalpm v13.0.2

### Describe the bug
From using the `debug` option while running Yay, I noticed that Yay does not pass the `--root` argument to pacman when removing make dependencies. This is an issue because, when the `--root` argument is used, pacman installs the make dependencies to the directory provided. However, when pacman tries to remove those dependencies later on, it looks for them in the system database, not the root directory provided, which causes it to error with "target not found".

### Reproduction Steps
This is a bit annoying to reproduce, but I'll try to explain it the best I can
1. First, create an executable file that passes pacman the root argument. This is needed because makepkg needs to use pacman with the `--root` argument. For makepkg to use this, the `PACMAN` env variable needs to be set.
File I used:
```
#!/bin/bash
_pacman() {
    pacman --root "/path/to/test-root-dir" "$@"
}

_pacman "$@"
```
2. Next, make a clean root using `mkarchroot <test-root-dir> base`
3. Last, execute an AUR install with yay. I can reproduce this issue with `proot`. Run the command `PACMAN=<path/to/pacman-executable> yay --root <path/to/test-root-dir> -S proot`.
Note: Make sure the removeMake option is not set to `no` and at least one make dependency is not already installed on the system (the example that errored for me was `python-docutils`). Then, watch how yay throws an error that `target not found: python-docutils` after running the removeMake function.

### Expected behavior
The `--root` argument should be passed into pacman when removing make dependencies if it was passed into yay.

### Output
Here is the output from yay in debug mode:
```sh
[DEBUG] running /usr/bin/sudo pacman --root /<root-dir> -D -q --asexplicit --noconfirm --config /etc/pacman.conf -- proot
[DEBUG] running /usr/bin/sudo pacman -R -u --noconfirm --config /etc/pacman.conf -- libnsl python python-docutils libxslt
error: target not found: python-docutils
```

Here is the relevant config:
```sh
{
    "removemake": "ask",
    "debug": "true"
}
```
This issue most likely persists with other global arguments besides `--root`.

You can see in [this](https://github.com/Jguer/yay/blob/9be51052f721de7f286d3b41bb389de7c6692ac9/install.go#L222) block that the `arguments` are not passed to the `removeMake` function. I can submit a fix for this. I think this can be solved by just passing in arguments into removeMake and adding the global arguments to the `removeArguments` variable in the `removeMake` function.

jdholtz:
Fixes #1869. 

Ensure all global arguments are used when calling `pacman` to remove make dependencies.

### DIFF
--- a/install.go
+++ b/install.go
@@ -224,7 +224,7 @@ func install(ctx context.Context, cmdArgs *parser.Arguments, dbExecutor db.Execu
 		switch config.RemoveMake {
 		case "yes":
 			defer func() {
-				err = removeMake(ctx, config.Runtime.CmdBuilder, do.GetMake())
+				err = removeMake(ctx, config.Runtime.CmdBuilder, do.GetMake(), cmdArgs)
 			}()
 
 		case "no":
@@ -232,7 +232,7 @@ func install(ctx context.Context, cmdArgs *parser.Arguments, dbExecutor db.Execu
 		default:
 			if text.ContinueTask(os.Stdin, gotext.Get("Remove make dependencies after install?"), false, settings.NoConfirm) {
 				defer func() {
-					err = removeMake(ctx, config.Runtime.CmdBuilder, do.GetMake())
+					err = removeMake(ctx, config.Runtime.CmdBuilder, do.GetMake(), cmdArgs)
 				}()
 			}
 		}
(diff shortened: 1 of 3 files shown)


Jguer (approved):

Jguer:
:+1: thanks @jdholtz

PR #1977, opened 2023-03-19
fix(aur_install): fix debug pacakges not being found

Ferdi265:
This PR fixes debug packages no longer being installed correctly.

The bug was a simple accidental swapping of package path and package name.


Jguer (approved):

Jguer:
Thanks for the PR @Ferdi265 , nice catch

Jguer:
TODO: I'll write a test to ensure this does not regress

PR #2115, opened 2023-04-12
Added documentation for Steam Deck installation

ProtocolNebula:
I've added the commands necessary to install in steam deck, the first time I did it took me like an hour or two due a lot of errors, so will be great to have the docs accessible for all users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,40 @@ pacman -S --needed git base-devel yay
 
 ⚠️ distributions sometimes lag updating yay on their repositories.
 
+#### Steam Deck
+
+In Steam Deck we will have some additional issues: Errors with `GPG Keys` and `system read-only protection`, to do the installation without a lot of headaches just run:
+
+```sh
+# Allow writing over file system (steam os block)
+sudo steamos-readonly disable
+
+# Ensure we have the keys updated
+sudo pacman-key --init
+sudo pacman-key --populate archlinux
+
+# Install git and base-devel dependencies
+sudo pacman -S --needed git base-devel
+# Ensure that dependencies get installed correctly (this may be not necessary in some cases)
+sudo pacman -S git base-devel
+# When prompted, select all options
+
+# INSTALL FROM SOURCE
+# git clone https://aur.archlinux.org/yay.git
+# cd yay
+# makepkg -si
+
+# OR INSTALL FROM BINARY (fastest)
+git clone https://aur.archlinux.org/yay-bin.git
+cd yay-bin
+makepkg -si
+
+
+# Disable again the write file system permissions
+sudo steamos-readonly disable
+```
+
+
 ## First Use
 
 #### Development packages upgrade


0xDEADCADE:
It should be noted that the Steam Deck resets its rootfs on update, and sometimes reboot after a crash due to integrity checks.

While these instructions will install yay, it is not permanent, and `steamos-readonly disable` would need to be used after every reboot when a package needs to be installed or updated. A decent chunk of files from packages in `base-devel` and other packages (specifically header files) are missing from SteamOS by default, so many AUR packages will fail to build without manually reinstalling repository dependencies. SteamOS is also not rolling release, and is behind Arch Linux repositories, making the `yay-bin` package potentially incompatible due to library updates. Maybe include warnings for these SteamOS specific quirks?

It is also possible to install ArchLinux in something similar to [distrobox](https://github.com/89luca89/distrobox), which doesn't have the SteamOS specific downsides, and can use the generic Arch Linux install instructions.

ProtocolNebula:
I have Steam Deck for just  2 weeks and I had no problem yet related to the dependencies removal or required to execute `steamos-readonly disable`.

For sure, there are some packages that have troubles, but I'm not linux pro to solve that.
I'm struggling to change the installation folder for pacman (and yay) and is giving some headaches related to the paths/system integration.

README.md
```diff
@@ -73,6 +73,40 @@ pacman -S --needed git base-devel yay
...
+# makepkg -si
+
+# OR INSTALL FROM BINARY (fastest)
+git clone https://aur.archlinux.org/yay-bin.git
+cd yay-bin
+makepkg -si
+
+
+# Disable again the write file system permissions
+sudo steamos-readonly disable
```
Jguer, line 106: does need to be disabled again as it was disabled at the beginning?

README.md
```diff
@@ -73,6 +73,40 @@ pacman -S --needed git base-devel yay
...
+# Install git and base-devel dependencies
+sudo pacman -S --needed git base-devel
+# Ensure that dependencies get installed correctly (this may be not necessary in some cases)
+sudo pacman -S git base-devel
+# When prompted, select all options
+
+# INSTALL FROM SOURCE
+# git clone https://aur.archlinux.org/yay.git
+# cd yay
+# makepkg -si
```
Jguer, line 97: I would avoid putting commented code. Prefer adding a mention at the end saying `Install from source`` can be used instead and pointing to the part of the readme

Jguer:
I do not own a steam deck so I can't be sure if the instructions are correct, but I appreciate adding these in.

I think adding the warnings that @0xDEADCADE suggests are a nice addition before merging

README.md
ProtocolNebula: Yes, it would be even better, that was just for the new users to open source who do not know how to read docs, but that was my first idea.

pmarreck:
I just noticed this. Not sure why I wasn't seeing notifications for these! I will try to merge any outstanding PR's

Issue #1961, opened 2023-03-11
Upgrade menu: semver highlighting is sometimes wrong


ariasuni:
### Affected Version

yay v11.3.2 - libalpm v13.0.2

### Describe the bug

When updating some packages, the semver highlighting is wrong. I noticed it being off by one, but maybe in some cases it’s different.

Here, wen updating `linux` and `linux-headers` from `6.2.2.arch1-1` to `6.2.2.arch2-2`, only `rch1-1` and `rch2-2` parts are highlighted, missing the `a`:
![image](https://user-images.githubusercontent.com/2446451/224484797-8cafb306-8745-415b-83cc-0096e109d0e2.png)

### Reproduction Steps

1. Upgrade package from specific versions (e.g. `6.2.2.arch1-1` to `6.2.2.arch2-2`, not sure if other patterns are applicable)

### Expected behavior

In the example provided, `arch1-1` is colored in red and `arch2-2` is colored in green.

### Output

```sh
$ yay -Pg
{
        "aururl": "https://aur.archlinux.org",
        "aurrpcurl": "",
        "buildDir": "/home/ariasuni/.cache/yay",
        "editor": "",
        "editorflags": "",
        "makepkgbin": "makepkg",
        "makepkgconf": "",
        "pacmanbin": "pacman",
        "pacmanconf": "/etc/pacman.conf",
        "redownload": "no",
        "rebuild": "no",
        "answerclean": "",
        "answerdiff": "",
        "answeredit": "",
        "answerupgrade": "",
        "gitbin": "git",
        "gpgbin": "gpg",
        "gpgflags": "",
        "mflags": "",
        "sortby": "votes",
        "searchby": "name-desc",
        "gitflags": "",
        "removemake": "ask",
        "sudobin": "sudo",
        "sudoflags": "",
        "requestsplitn": 150,
        "completionrefreshtime": 7,
        "maxconcurrentdownloads": 0,
        "bottomup": true,
        "sudoloop": false,
        "timeupdate": false,
        "devel": false,
        "cleanAfter": false,
        "provides": true,
        "pgpfetch": true,
        "upgrademenu": true,
        "cleanmenu": true,
        "diffmenu": true,
        "editmenu": false,
        "combinedupgrade": true,
        "useask": false,
        "batchinstall": false,
        "singlelineresults": false,
        "separatesources": true,
        "version": "11.3.2"
}
```

Jguer:
Closed with https://github.com/Jguer/yay/pull/2124

PR #2124, opened 2023-04-18
Fix version diff word detection

tchaudhry91:
Attempted Fix for https://github.com/Jguer/yay/issues/1961

The core problem is the `checkWords` function.
e.g while traversing through `arch`, it finds `rc` as well. Thereby goes off by 1.
I have added some tests and changed the `checkWords` to return true only if the current character is not a Letter (i.e not part of a bigger word).
Other tests are also passing.

### DIFF
--- a/pkg/query/version_diff_test.go
+++ b/pkg/query/version_diff_test.go
@@ -1,0 +1,62 @@
+package query
+
+import (
+	"testing"
+
+	"github.com/Jguer/yay/v12/pkg/text"
+)
+
+func TestVersionDiff(t *testing.T) {
+	testCases := []struct {
+		name     string
+		a        string
+		b        string
+		wantDiff string
+	}{
+		{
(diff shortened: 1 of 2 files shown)


pkg/query/version_diff_test.go
```diff
@@ -0,0 +1,62 @@
+package query
+
+import (
+	"testing"
+
+	"github.com/Jguer/yay/v12/pkg/text"
+)
+
+func TestVersionDiff(t *testing.T) {
```
Jguer, line 9: I think you have to force `text.Color = true` and then unset it at the end of the test to stop the automatic tty detection from turning off colors.

One of the unfortunate places where we haven't gotten rid of globals

tchaudhry91: Added the change.

Jguer (approved):
Thanks for the contribution, awesome to see this fixed

PR #2154, opened 2023-05-06
Fix formatting of error message that occurs on AUR errors

MonterraByte:
If there's an error while querying the AUR for a package, the following error message is shown:

```
 -> Failed to find AUR package forautotiling-git:1 error occurred:
    * AUR is unavailable at this moment
```

It's missing a space before the package name. This PR fixes that.


Jguer (approved):
Thanks for fixing this @jmonteiro64

Issue #2061, opened 2023-04-03
Implement tests for -Qu/q


Jguer:
Definition of Done:

- [ ] Implement output tests for `-Qu` and `-Quq`. It should be possible to test it like other tests present in `sync_test.go`

smolx:
@Jguer , on your PR review:
> The only case I think would miss would be asserting empty update list to make sure an error is returned (so that then the exit error is 1)

While looking into yay's return codes for different test cases with empty update list, I came across with discrepancy with pacman's behavior. For example, when we have some native packages to update, but no foreign ones:
```
$ yay -Qum
$ echo $?
0
$ sudo pacman -Qum
$ echo $?
1
```

Should yay return 1 as exit code as well? May I try to fix this?

Jguer:
Hey @smolx , yes this exit code should match :slightly_smiling_face:, if you can fix it that would be great

PR #2156, opened 2023-05-09
Implement tests for -Qu/q

smolx:
Draft version of tests for -Qu (#2061) with some debatable tricks to make it work.

Guidance and rebukes are welcome.

### DIFF
--- a/pkg/db/mock/executor.go
+++ b/pkg/db/mock/executor.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/Jguer/yay/v12/pkg/db"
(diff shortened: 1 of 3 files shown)


Jguer:
Hey @smolx, the tests look good.

The only case I think would miss would be asserting empty update list to make sure an error is returned (so that then the exit error is 1)

On the logging + noConfirm these are hacks I've also had to do for other tests until we can get rid of those globals.

Thanks for the PR, I can merge it in and if you want to pick up that last case feel free to :slightly_smiling_face:

Jguer (approved):

PR #2163, opened 2023-05-16
Fix --rebuild option

### LINKED ISSUE #2153: Rebuild option is ignored
### Affected Version

yay v12.0.4 - libalpm v13.0.2

### Describe the bug

Yay does not rebuild AUR packages, even though it is explicitly asked to via `--rebuild` e.g.

```
❯ yay -S autojump --rebuild
... blah blah ...
-> autojump-22.5.3-9 already made -- skipping build
```

Judging by the debug output, this is caused here -> https://github.com/Jguer/yay/blob/f7731d7cf936e301ab34aebdb7ad0eadc721c214/aur_install.go#L286-L299 - there is no check for the rebuild option in sight, only if the file exists.

### Expected behavior

`--rebuild` should always REBUILD AUR packages.

### Output

```sh
{
"buildDir": "/var/local/build",
"editor": "",
"editorflags": "",
"makepkgbin": "makepkg",
"makepkgconf": "",
"pacmanbin": "pacman",
"redownload": "no",
"rebuild": "no",
"answerclean": "",
"answerdiff": "",
"answeredit": "",
"answerupgrade": "",
"gitbin": "git",
"gpgbin": "gpg",
"gpgflags": "",
"mflags": "",
"sortby": "votes",
"searchby": "name-desc",
"gitflags": "",
"removemake": "ask",
"sudobin": "sudo",
"sudoflags": "",
"version": "12.0.4",
"requestsplitn": 150,
"completionrefreshtime": 7,
"maxconcurrentdownloads": 1,
"bottomup": true,
"sudoloop": false,
"timeupdate": false,
"devel": true,
"cleanAfter": false,
"provides": false,
"pgpfetch": true,
"upgrademenu": true,
"cleanmenu": true,
"diffmenu": true,
"editmenu": true,
"combinedupgrade": false,
"useask": true,
"batchinstall": false,
"singlelineresults": false,
"separatesources": true,
"newinstallengine": true,
"debug": false,
"rpc": true,
"doubleconfirm": true
}
```

smolx:
Now it seems to be working.
Fixes #2153

### DIFF
--- a/aur_install.go
+++ b/aur_install.go
@@ -27,7 +27,7 @@ type (
 		exeCmd           exe.ICmdBuilder
 		vcsStore         vcs.Store
 		targetMode       parser.TargetMode
-		reBuild          string
+		rebuildMode      parser.RebuildMode
 		origTargets      mapset.Set[string]
 		downloadOnly     bool
 		log              *text.Logger
@@ -38,7 +38,7 @@ type (
 
 func NewInstaller(dbExecutor db.Executor,
 	exeCmd exe.ICmdBuilder, vcsStore vcs.Store, targetMode parser.TargetMode,
-	reBuild string, downloadOnly bool, logger *text.Logger,
+	rebuildMode parser.RebuildMode, downloadOnly bool, logger *text.Logger,
 ) *Installer {
 	return &Installer{
 		dbExecutor:            dbExecutor,
@@ -47,7 +47,7 @@ func NewInstaller(dbExecutor db.Executor,
 		exeCmd:                exeCmd,
 		vcsStore:              vcsStore,
 		targetMode:            targetMode,
-		reBuild:               reBuild,
+		rebuildMode:           rebuildMode,
 		downloadOnly:          downloadOnly,
 		log:                   logger,
 		manualConfirmRequired: true,
@@ -275,7 +275,7 @@ func (installer *Installer) installAURPackages(ctx context.Context,
 
 func (installer *Installer) buildPkg(ctx context.Context,
 	dir, base string,
-	installIncompatible, needed bool, isTarget bool,
+	installIncompatible, needed, isTarget bool,
 ) (map[string]string, error) {
 	args := []string{"--nobuild", "-fC"}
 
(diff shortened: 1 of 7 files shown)


smolx:
@Jguer , I'm sorry, but do not rush to review. :) I think this fix isn't quite right, because it abuses bug #2164. If that bug is fixed, --rebuild option won't work for packages that were installed as dependency.

Probably, correct fix for --rebuild option is more complicated and cumbersome as you warned. I'll try to come up with something.

smolx:
Finally, logically correct implementation. Nuances are questionable.

aur_install.go
```diff
@@ -27,6 +27,8 @@ type (
 		exeCmd           exe.ICmdBuilder
 		vcsStore         vcs.Store
 		targetMode       parser.TargetMode
+		reBuild          string
```
Jguer: Hey, I'll test it out this week,  

- I'd turn this into a `type rebuildMode string` and declare some constants to avoid dynamic behavior (user's putting the wrong value in config and thinking it's configure)

nit: Don't really care but https://www.merriam-webster.com/dictionary/rebuild does not need to be cased as it's its own word

smolx: I also had a similar idea, but abandoned it, at least at this stage. It doesn't solve anything, since configuration machinery doesn't have any config value validation mechanism(perhaps this deserves a separate issue).

But if you insist that it's good to have it with an eye on the future, I'll do so.

I named the `Installer`'s field `reBuild` so just to be in line with the `Configuration`'s field `ReBuild`. I can rename both if I touch `Configuration`.

smolx:
I made the changes you suggested for clarity. But didn't rename `Configuration.ReBuild` field.
And also refactored some parts of my code.

Jguer:
I didn't look in detail into the tests but the base logic makes sense. Merging it, let's give ourselves some time to polish before the next release. 

---

On my side I'll see if I can make `doubleConfirm` work as intended and make sure `--noconfirm` is ignored when there are conflicts

Jguer (approved):

PR #2169, opened 2023-05-18
Fix AUR dependency resolving

### LINKED ISSUE #2157: yay attempts to install net-new AUR package for dependencies when dependency is already satisfied
### Affected Version
`yay v12.0.4.r21.g490ebe4 - libalpm v13.0.2`
Also affects the v12.0.4 release tag, **without** the behavioral inconsistency - i.e. v12.0.4 always tries to install `ceph-libs`. Will explain in further detail below.
### Describe the bug

In AUR, ceph is distributed as:
* Source: [ceph](https://aur.archlinux.org/packages/ceph) and [ceph-libs](https://aur.archlinux.org/packages/ceph-libs), which require compilation
* Binary: [ceph-bin](https://aur.archlinux.org/packages/ceph-bin) and [ceph-libs-bin](https://aur.archlinux.org/packages/ceph-libs-bin), which do not require compilation

`ceph-bin` provides `ceph`, and `ceph-libs-bin` provides `ceph-libs`.

I have the `-bin` variants installed, since the compilation time of the source variants is very long.

Both `ceph-bin` and `ceph-libs-bin` have recently been upgraded from `17.2.5-6` to `17.2.6-1`. It should be sufficient to upgrade both of these packages in place to satisfy all dependencies.

However, yay v12.0.4 (i.e. the **non-git** release of yay) incorrectly attempts to install the `ceph-libs` source package as a dependency. This behavior is consistent:
```
[eyanulis@crimson ~]$ yay
 core is up to date
 extra is up to date
 community is up to date
 multilib is up to date
:: 3 Packages to upgrade/install.
3  aur/ceph-bin       17.2.5-6 -> 17.2.6-1
2  aur/ceph-libs               -> 17.2.6-1
1  aur/ceph-libs-bin  17.2.5-6 -> 17.2.6-1
==> Packages to exclude: (eg: "1 2 3", "1-3", "^4" or repo name)
 -> Excluding packages may cause partial upgrades and break systems
==> ^C
```
I tested `yay-git` to see if #2147 might help here (since the behavior seems kinda similar to #2135). However, with `yay-git v12.0.4.r21.g490ebe4` the behavior gets stranger: consecutive runs of yay sometimes **do** try to install `ceph-libs`, and sometimes **do not** try to install `ceph-libs`. Condensed output below; full `--debug` output will be attached.
```
[eyanulis@crimson ~]$ date && yay 
Thu May 11 04:54:40 PM MDT 2023
 core is up to date
 extra is up to date
 community is up to date
 multilib is up to date
:: 2 packages to upgrade/install.
2  aur/ceph-bin       17.2.5-6 -> 17.2.6-1
1  aur/ceph-libs-bin  17.2.5-6 -> 17.2.6-1
==> packages to exclude: (eg: "1 2 3", "1-3", "^4" or repo name)
 -> Excluding packages may cause partial upgrades and break systems
==> ^C
[eyanulis@crimson ~]$ 
[eyanulis@crimson ~]$ date && yay 
Thu May 11 04:54:45 PM MDT 2023
 core is up to date
 extra is up to date
 community is up to date
 multilib is up to date
:: 1 dependency will also be installed for this operation
   aur/ceph-libs   -> 17.2.6-1

:: 2 packages to upgrade/install.
2  aur/ceph-bin       17.2.5-6 -> 17.2.6-1
1  aur/ceph-libs-bin  17.2.5-6 -> 17.2.6-1
==> packages to exclude: (eg: "1 2 3", "1-3", "^4" or repo name)
 -> Excluding packages may cause partial upgrades and break systems
==> ^C
[eyanulis@crimson ~]$ 
[eyanulis@crimson ~]$ date && yay 
Thu May 11 04:54:54 PM MDT 2023
 core is up to date
 extra is up to date
 community is up to date
 multilib is up to date
:: 2 packages to upgrade/install.
2  aur/ceph-bin       17.2.5-6 -> 17.2.6-1
1  aur/ceph-libs-bin  17.2.5-6 -> 17.2.6-1
==> packages to exclude: (eg: "1 2 3", "1-3", "^4" or repo name)
 -> Excluding packages may cause partial upgrades and break systems
==> ^C
```
I cannot identify a pattern any more specific than "sometimes it gets the dependency wrong, sometimes it gets it right". I hesitate to claim the behavior is non-deterministic, but it "feels" that way.

### Reproduction Steps

1. Have `ceph-bin 17.2.5-6` and `ceph-libs-bin 17.2.5-6` installed.
2. Execute `yay` (implied `yay -Syu`) which will identify that both packages can be upgraded to `17.2.6-1`.
3a. On yay v12.0.4, observe that yay (seemingly) always tries to install `ceph-libs` as a net-new package.
3b. On yay-git v12.0.4.r21.g490ebe4, observe that yay **intermittently/inconsistently** tries to install `ceph-libs` as a net-new package.

### Expected behavior

`ceph-libs` should not be installed, because this dependency is already satisfied by the installed package `ceph-libs-bin`.

### Output
Note: These were captured with `yay --debug -a` to limit operations to AUR-only. Behavior still occurs without `-a` flag.
[yay-config.txt](https://github.com/Jguer/yay/files/11456886/yay-config.txt)
[debug-output-wrong.txt](https://github.com/Jguer/yay/files/11456887/debug-output-wrong.txt)
[debug-output-good.txt](https://github.com/Jguer/yay/files/11456888/debug-output-good.txt)

smolx:
Before this fix dependencies for AUR targets were added to the graph after each addition of a target node. Now dependencies are added only after all target nodes are added to the graph.

Also added some tests for previously bugged cases.

Fixes #2157.

### DIFF
--- a/pkg/dep/dep_graph.go
+++ b/pkg/dep/dep_graph.go
@@ -270,6 +270,12 @@ func (g *Grapher) GraphFromSrcInfo(ctx context.Context, graph *topo.Graph[string
 	return graph, nil
 }
 
+func (g *Grapher) AddDepsForPkgs(ctx context.Context, pkgs []*aur.Pkg, graph *topo.Graph[string, *InstallInfo]) {
+	for _, pkg := range pkgs {
+		g.addDepNodes(ctx, pkg, graph)
+	}
+}
+
 func (g *Grapher) addDepNodes(ctx context.Context, pkg *aur.Pkg, graph *topo.Graph[string, *InstallInfo]) {
 	if len(pkg.MakeDepends) > 0 {
 		g.addNodes(ctx, graph, pkg.Name, pkg.MakeDepends, MakeDep)
@@ -316,8 +322,6 @@ func (g *Grapher) GraphAURTarget(ctx context.Context,
 		graph = topo.New[string, *InstallInfo]()
 	}
 
-	exists := graph.Exists(pkg.Name)
-
 	graph.AddNode(pkg.Name)
 
 	for i := range pkg.Provides {
@@ -334,10 +338,6 @@ func (g *Grapher) GraphAURTarget(ctx context.Context,
 		Background: bgColorMap[AUR],
 		Value:      instalInfo,
 	})
-
-	if !exists {
-		g.addDepNodes(ctx, pkg, graph)
-	}
 
 	return graph
 }
@@ -365,6 +365,8 @@ func (g *Grapher) GraphFromAUR(ctx context.Context,
 			g.providerCache[pkg.Name] = []aurc.Pkg{*pkg}
 		}
 	}
+
+	aurPkgsAdded := []*aurc.Pkg{}
 
 	for _, target := range targets {
 		if cachedProvidePkg, ok := g.providerCache[target]; ok {
@@ -405,7 +407,10 @@ func (g *Grapher) GraphFromAUR(ctx context.Context,
 			Source:  AUR,
 			Version: aurPkg.Version,
 		})
+		aurPkgsAdded = append(aurPkgsAdded, aurPkg)
 	}
+
+	g.AddDepsForPkgs(ctx, aurPkgsAdded, graph)
 
 	return graph, nil
 }
--- a/pkg/dep/dep_graph_test.go
+++ b/pkg/dep/dep_graph_test.go
@@ -319,3 +319,348 @@ func TestGrapher_GraphProvides_androidsdk(t *testing.T) {
 		})
 	}
 }
+
+func TestGrapher_GraphFromAUR_Deps_ceph_bin(t *testing.T) {
+	mockDB := &mock.DBExecutor{
+		SyncPackageFn:       func(string) mock.IPackage { return nil },
+		PackagesFromGroupFn: func(string) []mock.IPackage { return []mock.IPackage{} },
+		SyncSatisfierFn: func(s string) mock.IPackage {
+			switch s {
+			case "ceph-bin", "ceph-libs-bin":
+				return nil
+			case "ceph", "ceph-libs", "ceph-libs=17.2.6-2":
+				return nil
+			}
+
+			panic("implement me " + s)
+		},
+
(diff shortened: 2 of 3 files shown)


smolx:
I had the audacity to propose my solution to the issue.
Perhaps new and existing related methods could be named better. And there are definitely other things to improve.

I also found a similar pattern of dependency resolving in the GraphFromSrcInfo method. Maybe it should be fixed too. I haven't dug in it yet.

Jguer (commented):
From a first read it seems correct. 

I had a doubt if it was working for recursive AUR dependencies but I haven't tried it out locally yet.

How does it handle `yay -Syu ros-melodic-desktop`?

The exists check that was removed had as an objective to prevent the computation of the dep graph for the same AUR package from repeating on each dependency layer

smolx:
With this fix there is a significant difference only for cases with **multiple** targets, since it doesn't affect the recursive part.

`yay -Syu ros-melodic-desktop` and `yay -S ros-melodic-desktop` are handled well:
```
Sync Make Dependency (16)
Sync Dependency (18)
AUR Explicit (1)
AUR Dependency (136)
AUR Make Dependency (66)
```
It gives more or less the same result as the `next` branch. "More or less" is because amount of make and regular dependencies change on each run, even on the `next` branch. :)

> The exists check that was removed had as an objective to prevent the computation of the dep graph for the same AUR package from repeating on each dependency layer

If I figured it out correctly, repeated computation during recursive traversal is prevented completely in the `Grapher.addNodes` method. So that removed check would help(in combination with this fix) only with cases of duplicate targets. Though performance effect is minimal. Target duplicates could be filtered out much more earlier.

Jguer:
Still very happy to see this implemented. Let's merge it in

Jguer (approved):

Jguer:
> I also found a similar pattern of dependency resolving in the GraphFromSrcInfo method. Maybe it should be fixed too. I haven't dug in it yet.

`GraphFromSrcInfo` has some low hanging fruit to fix if I remember correctly.

PR #2175, opened 2023-05-21
Fix yay -Sc wiping ~/.cache/yay on 3rd question.

### LINKED ISSUE #2152: yay --aur -Sc no longer keep installed AUR package cache folders
### Affected Version

yay v12.0.4.r18.gf7731d7 - libalpm v13.0.2

### Describe the bug

I regularly use `yay --aur -Sc` command to clean up uninstalled AUR packages in the yay cache folder. When I upgraded yay from 11.3.2 to the 12.0.4, this command will remove all packages in the cache folder.

### Reproduction Steps

1. set the CleanMethod in /etc/pacman.conf to be `keepInstalled`
2. use yay to install one package from AUR
3. execute `yay --aur -Sc` and select yes to the first prompt
4. check the just installed package cache folder is gone

### Expected behavior

Keep all installed AUR package cache folders.

pteromys:
If you answer yes to

>    :: Do you want to remove all other AUR packages from cache? [Y/n]

then we run `cleanAUR()`, intending to remove subdirectories of `~/.cache/yay` that do not share a name with installed packages not found in the sync repositories.

Where this was going wrong was `cleanAUR`'s invocation was getting an empty map from `dbExecutor.InstalledRemotePackages()`—because `InstalledRemotePackages` only recomputed its result if `installedRemotePkgMap` was `nil` whereas `NewExecutor` initialized it to an empty map. The symptom was it emptied my `~/.cache/yay`.

We do want a non-nil, empty `installedRemotePkgMap` to block recomputing (that is, to indicate the user really has no remote packages), so now `NewExecutor` initializes it to nil, and `getPackageNamesBySource` is responsible for making sure it's non-nil before writing to it.

Fixes #2152, which seems to have been introduced in 4626a0409c4d34bcffe0d5ed499ebff893115c69.

### DIFF
--- a/pkg/db/ialpm/alpm.go
+++ b/pkg/db/ialpm/alpm.go
@@ -38,7 +38,7 @@ func NewExecutor(pacmanConf *pacmanconf.Config, logger *text.Logger) (*AlpmExecu
 		conf:                    pacmanConf,
 		log:                     logger,
 		installedRemotePkgNames: nil,
-		installedRemotePkgMap:   map[string]alpm.IPackage{},
+		installedRemotePkgMap:   nil,
 		installedSyncPkgNames:   nil,
 	}
 
--- a/pkg/db/ialpm/high_level.go
+++ b/pkg/db/ialpm/high_level.go
@@ -8,6 +8,9 @@ import (
 
 // GetPackageNamesBySource returns package names with and without correspondence in SyncDBS respectively.
 func (ae *AlpmExecutor) getPackageNamesBySource() {
+	if ae.installedRemotePkgMap == nil {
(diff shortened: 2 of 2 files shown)


Jguer:
Should the fix not be setting the `installedRemotePkgMap` to nil on instantiation?

In this case if the user does not have remote packages it will always be recomputed

pteromys:
Oh, yes, that would be better. Updated!

Jguer:
Thanks for the PR and the fix @pteromys :smiley: , I'm merging it as soon as it passes the CI

Jguer (approved):

PR #2181, opened 2023-05-22
chore(yay): add warning for yay -c

Jguer:
Based on feedback, return warning on usage of yay -c


jdholtz:
Why does the comment say `Remove after 2023-10-01`? Should this contribution be removed now?

Jguer:
@jdholtz Yes, this can be removed now, it was just to give a period of deprecation

jdholtz:
@Jguer just made a small PR to remove that code.

PR #2190, opened 2023-05-24
Fix deps resolving for installs from SRCINFO

smolx:
Fixed in the same way as for recent fix of AUR dependencies(#2157).
Also added new test case and slightly fixed a few old ones.

### DIFF
--- a/local_install.go
+++ b/local_install.go
@@ -68,10 +68,7 @@ func installLocalPKGBUILD(
 		return errors.New(gotext.Get("no target directories specified"))
 	}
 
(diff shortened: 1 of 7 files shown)


smolx:
@Jguer  I had to redo the fix, because previously I did not take into account a case with several .SRCINFO targets. It turned out to be a little more complicated, but still it's tested and works. I also made MockRunner(and MockBuilder just in case) thread-safe to make a new test-scenario work.

Jguer (approved):
Let's go :), great PR

PR #2191, opened 2023-05-24
fix(pkgbuild): convert package name for gitlab URLs

moson-mo:
Some package(base) names contain characters that are not allowed in gitlab repository names.
Arch converted them according to the following rules:

https://gitlab.archlinux.org/archlinux/devtools/-/blob/6ce666a1669235749c17d5c44d8a24dea4a135da/src/lib/api/gitlab.sh#L84

### DIFF
--- a/pkg/download/abs.go
+++ b/pkg/download/abs.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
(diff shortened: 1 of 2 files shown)


Jguer:
Nice @moson-mo, thank you very much

Jguer (approved):

PR #2195, opened 2023-05-26
Fix install reason preservation when package is reinstalled

### LINKED ISSUE #2164: Not preserving of installation reasons when package is reinstalled
### Affected Version

`yay v12.0.4.r23.gec15a5b - libalpm v13.0.2`

### Describe the bug

Yay doesn't preserve the installation reason of reinstalled package. As said in [pacman wiki](https://wiki.archlinux.org/title/Pacman#Installation_reason), "when reinstalling a package, the current installation reason is preserved by default".
This unexpected behavior is present for at least AUR and Repo packages.

### Reproduction Steps And Output

```sh
$ yay -S --asdeps autojump
...
$ yay -Qi autojump | grep Reason
Install Reason  : Installed as a dependency for another package
$ yay -S autojump
...
$ yay -Qi autojump | grep Reason
Install Reason  : Explicitly installed
```

### Expected behavior

Install reason is preserved.

smolx:
First approach. Maybe a little messy and bloated in `GraphFromTargets` method. There is one side-effect: when installing packages, that previously were installed as dependencies, they are being showed as dependency at the preview section(where packages counts are). Is it okay?
 
Fixes #2164.

### DIFF
--- a/local_install_test.go
+++ b/local_install_test.go
@@ -138,6 +138,7 @@ func TestIntegrationLocalInstall(t *testing.T) {
 
 			return nil
 		},
+		LocalPackageFn: func(s string) mock.IPackage { return nil },
 	}
 
 	config := &settings.Configuration{
@@ -258,6 +259,7 @@ func TestIntegrationLocalInstallMissingDep(t *testing.T) {
 
 			return nil
 		},
+		LocalPackageFn: func(string) mock.IPackage { return nil },
 	}
 
 	config := &settings.Configuration{
@@ -557,6 +559,7 @@ func TestIntegrationLocalInstallGenerateSRCINFO(t *testing.T) {
 
 			return true
 		},
+		LocalPackageFn: func(string) mock.IPackage { return nil },
 		SyncSatisfierFn: func(s string) mock.IPackage {
 			switch s {
 			case "dotnet-runtime>=6", "dotnet-runtime<7":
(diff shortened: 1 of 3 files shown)


Jguer (approved):
Seems fine, I don't see a simpler way of doing it. Will merge as soon as it's not conflicting :smiley:

smolx:
@Jguer, I'm not sure that this PR is "next"-ready. I found out that `--asexplicit` option doesn't work for repo-packages properly either without or with this fix. Should that be fixed with this PR too?

smolx:
Also, if the general idea of this fix is acceptable, I can try to make the `GraphFromTargets` method less bloated with refactoring/little changes. Specifically, in the **default case(specified DB)** of `switch target.DB`, we can make sure that a `target.DB/target.Name` package exists and do almost the same stuff as in case of **unspecified DB**. What do you say on that?

smolx:
@Jguer, I've tried to implement my latest suggestion(on `GraphFromTargets` refactoring and making its default switch case similar to "unspecified db" case). But it went down the rabbit hole. I even had to extend the **go-alpm** library a bit to make things in a correct way. And there is more work to do. If you are interested, I would continue and present it in a **separate PRs** when work is done.

As for **this PR**, I've pushed the final non-refactored version which is ready to be reviewed and merged.

Jguer (approved):
LGTM, I'll start tidying up for a release as a minor for the end of the month.

I'm curious to see your proposed changes to `go-alpm`, I think it has a lot of improvement potential

PR #2196, opened 2023-05-26
Remove redundant attempts to set install reasons

### LINKED ISSUE #2167: yay wants to set install reason for non-installed syncdeps
### Affected Version

yay v12.0.4 - libalpm v13.0.2

### Describe the bug

When a package needs sync (repo) dependencies, yay/pacman asks to confirm their installation. If user refuses, yay still wants to set an install reason for these dependencies (and fails to do so). This seems to be #2137 for all packages, not only groups.

### Reproduction Steps

1. Try to install an AUR or a repo package _xxx_ requiring syncdeps with `yay -Syu xxx`.
2. yay/pacman asks to confirm syncdeps' installation with `:: Proceed with installation? [Y/n]`.
3. If user refuses, yay still wants to set an install reason for these dependencies (and fails to do so).

### Expected behavior

yay doesn't try to do anything after an installation is cancelled.

### Output

```sh
$ yay -Syu xxx
...
:: Proceed with installation? [Y/n] n
error: could not set install reason for package xxx (could not find or read package)
 -> error installing repo packages

```

smolx:
Redundant attempts took place when the installation by pacman was exited with error code, i.e. an real error occurred or installation was simply cancelled.

Straightforward quick fix. Maybe there is some pitfall, that I recklessly fell into.

Fixes #2167.

### DIFF
--- a/aur_install.go
+++ b/aur_install.go
@@ -426,6 +426,9 @@ func (installer *Installer) installSyncPackages(ctx context.Context, cmdArgs *pa
 
 	errShow := installer.exeCmd.Show(installer.exeCmd.BuildPacmanCmd(ctx,
 		arguments, installer.targetMode, noConfirm))
+	if errShow != nil {
+		return errShow
+	}
 
 	if errD := asdeps(ctx, installer.exeCmd, installer.targetMode, cmdArgs, syncDeps.ToSlice()); errD != nil {
 		return errD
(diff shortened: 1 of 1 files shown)


Jguer:
I think it's fine, there might be a pitfall there for sure, but it seems to make sense

Jguer (approved):

PR #2199, opened 2023-05-30
Add --askyesremovemake option

### LINKED ISSUE #1531: Changing default answer
When installing a package, if the config has `"removemake": "ask"`, then the default answer is "N".

```
==> Remove make dependencies after install? [y/N] 
```

Is there a way to make the default answer "Y"? I know that we can automatically pick yes by making the config `"removemake": "yes"`, but this suppresses the question.

My understanding is there are three options. From the `man` page:

       --askremovemake
              Ask to remove makedepends after installing packages.

       --removemake
              Remove makedepends after installing packages.

       --noremovemake
              Do not remove makedepends after installing packages.

It would be great if there were an option to `--askremovemake`, but with "Y" as default.

### Describe alternatives you've considered
None that I can think of. I guess just picking `removemake` and manually overriding it when necessary.

smolx:
Same as --askremovemake option but with "Y" as a default answer.

Closes #1531.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -95,6 +95,7 @@ Permanent configuration options:
     --nodiffmenu          Don't show diffs for build files
     --noeditmenu          Don't edit/view PKGBUILDS
     --askremovemake       Ask to remove makedepends after install
+    --askyesremovemake    Ask to remove makedepends after install("Y" as default)
     --removemake          Remove makedepends after install
     --noremovemake        Don't remove makedepends after install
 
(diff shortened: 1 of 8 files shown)


Jguer:
I'm not opposed to implementing this as a new cmd option. Thought it could be better to remove all those existing askcmds and turn them into config options as they're not normally changed.

Jguer (approved):

smolx:
If I understand correctly, there is no documentation for config options and values at all. And the only documented way of config editing is through the command options. In short, creating documentation for a config-file is quite a different story. I am not ready for this yet.

Jguer (approved):

Jguer:
Merged :smile: , I am hoping to get some time to write specifications on migration to a TOML based config file, this would be much friendlier for user edits and storage.

PR #2203, opened 2023-06-03
Fix excluding of packages with unsatisfied deps

### LINKED ISSUE #2127: unable to continue build if an excluded package requires a higher/missing package than installed
### Affected Version

yay v12.0.4.r6.g64bb346 - libalpm v13.0.2

### Describe the bug

unable to build any aur packages with a "Could not find required packages" if any of the other aur packages require a higher version of a package than is available even if these effected packages are excluded.

### Reproduction Steps

1. have an aur package that requires a higher version of a package
2. exclude package
3. package fails with "Could not find required packages"

### Expected behavior

the process continues with the packages beginning to build

### Output

output log: https://paste.ee/p/cQyns

smolx:
When dependency is unsatisfied, add to the graph not only a dep node,
but relationship with parent too.

Tried to fix #2127 this way.

Does it look OK? Should I also add a test scenario for this case?

### DIFF
--- a/pkg/dep/dep_graph.go
+++ b/pkg/dep/dep_graph.go
@@ -652,7 +652,9 @@ func (g *Grapher) addNodes(
 	for _, depString := range targetsToFind.ToSlice() {
 		depName, mod, ver := splitDep(depString)
 		// no dep found. add as missing
-		graph.AddNode(depName)
+		if err := graph.DependOn(depName, parentPkgName); err != nil {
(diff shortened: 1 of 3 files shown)


Jguer:
Hey @smolx , yeah, it does make sens, if it's easy to put up a test case for it, it would be great, it's an easy to miss scenario

Jguer (approved):

PR #2205, opened 2023-06-05
Add a package update command

adivardhan1000:
when running these commands on archlinux-base proxmox lxc, found out that these commands return 404 error unless the pacman -Sy command is run.

This although seeming simplistic might be something unknown when someone uses yay and archlinux for the first time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,7 @@ building with makepkg:
 We start with updating the package lists and make sure we have the `base-devel` package group installed.
 
 ```sh
-pacman -Sy
-pacman -S --needed git base-devel
+pacman -Sy --needed git base-devel
 git clone https://aur.archlinux.org/yay.git
 cd yay
 makepkg -si
--- a/README.md
+++ b/README.md
@@ -35,9 +35,10 @@ If you are migrating from another AUR helper, you can simply install Yay with th
 The initial installation of Yay can be done by cloning the PKGBUILD and
 building with makepkg:
 
-Before you begin, make sure you have the `base-devel` package group installed.
+We start with updating the package lists and make sure we have the `base-devel` package group installed.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,9 +35,10 @@ If you are migrating from another AUR helper, you can simply install Yay with th
 The initial installation of Yay can be done by cloning the PKGBUILD and
 building with makepkg:
 
-Before you begin, make sure you have the `base-devel` package group installed.
+We start with updating the package lists and make sure we have the `base-devel` package group installed.
 
 ```sh
+pacman -Sy
 pacman -S --needed git base-devel
```
Jguer:
```suggestion
pacman -Sy --needed git base-devel
```

How about this instead if we're asking the users to update anyway?

Jguer:
Thanks for the PR, I've made a small suggestion to keep it concise

Jguer:
Hey @adivardhan1000 , you'll need to commit or modify your patch locally to include the suggestions I added as I can't push to your branch. Once that is added I'll merge :)

adivardhan1000:
@Jguer Where can I see the suggestions you have made? I don't see any remarks

Jguer:
@adivardhan1000 you should be able to see it here 

https://github.com/Jguer/yay/pull/2205/files/0f10ed7029085ac3c37874702e23ea14f9df6c8f

smolx:
@Jguer , Just want to say, that I don't see any suggestions there too. Maybe some odd permissions are in effect.

Jguer:
image

ok that's weird, let me try to put it again, but this is basically the change

Jguer (approved):
Thanks for your contribution @adivardhan1000

PR #2211, opened 2023-06-12
Fixed now user can cancel installation no error

### LINKED ISSUE #2166: Installation cancellation produces an error (while interruption doesn't)
### Affected Version

yay v12.0.4 - libalpm v13.0.2

### Describe the bug

When user cancels an installation almost anytime during yay/pacman/makepkg operations with `SIGINT`, yay exits gracefully and no error is produced, even though the system might actually be left in an inconsistent state.

However, if user simply refuses a pacman operation in a dialogue (therby cancelling yay operation(s)), an error is produced.

### Expected behavior

yay raises an error with `SIGINT`, but not when user simply refuses.

### Output

```sh
:: Proceed with installation? [Y/n] n
 -> error installing repo packages
```

Claudio-code:
Tried to fix https://github.com/Jguer/yay/issues/2166

Now intalation verify if error is exitError in that case not return error

Screenshot from 2023-06-12 13-24-19


Jguer:
I'm divided on this, in my view the user already decided to install on the previous upgrade menu, any cancellation after is an error.

what do you think @smolx ?

Claudio-code:
In my view if the user cancels an action of the program this is not an error, it is an expected behavior.

smolx:
I'm not sure I have sufficient competence, but from what I see, yay can invoke pacman multiple times because of layered installation process, therefore only the first layer can safely be cancelled in a way of this PR(without error). Later cancellations better to be reported as errors because of "inconsistent" state. Is it worth to suppress that error for the first layer only? I don't know. It depends on @Claudio-code. :)

Claudio-code:
The reason I tried to resolve this issue is that this behavior was reported as a bug in one issue and referenced as a bug in another. So it's something that bothers the community, I guess it's because it's a different behavior from other package managers, which would justify the community's annoyance. In my view, this pr solves this pain :relaxed:

Claudio-code:
if it is not in the interest of the project, you can close the pr and the issues

PR #2218, opened 2023-06-21
Fix handling targets with specified db

smolx:
Handle it in a similar way to handling targets with unspecified db.

Also refactored GraphSyncPkg method to make stuff more DRY.

It doesn't compile because it requires changes to be made in **go-alpm**: [https://github.com/Jguer/go-alpm/pull/62](https://github.com/Jguer/go-alpm/pull/62). I think it's good to have both PRs to see the full picture.


Jguer:
:+1: looks good, tried some scenarios locally, let's merge it and go from there :)

Jguer (approved):

smolx:
Oops. It was merged without many words. I expected some remarks or discussion. :)
OK then. I'll raise one question here myself.
While working on this I've noticed dubious behavior. When user tries to install several targets and one of them is not found(for example user makes a typo), yay just skips it with non-fatal error message("AUR package not found", even if user implied a repo target) and proceeds to install other targets. Inattentive user may get an undesirable result. Is it intended feature? Wouldn't it be better to show all error messages for not found targets and just exit like for example pacman does?

Jguer:
> Wouldn't it be better to show all error messages for not found targets and just exit like for example pacman does

I agree with this statement. There was a lot of feedback though on how `yay` should ignore failed builds and not found targets specially from users that maintain a list of AUR packages to install on newly bootstrapped systems.

I see merit to both arguments

PR #2234, opened 2023-06-29
Don't upgrade repo packages when --aur is specified

### LINKED ISSUE #2149: Running with --aur targets repo packages
### Affected Version
yay v12.0.4 - libalpm v13.0.2

### Describe the bug
Even with `--aur` Yay proposes to update repository packages. I believe this happens when a repo dependency needs to be installed before AUR build and when Pacman is ran it wants to update everything.

### Reproduction Steps

1. Run an update with Yay, limited to AUR packages with `--aur`
2. In case a dependency from repository needs to be installed, updates for all packages are proposed alongside.

### Expected behavior
Only the required dependencies are installed in this step.

jdholtz:
Fixes #2149.

If the target mode is AUR (i.e. `--aur` or `-a` were passed as arguments), the `-u` and `--upgrades` flags will be removed to ensure that Pacman doesn't run upgrades when installing dependencies.

Additionally, I added a few tests to ensure this functionality works correctly.

### DIFF
--- a/aur_install.go
+++ b/aur_install.go
@@ -420,6 +420,12 @@ func (installer *Installer) installSyncPackages(ctx context.Context, cmdArgs *pa
 	arguments.Op = "S"
 	arguments.ClearTargets()
 	arguments.AddTarget(repoTargets...)
+
+	// Don't upgrade all repo packages if only AUR upgrades are specified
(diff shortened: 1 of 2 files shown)


Jguer (approved):
Thanks for the patch @jdholtz , merging it in

PR #2239, opened 2023-07-06
fix(aur): only check gpg signature after gpg import . fixes #2165

### LINKED ISSUE #2165: yay imports new PGP key after failed signature check
### Affected Version

yay v12.0.4 - libalpm v13.0.2

### Describe the bug

When installing an AUR package that needs importing of a new PGP key, yay/pacman fails the signature checking at first (because of a missing key), but then continues on with the installation (regardless of an integrity error) and only then imports the relevant PGP key, successfully builds the package and installs it.

### Reproduction Steps

1. Run `yay -Syu ffmpeg-headless` (for example)
2. yay downloads the PKGBUILD, processes it and starts making package
3. It fails in the source verification step - see _Output_ below

### Expected behavior

PGP key importing is done before any integrity checks.

### Output

```sh
$ yay -Syu ffmpeg-headless
...
==> Verifying source file signatures with gpg...
    ffmpeg git repo ... FAILED (unknown public key B18E8928B3948D64)
==> ERROR: One or more PGP signatures could not be verified!
 -> error downloading sources: .../yay/build/ffmpeg-headless
         context: error downloading sources: .../yay/build/ffmpeg-headless
         context: exit status 1

:: Remove make dependencies after install? [y/N]
:: (1/1) Parsing SRCINFO: ffmpeg-headless
gpg: error reading key: No public key

:: PGP keys need importing:
 -> DD1EC9E8DE085C629B3E1846B18E8928B3948D64, required by: ffmpeg-headless
:: Import? [Y/n]
:: Importing keys with gpg...
gpg: key B18E8928B3948D64: public key "Michael Niedermayer (key used for git commits) <michael@niedermayer.cc>" imported
gpg: Total number processed: 1
gpg:               imported: 1
 core is up to date
 extra is up to date
 community is up to date
resolving dependencies...
looking for conflicting packages...

(installation succeeds afterwards)
```

Jguer:
only check gpg signature after gpg import

fixes #2165

### DIFF
--- a/aur_source.go
+++ b/aur_source.go
@@ -32,7 +32,7 @@ func (e *ErrDownloadSource) Unwrap() error {
 func downloadPKGBUILDSource(ctx context.Context,
 	cmdBuilder exe.ICmdBuilder, pkgBuildDir string, installIncompatible bool,
 ) error {
-	args := []string{"--verifysource", "-Ccf"}
+	args := []string{"--verifysource", "--skippgpcheck", "-Ccf"}
 
 	if installIncompatible {
 		args = append(args, "--ignorearch")
--- a/aur_source_test.go
+++ b/aur_source_test.go
@@ -54,7 +54,7 @@ func Test_downloadPKGBUILDSource(t *testing.T) {
 	cmdBuilder := &TestMakepkgBuilder{
 		parentBuilder: &exe.CmdBuilder{MakepkgConfPath: "/etc/not.conf", MakepkgFlags: []string{"--nocheck"}, MakepkgBin: "makepkg"},
 		test:          t,
-		want:          "makepkg --nocheck --config /etc/not.conf --verifysource -Ccf",
+		want:          "makepkg --nocheck --config /etc/not.conf --verifysource --skippgpcheck -Ccf",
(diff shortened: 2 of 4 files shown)


Jguer:
@jdholtz & @smolx, yay works even when removing `downloadPKGBUILDSourceFanout` (preparer.go:229), it would just lose concurrent downloading of pkgbuild sources and sources wouldn't be available for pre install hooks if they were to be added. 

What's your opinion on PKGBUILD source downloading before install? Do you find the concurrency useful?

jdholtz:
> What's your opinion on PKGBUILD source downloading before install? Do you find the concurrency useful?

I definitely see it being useful as it can significantly speed up the total time spent downloading large AUR packages, especially during a full system upgrade. Are there any other disadvantages to removing this functionality besides the need to maintain it?

smolx:
I don't have a strong opinion on this matter. Personally I rarely see more than 2-3 parallel downloads, so I can't really feel the effect. But definitely there are users with high speed internet connection who use AUR more extensively. They feel the maximum effect.
I also think that downloading of multiple targets in parallel is de facto standard now. Perhaps better to have it than not to have.

PR #2244, opened 2023-07-16
perf: smaller git clone

ragnarok22:
At the end, we don't need the commit history or others branchs to install an aur package. The git clone can be smaller and faster using some flags like:

    git clone --depth=1 --single-branch


Jguer:
Hey @ragnarok22 , this change shouldn't make much difference as AUR repos are relatively small and it would cause complications when diffing. If you have some pratical cases where this helps or benchmarks executed please share and we can reopen this.

I'm linking the last request for this https://github.com/Jguer/yay/pull/935

PR #2272, opened 2023-09-07
feat(install): add --keepsrc to keep pkg/ and src/ directories

### LINKED ISSUE #798: Is it possible to keep the src directory?
When I update a package from a `git` repository, the whole package is rebuilt every time. Is it possible to keep the `src` directory and to recompile only what needs to be recompiled? (Using `makepkg -f`, maybe)

jdholtz:
Should fix #798. 

Providing this flag during installation of AUR packages allows for keeping the src/ and pkg/ directories produced my makepkg. If the user wants to delete the directories, they can either select to cleanBuild in the cleanmenu or run the installation without the `--nocleanbuild` flag. 

The one inconsistency I found when running Yay without the `--nocleanbuild` flag (and if pkg/ and src/ directories previously existed), since Yay runs `pacman -U` separately from makepkg, Yay will not clean the directories if the package is cached already whereas makepkg will. I don't think this is a big deal since Yay provides the cleanmenu, but we could discuss solutions to fix this inconsistency if you want.

I added tests in both places that make this check.

I opted not to add a `--cleanbuild` option that would override the option in the config file as I thought it would be confusing due to it being a makepkg flag. However, I can add it if desired. Also, I can change the name of the flag (paru uses `--keepsrc` but we could think of a different name to use as well).

Last, I exposed the `CleanBuild` attribute in the `cmdBuilder` as that is where I thought it was the easiest to access it. I'm not super familiar with Go so please let me know if it could be refactored to be put in a more logical place.


sevmonster:
This is absolutely critical for building large packages like kernels, where a failed build step (for whatever reason) can really take a lot of time to get back to.

Jguer:
Hey @jdholtz , I think we can merge this, let try and fix the CI.

I think `keepsrc` is a better flag name to expose to the user (and if paru already uses it, it's more familiar)

jdholtz:
> I think keepsrc is a better flag name to expose to the user (and if paru already uses it, it's more familiar)

Sounds good. Do you want me to go with `keepsrc` and nokeepsrc` to have both options then?

Jguer:
I think `keepsrc` is enough.
(In the future we should look at allowing 'real' boolean flags for this ex: --keepsrc=false)

Morganamilo:
Would be good. Though =yes and =no I think is nicer.

jdholtz:
>  (In the future we should look at allowing 'real' boolean flags for this ex: --keepsrc=false)

Yes, that would be nice. Many of the options prefixed with `no` can be removed and replaced with this format.

> Though =yes and =no I think is nicer.

There could be multiple booleans used, like "yes", "true", and "y" for true and "no", "false", and "n" for false.

Jguer (approved):

Jguer:
merging :)

@jdholtz would you be interested in implementing this boolean flag system?

We'd just need to keep old toggles working for a while

Jguer:
sorry @jdholtz , ended forgetting to remove the coauthored line from the merge message, if it's important for you I'll revert and recommit with the proper author information

jdholtz:
> would you be interested in implementing this boolean flag system?

Yes, I can. 

> it's important for you I'll revert and recommit with the proper author information

No worries, you can keep it as is.

PR #2285, opened 2023-09-18
feat(yay): add boolean flags

jdholtz:
This feature allows users to specify `--option=bool` instead of using `--option` and `--nooption`. Specifying nothing (or a `bool` value that isn't falsy) results in the boolean value being true.

A few notes on this PR:
1. I didn't add documentation in the manpage yet because I wasn't sure how to present it to users. Suggestions would be helpful
2. I marked the deprecated features with a `DEPRECATED` comment, but I can use a more robust deprecation system (such as a warning), if desired.
3. These options do not work for pacman/makepkg options (except for `debug` and `noconfirm`). I can add this but it may make more sense for users to only use these boolean flags for yay options.
4. The boolean flags only work with an equal sign unlike other flags that have values. I can switch it to allow `--option <bool>` too.

### DIFF
--- a/doc/yay.8
+++ b/doc/yay.8
@@ -293,6 +293,9 @@ Unset the answer for the upgrade menu.
 Show the clean menu. This menu gives you the chance to fully delete the
 downloaded build files from Yay's cache before redownloading a fresh copy.
 
+If 'cleanmenu' is enabled in the configuration file, you can temporarily disable it by
+using '--cleanmenu=false' on the command line
+
 .TP
 .B \-\-diffmenu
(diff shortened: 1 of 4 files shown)


Jguer:
your reasoning makes a lot of sense.

1. I think adding an example to the manpage/readme should be enough to show the 'yes, this works like this' for 1 or 2 of the affected options

2. I think you can inject a log warn when parsing the deprecated options

3. Agree

4. I think we can leave it as is for now

Jguer (approved):
Thanks for taking the time to do this @jdholtz

PR #2291, opened 2023-09-24
refactor: remove redundant `len` check

Juneezee:
`len` returns 0 if the slice or map is nil. From the Go specification [^1]:

> "1. For a nil slice, the number of iterations is 0."
> "3. If the map is nil, the number of iterations is 0."

Therefore, an additional `len(v) != 0` check for before the loop is unnecessary.

[^1]: https://go.dev/ref/spec#For_range


Jguer (approved):
thanks for the improvement @Juneezee

PR #2347, opened 2024-01-23
Remove deprecated warning when using -yc together

jdholtz:
This PR removes the deprecation warning that users got when using `-yc` instead of `-Yc`. More information found on #2181.


jdholtz:
@Jguer I can also remove the options that were deprecated in #2285 as the Boolean flags were a part of v12.2.0.

Jguer (approved):

Jguer:
thanks for the PR @jdholtz , if you want to also remove the options from https://github.com/Jguer/yay/pull/2285 go ahead :)

PR #2350, opened 2024-01-25
Remove deprecated flags in favor of boolean flags

jdholtz:
This removes the deprecated flags from #2285. The deprecated messages are still in the `*.po` files. I can remove those manually, but is there a better way to do that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ pacman -S --needed git base-devel yay
 
 - **Yay is not asking me to edit PKGBUILDS, and I don't like the diff menu! What can I do?**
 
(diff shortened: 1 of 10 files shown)


Jguer (approved):

Jguer:
done in https://github.com/Jguer/yay/pull/2351

you can install xgotext
```
go install github.com/leonelquinteros/gotext/cli/xgotext@latest
```

and then run

```sh
xgotext -in . -out po
mv po/default.pot po/en.po
```

the other ones will be updated by transifex automatically

PR #2386, opened 2024-03-06
fix: added missing spaces

Absobel:
image

There were missing spaces there, so I added them to make it pretty.

### DIFF
--- a/pkg/dep/topo/errors.go
+++ b/pkg/dep/topo/errors.go
@@ -3,7 +3,7 @@ package topo
 import "errors"
 
 var (
-	ErrSelfReferential  = errors.New("self-referential dependencies not allowed")
-	ErrConflictingAlias = errors.New("alias already defined")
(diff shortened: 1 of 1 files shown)


Jguer:
Hey @Absobel is it possible to do this formatting where the errors are output instead?

Absobel:
I forgot I made this, I just had the problem again.
Anyway, would it be better to modify al the calls to `Warnln` following a call to `DependOn` ? (it does that only for circular dependencies tho)

```
if err := graph.DependOn(depName, parentPkgName); err != nil {
g.logger.Warnln(depString, " ", parentPkgName, " ", err)
}
```

But its strange because the calls to Errorln or things like that don't seem to have this problem. And I don't know how to reliably test that because I would need to find a packages with circular dependencies or smth

Don't hesitate to tell me if I'm missing something totally obvious, I'm not used to make contributions

Jguer (approved):
Let's merge this one for now

PR #2400, opened 2024-03-16
Replace github.com/pkg/errors with stdlib errors/fmt

JRaspass:
https://github.com/pkg/errors is archived and effectively sunsetted, see https://github.com/pkg/errors/issues/245 for details. The standard library suffices nowadays.


Jguer (approved):
thanks for fixing this

PR #2405, opened 2024-03-19
Use Go 1.21's min/max built-ins

JRaspass:
This simplifies the code compared to either rolling our own or awkwardly using math's float functions with integers.


Jguer (approved):
thanks for the contribution @JRaspass

PR #2412, opened 2024-03-25
Fixed a few translations in fr_FR.po

barsanuphe:
Fixed a few translation strings (missing %s, grammar)


Jguer:
Hey @barsanuphe , I think your changes will be overridden if they're not contributed through transifex https://www.transifex.com/yay-1/yay/

barsanuphe:
OK thx for the info @Jguer, hopefully I made the equivalent fixes on transifex.

Jguer:
:+1: thanks @barsanuphe , the PR #2414 should include those changes

PR #2437, opened 2024-04-22
Fix incorrect formatting of some msgstr of the spanish translation

Tom5521:
More specifically for "%s to upgrade/install." and "%s will also be installed for this operation."


Jguer:
Hi Tom, can you submit this fix through transifex? I can merge your PR but it will probably be overridden by transifex after

Jguer (approved):

Tom5521:
> Hi Tom, can you submit this fix through transifex? I can merge your PR but it will probably be overridden by transifex after

Ok, thanks for the clarification

PR #2460, opened 2024-06-22
Respect provided targets when using -Si flag

jdholtz:
This adds more to the fix in #2459 to work with the `-a (--aur)` and `--repo` flags. Checks to make sure each target is specifically (through the `--aur` or `--repo` flags) or implicitly (none of those flags are specified) specified before filling the `aurS` and `repoS` slices.


Jguer (approved):
Fair point :+1: thanks for the PR

PR #2462, opened 2024-06-25
feat(su): add run0 to the list of alternative privilege elevators

sylv-io:
systemd 256 introduces the run0 based on systemd-run, as an alternative to sudo/doas without being a SUID binary.

TODO:
- [ ] Add config tests
- [ ] Add `--background=` flag to prevent red background color

### DIFF
--- a/pkg/settings/config.go
+++ b/pkg/settings/config.go
@@ -179,7 +179,7 @@ func (c *Configuration) setPrivilegeElevator() error {
 	c.SudoFlags = ""
 	c.SudoLoop = false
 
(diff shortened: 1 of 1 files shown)


sylv-io:
Greetings,
With ArchLinux now using systemd 256, it is time to support `run0`.
I added it to the list after `doas`, but imho it should be prioritized since it is available by default with the `systemd` package.

Any thoughts?

Jguer:
I agree, I think run0 support should be better :+1:, I should make more of an effort to use it myself

PR #2483, opened 2024-08-07
Add sudo to install instructions so it can be run without user switching

Sleuth56:
Currently you have to either su to root to run pacman or manually prepend sudo to it. This fixes that.
Because makepkg doesn't let you run as root su-ing to root doesn't work so sudo-ing the pacman command makes it easier for a normal user.


Jguer (approved):

Jguer:
@Sleuth56 can you make the same change to the chain command and add the disclaimer before to switch sudo to the user's preferred alternative such as run0 or doas?

Issue #2506, opened 2024-09-17
Build error on 'msgfmt' syntax error


tijko:
### 12.4.1.r0.g3a118b76-1

1. git clone https://aur.archlinux.org/yay-git.git
2. cd yay-git makepkg -si

### Running makepkg -si
```
msgfmt po/en.po -o po/en.mo
po/en.po:368: missing 'msgstr' section
po/en.po:368:36: syntax error
msgfmt: found 2 fatal errors
make: *** [Makefile:134: po/en.mo] Error 1
==> ERROR: A failure occurred in package().
```

Having used this a fair amount I am familiar with how the 'prompt menu' looks so I just copied the line style from [line 372 in en.po](https://github.com/Jguer/yay/blob/84d8f1b7b3bc7448f24abd57f48ff0e9dfbdb388/po/en.po#L372)

Jguer:
fixed by #2507

PR #2507, opened 2024-09-17
Fixes syntax error #2506

tijko:
https://github.com/Jguer/yay/issues/2506

### DIFF
--- a/po/en.po
+++ b/po/en.po
@@ -365,7 +365,7 @@ msgid "Packages to exclude"
 msgstr ""
 
 #: pkg/upgrade/service.go:294
-msgid "Packages to exclude: (eg: "1 2 3", "1-3", "^4" or repo name)"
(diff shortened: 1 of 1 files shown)


radiooperator-19 (changes requested):
Do not remove quotation marks; just use `\"` like other .po files

tijko:
> Do not remove quotation marks; just use `\"` like other .po files

~~I actually did that in my original change but reverted when I noticed in [en.po](https://github.com/Jguer/yay/blob/84d8f1b7b3bc7448f24abd57f48ff0e9dfbdb388/po/en.po#L372)~~ I see the overall format now.

I made the change though.

radiooperator-19 (approved):

Jguer (approved):
seems there's an issue with xgotext, I will check out what's happening there

PR #2520, opened 2024-10-11
Update README.md to include propper install with prequisites and binaries

VoltagedDebunked:
A small fix, but would fix confusion (probably)


Jguer:
Hey @VoltagedDebunked, normally the makepkg flags will install make dependencies. Installing them like this will set them as explicitly installed which might not be the users' intentions

VoltagedDebunked:
Ok ty

PR #2523, opened 2024-10-17
fix swapped fish completion descriptions for --repo and --aur

anmolw:
Just a small fix for both the long and short forms of --repo and --aur having their descriptions swapped in the fish completions.


Jguer (approved):
Thanks for your contribution @anmolw

PR #2525, opened 2024-10-23
Add missing sudo to binary install instructions

non-bin:


Jguer (approved):

Jguer:
Thanks for your contribution @non-bin

PR #2531, opened 2024-11-12
Replace "⚠️ {warning}" with Github Warning Blocks

arithefirst:
This pull requests took the warnings in the README (Formatted as ⚠️ {warning here}) and switched them out for the more universal Github warning block (see below) since some browsers may not render the emojis correctly

**Example warning block**
> [!WARNING]
> {warning here}


Jguer:
That's pretty nice, thanks @arithefirst

Jguer (approved):

PR #2543, opened 2024-12-03
socks5 support

dmitrodem:
Golang implementation of `http.Client` does not handle socks5 support via LD_PRELOAD (a.k.a. tsocks/proxychains). Proposed patch adds explicit socks5 support via environment variable, e.g. `SOCKS5_PROXY=localhost:1080 yay -Ss vivado`.

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -41,7 +41,7 @@ func NewRuntime(cfg *settings.Configuration, cmdArgs *parser.Arguments, version 
 	logger := text.NewLogger(os.Stdout, os.Stderr, os.Stdin, cfg.Debug, "runtime")
 	runner := exe.NewOSRunner(logger.Child("runner"))
 
-	var transport = &http.Transport{}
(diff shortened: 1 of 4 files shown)


pkg/runtime/runtime.go
```diff
@@ -39,10 +41,20 @@ func NewRuntime(cfg *settings.Configuration, cmdArgs *parser.Arguments, version
 	logger := text.NewLogger(os.Stdout, os.Stderr, os.Stdin, cfg.Debug, "runtime")
 	runner := exe.NewOSRunner(logger.Child("runner"))
 
+	var transport = &http.Transport{}
```
Jguer: Can you copy the default transport? I think this works
```go
http.DefaultTransport.(*http.Transport).Clone() 
```

dmitrodem: of course it is possible. I'm not an expert in Go at all :)

Jguer (approved):

PR #2561, opened 2025-01-05
fix(installer): Fix the same pkgbase being built multiple times on error

### LINKED ISSUE #2560: `yay` tries to build each split package, ignoring previous failures in the same `pkgbase`
### Affected Version

`yay v12.4.2 - libalpm v15.0.0`

### Describe the bug

`yay` tries to build each split package in the same run, even if attempt to build the same `pkgbase` was failed.

### Reproduction Steps

I discovered this bug during upgrade with `--devel`, but it's also reproducible with new installation.

These packages are currently failing, but I'm unsure how long it will take upstream for to fix them.

```sh
yay -S linux-drm-tip-git linux-drm-tip-headers-git linux-drm-tip-docs-git
```
Or 
```sh
yay -S linux-drm-intel-next-git linux-drm-intel-next-headers-git linux-drm-intel-next-docs-git
```

`yay` makes 3 build attempts during install or upgrade.

### Expected behaviour

`yay` should make exactly one build attempt per `pkgbase`.

### Output

I uploaded [debug.log](https://gist.github.com/yurikoles/79a07357147eafb8cccddad113988206) to gist due to extreme size.

```sh
$ yay -Pg

{
	"buildDir": "/home/yurikoles/.cache/yay",
	"editor": "",
	"editorflags": "",
	"makepkgbin": "makepkg",
	"makepkgconf": "",
	"pacmanbin": "pacman",
	"redownload": "no",
	"answerclean": "no",
	"answerdiff": "no",
	"answeredit": "no",
	"answerupgrade": "yes",
	"gitbin": "git",
	"gpgbin": "gpg",
	"gpgflags": "",
	"mflags": "",
	"sortby": "votes",
	"searchby": "name-desc",
	"gitflags": "",
	"removemake": "yes",
	"sudobin": "sudo",
	"sudoflags": "",
	"version": "11.3.2",
	"requestsplitn": 150,
	"completionrefreshtime": 1,
	"maxconcurrentdownloads": 0,
	"bottomup": true,
	"sudoloop": false,
	"timeupdate": false,
	"devel": false,
	"cleanAfter": false,
	"keepSrc": false,
	"provides": false,
	"pgpfetch": true,
	"cleanmenu": false,
	"diffmenu": false,
	"editmenu": false,
	"combinedupgrade": true,
	"useask": false,
	"batchinstall": false,
	"singlelineresults": false,
	"separatesources": true,
	"debug": false,
	"rpc": true,
	"doubleconfirm": true,
	"rebuild": "no"
}
```

Ferdi265:
The previous fix commit ec837c8 failed to address the case where the build fails, as packages are only added to builtPkgDests on a successful build. This commit addresses this by adding the package to the map earlier.

Fixes #2560.

The patch seems trivial and initial testing shows it works and I've not run into any issues with this patch yet, but I'll daily drive this for the next week anyway to see if anything comes up.

### DIFF
--- a/pkg/sync/build/installer.go
+++ b/pkg/sync/build/installer.go
@@ -240,6 +240,7 @@ func (installer *Installer) installAURPackages(ctx context.Context,
 			installer.log.Debugln("building pkgbase", base, "package", name)
 			pkgdests, errMake = installer.buildPkg(ctx, dir, base,
 				installIncompatible, cmdArgs.ExistsArg("needed"), aurOrigTargetBases.Contains(base))
+			builtPkgDests[base] = pkgdests
 			if errMake != nil {
 				if !lastLayer {
 					return fmt.Errorf("%s - %w", gotext.Get("error making: %s", base), errMake)
@@ -249,8 +250,6 @@ func (installer *Installer) installAURPackages(ctx context.Context,
 				installer.log.Errorln(gotext.Get("error making: %s", base), "-", errMake)
 				continue
 			}
-
-			builtPkgDests[base] = pkgdests
 		}
 
 		if len(pkgdests) == 0 {
--- a/pkg/sync/build/installer_test.go
+++ b/pkg/sync/build/installer_test.go
@@ -612,8 +612,18 @@ func TestInstaller_CompileFailed(t *testing.T) {
 			failed, err := installer.CompileFailedAndIgnored()
 			if tc.wantErrCompile {
 				require.Error(td, err)
-				assert.ErrorContains(td, err, "yay")
-				assert.Len(t, failed, len(tc.targets))
+				for key := range failed {
+					assert.ErrorContains(td, err, key)
+				}
+				uniqueBases := make(map[string]struct{})
+				for _, layer := range tc.targets {
+					for _, info := range layer {
+						if info.AURBase != nil {
+							uniqueBases[*info.AURBase] = struct{}{}
+						}
+					}
+				}
+				require.Len(td, failed, len(uniqueBases))
 			} else {
 				require.NoError(td, err)
 			}


Ferdi265:
Classic, forgot to run the tests. Something's off here, I'll look into it.

Ferdi265:
This should fix part of the test failures, I accidently used the wrong pkgdests when filling `builtPkgDests`. This still doesn't fix everything though, still looking for the final fix.

Yeah, it seems I don't know exactly how this should interact with the `TestInstaller_CompileFailed` "two layers" test. IIUC it fails to build `bob`, but then doesn't attempt to build `yay` any more since a package with the same pkgbase has already been attempted to be built and failed.

Jguer:
I think if the pkgbase has failed then it's ok to abort the full build for the test

Ferdi265:
> I think if the pkgbase has failed then it's ok to abort the full build for the test

Yeah that makes sense. The whole reason of this PR is avoiding multiple builds of the same pkgbase. So if "bob" fails, "yay" should never be attempted. If a third package "foo" existed, that one should still be attempted though if I understand correctly.

I'm still not sure how to fix the "two layers" test specifically. The assert logic is pretty generic and I don't know how to encode that "bob" and "yay" should count as only one in the error length check without hardcoding 1. Please advise.

Jguer:
```patch
diff --git a/pkg/sync/build/installer_test.go b/pkg/sync/build/installer_test.go
index 615b15f..295cc92 100644
--- a/pkg/sync/build/installer_test.go
+++ b/pkg/sync/build/installer_test.go
@@ -612,8 +612,18 @@ func TestInstaller_CompileFailed(t *testing.T) {
 			failed, err := installer.CompileFailedAndIgnored()
 			if tc.wantErrCompile {
 				require.Error(td, err)
-				assert.ErrorContains(td, err, "yay")
-				assert.Len(t, failed, len(tc.targets))
+				for key := range failed {
+					assert.ErrorContains(td, err, key)
+				}
+				uniqueBases := make(map[string]struct{})
+				for _, layer := range tc.targets {
+					for _, info := range layer {
+						if info.AURBase != nil {
+							uniqueBases[*info.AURBase] = struct{}{}
+						}
+					}
+				}
+				require.Len(td, failed, len(uniqueBases))
 			} else {
 				require.NoError(td, err)
 			}
```

We could check for the base failure (bob) in the error instead of "yay" and compile to the failure of the unique bases.

:shrug: it's not the cleanest but it fixes it for now

Ferdi265:
Implemented as suggested.

Jguer (approved):

PR #2637, opened 2025-06-29
feat: add run0 for privilege elevator

e-kwsm:


Jguer (approved):

Jguer:
I'm merging this as a beginning of conversation @e-kwsm , #2462 has a bunch of nice post points that should be taken into account

PR #2647, opened 2025-07-29
Update README.md

hugok79:
Fix Transifex URL


Jguer:
thank you for your contribution @hugok79 , didn't notice they had migrated off their www.

Jguer (approved):

Issue #2648, opened 2025-08-02
Improved transparency


ben2talk:
### Is your feature request related to a problem? Please describe.

Recent malware scares led me to notice this. My habit is to usually check the AUR page. I noticed many dodgy names in the last few days, but wouldn't know they were new if I were using yay.

### Describe the solution you'd like
1. Search 'google-chrome'
2. See this:
```
1 aur/google-chrome 138.0.7204.183-1 (+2291 10.86)
The popular web browser by Google (Stable Channel)
- Submitted: 2010-05-25 20:25 (UTC) Updated: 2025-07-29 21:37 (UTC)
- https://aur.archlinux.org/packages/google-chrome
```

This because: all the 'bad packages' discovered were brand new.

Sometimes packages are problematic because they weren't updated for an extremely long time.

There's no substitute for a clickable link that takes you to the AUR page.

### Describe alternatives you've considered

I looked at other AUR helpers, pamac/paru are just as 'bad'.

People say you MUST be able to 'vet' your own, however (look at chrome) the pkgbuilds can be extremely tough to analyse.

Seeing popularity is good - votes not so much of a guarantee (perhaps the uploaders of malware will have a dozen friends ready to get some votes registered).

Jguer:
I agree, the recent malware scares should move the needle on what "minimal display is"

https://github.com/Jguer/yay/pull/2330 has been gathering dust for a while as well to see if people upvoted it or the over issue, but I think would be a good inclusion. 
I think it would be a good idea to make this the default setting and keep an option for a minimal display. Even though I personally don’t like having so many different views, it would be helpful for others who don't like change.

tam1m:
I’ve been working on a feature that might go in a similar direction. It flags when packages changed maintainers since the last install/update, so you get a warning before proceeding. I think it's really important since a lot of using the AUR comes down to trusting maintainers. Users should really get notified of an important like this change before letting an update go through.

Looks like this:

![Image](https://github.com/user-attachments/assets/c2844ea5-772b-4686-81cd-001eec2c8500)

I built this by maintaining an offline database similar to how vcs packages are tracked. It's easily extensible for other attributes like upstreamurl for example. If there is interest, I’d be happy to open a PR for this.

relma2:
I came here to file a similar issue; my recommendation would be to **print a warning** if:
- the package is brand-new and uploaded by a brand-new user account
- and/or the package has very few upvotes
- and/or the package has very few downloads
[...]

PR #2650, opened 2025-08-07
Fix cleaning AUR always cleaning all packages

jdholtz:
Since #2644, specifying the `--clean` or `-c` argument would always clean all packages in the AUR cache, rather than needing to provide `-cc`. This PR reverts that change to restore normal functionality (`yay -Sca` removes non-installed AUR package caches and `yay -Scca` removes all AUR package caches). I also added a regression test to catch this if it happens again.

### DIFF
--- a/clean.go
+++ b/clean.go
@@ -52,7 +52,7 @@ func syncClean(ctx context.Context, run *runtime.Runtime, cmdArgs *parser.Argume
 	keepInstalled := false
 	keepCurrent := false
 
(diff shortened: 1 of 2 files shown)


Jguer (approved):

Jguer:
Well caught, thanks @jdholtz. I'll be cutting a release this weekend probably.
After that I think I'll start bigger discussions in https://github.com/Jguer/yay/issues/2648

PR #2653, opened 2025-08-10
Turn search results into clickable links to their repo pages

tam1m:
This turns package names in search results (e.g., `aur/package`) into clickable links pointing to their respective repo pages.

It uses OSC 8 sequences, which are supported by most modern terminals afaik. Terminals without OSC 8 support aren't effected by this and just ignore the added escape characters. 

Repo urls are managed in a map currently. I kept it at offical arch repos and the AUR for now, but I’m not sure if having links to repos other than the AUR is necessary at all tbh.

### DIFF
--- a/pkg/db/mock/repo.go
+++ b/pkg/db/mock/repo.go
@@ -39,6 +39,7 @@ type Package struct {
 	PReason       alpm.PkgReason
 	PDepends      alpm.IDependList
 	PProvides     alpm.IDependList
+	PArchitecture string
 }
 
 func (p *Package) Base() string {
@@ -95,7 +96,7 @@ func (p *Package) Validation() alpm.Validation {
 
 // Architecture returns the package target Architecture.
 func (p *Package) Architecture() string {
-	panic("not implemented")
+	return p.PArchitecture
 }
 
 // Backup returns a list of package backups.
--- a/pkg/query/query_builder_test.go
+++ b/pkg/query/query_builder_test.go
@@ -45,9 +45,9 @@ func TestSourceQueryBuilder(t *testing.T) {
 			verbosity:       Detailed,
 			wantResults:     []string{"linux-ck", "linux-zen", "linux"},
 			wantOutput: []string{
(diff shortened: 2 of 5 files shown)


jdholtz:
This is a super cool feature! Works for me on the latest Kitty version.

One thing I noticed is that the `RepoUrls` map doesn't have `devel` in it, even though it is part of the check on line 78. Adding `"devel": "https://aur.archlinux.org/packages"` to the map works though (not entirely sure if all devel packages are from the AUR).

tam1m:
@jdholtz Good catch! Fixed that. The `devel` repo is indeed for dev packages like -git or -svn from the AUR. Yay handles them differently and and just calls them `devel`. This PR is meant to only handle search results where `devel` won't appear, so it doesn’t matter much here. 
 
The changes in `upgrade.go` slipped in by accident. I might push a follow up PR later to handle clickable links in upgrade lists etc, so I added it to the map anyways.

Jguer (approved):

tam1m:
Is there even interest in extending this beyond the search results? I've played with this in a few more places [here](https://github.com/tam1m/yay/tree/feature/package-links-wip), but it’s getting kind of messy since there are so many different spots we’d need to touch to add this consistently throughout yays interface. Then there is places, like the pacman output, were we can't really do this easily at all. So I'm not sure. Would appreciate some feedback from others on this.

jdholtz:
I personally like the clickable links in the upgrade menu the most as it allows me to easily access changelogs when upgrading (go to Arch repo -> click on source link). Would there be an easy way to centralize outputting package names? For example, creating an `outputPackageNames` function and everywhere in the code call this function?

> Then there is places, like the pacman output, were we can't really do this easily at all.

I think this is fine as there are many things we are able to do with yay that we can't do with pacman (e.g. the fancy upgrade menu).

Jguer:
This is very nice, I'm merging this in as is and we can iterate from there, thanks @tam1m

GrumpyDave:
This new feature broke my scripts that use the yay search output.  I could add some sed filtering to remove it, but it would be nicer if you added an option to not embed extraneous hidden data in the output.  Or perhaps just not do so when stdout is redirected.

FWIW, the clickable links don't work on Konsole.  They don't change the visible output on the screen, so that much is okay.  They just aren't clickable.

CarloDePieri:
This broke my scripts as well.
I can also confirm that in Konsole and in urxvt they are not clickable.

PR #2689, opened 2025-10-13
Don't use OSC 8 links when output is not a terminal

### LINKED ISSUE #2688: Suppress OSC 8 clickable link codes in search output when redirected
Pull request #2653 added hidden OSC 8 codes to search output to make clickable links to the repo entries.  Those codes are fine for interactive use on a terminal, but they break scripts that redirect the search output for further processing.

I suggest that the OSC 8 codes be suppressed when stdout is redirected, just as is done with color codes for the same reasons.

jdholtz:
Fixes #2688. This uses the same method that is used by Yay to determine if color codes should be in the output. 

Since `UseColor` is also available in the `text` package, this could just use the `UseColor` variable and therefore would also be configurable and disabled when you decide to use Yay without color. I didn't add any tests for this as I didn't find it feasible to simulate a TTY that would be detected, but it'd be great to get suggestions on this to add a test for it.

### DIFF
--- a/pkg/query/query_builder_test.go
+++ b/pkg/query/query_builder_test.go
@@ -45,9 +45,9 @@ func TestSourceQueryBuilder(t *testing.T) {
 			verbosity:       Detailed,
 			wantResults:     []string{"linux-ck", "linux-zen", "linux"},
 			wantOutput: []string{
-				"\x1b]8;;https://aur.archlinux.org/packages/linux-ck\x1b\\\x1b[1m\x1b[34maur\x1b[0m\x1b[0m/\x1b[1mlinux-ck\x1b[0m\x1b]8;;\x1b\\ \x1b[36m5.16.12-1\x1b[0m\x1b[1m (+450\x1b[0m \x1b[1m1.51) \x1b[0m\n    The Linux-ck kernel and modules with ck's hrtimer patches\n",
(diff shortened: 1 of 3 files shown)


Jguer:
I think using `useColor` would be a good idea, especially because it lets you test both scenarios easily.

basically: `useColor=true` - all bells and whisles, `useColor=false` - programatic consumption

jdholtz:
I switched to using `UseColor` and added some tests for the `CreateRepoLink` function as well.

By the way, I think the pacman-git test fails are due to commits added in the last week. Probably [this one](https://gitlab.archlinux.org/pacman/pacman/-/commit/748e62fa58e247f22d0262517e367ce07cdc5672).

Jguer (approved):

krydos:
Hi guys, I wonder if someone can make release off of the current master to fix the `12.5.2`. Current 12.5.2 doesn't work well in `eshell` too which apparently doesn't support those codes. 

I just pulled latest master and my eshell issue is gone. 

Just to confirm, do I understand it right, that if pacman.conf doesn't have "Color" option uncommented - this new OSC 8 links functionality isn't going to be triggered?

jdholtz:
> Just to confirm, do I understand it right, that if pacman.conf doesn't have "Color" option uncommented - this new OSC 8 links functionality isn't going to be triggered?

That’s correct, so if it isn’t doing that then let me know as that is a bug. You can also do it with `--color=never` which should turn off color and these links as well.

PR #2705, opened 2025-11-19
fix: Correctly return exit status 1 when no packages are found

### LINKED ISSUE #2701: Status code returned is zero even though package does not exist
### Affected Version

yay v12.5.2 - libalpm v15.0.0

### Describe the bug

When running `yay -S some-unknown-package`, the return code is `0` even though the command did not succeeded with the installation of the unknown or removed package.

### Reproduction Steps

1. Type `yay -S gh-cli`
2. See that the package does not exist in the output
3. Type `echo $last_status` with `fish` shell and see that the error is 0 (success)

### Expected behavior

The command should issue a non-zero return code when a package does not exist, or provide an option to do so.

### Output

```
yay -S gh-cli                                                                                              I 
 -> Aucun paquet trouvé sur l'AUR correspondant à gh-cli
 il n'y a rien à faire
```

sarbojitrana:
**Fixes #2701**

### Summary
This commit addresses the issue where `yay -S <unknown-package>` returns an exit status of `0` (Success) even though the package resolution failed. This violates standard Unix conventions for scripting and automation, as shell scripts rely on non-zero exit codes to detect failure.

### Technical Rationale
The root cause was traced to the `GraphFromAUR` function in `pkg/dep` returning a `nil` error when all package targets failed to resolve, as it considered the *search operation itself* successful.

The fix implements the following changes:

1.  **Error Definition:** A new structured error, `ErrTargetNotFound`, was defined in `pkg/query/errors.go` to provide a specific, translatable error type for this failure mode.
2.  **Failure Tracking:** In `pkg/dep/dep_graph.go`, a counter (`packagesNotFound`) now tracks the number of targets that failed to be resolved in the AUR.
3.  **Error Propagation:** If the final count equals the total number of targets (`packagesNotFound == len(targets)`), `GraphFromAUR` now returns `&query.ErrTargetNotFound{}`.

This non-nil error correctly bubbles up through the execution stack to `cmd/main.go`, ensuring the program triggers the essential `os.Exit(1)` signal to the shell.

### Verification Steps

The behavior was confirmed to be fixed using a local development build:

**Before Fix **
Screenshot From 2025-11-19 21-43-37

**After Fix**

Screenshot From 2025-11-19 21-32-14


sarbojitrana:
@Jguer  could you please review the PR and let me know if anything needs to be changed?

Jguer:
https://github.com/Jguer/yay/pull/2718#issuecomment-3678805344

I'm afraid this change breaks a bunch of CI users relying on yay's ignore and keep building. But since it's covering only the 100% coverage case, let's try it out.

Jguer (approved):

PR #2710, opened 2025-11-25
Don't rely on transparent decompression in sync list

### LINKED ISSUE #2709: Gibberish characters when running `yay -Sl aur`
### Affected Version

yay v12.5.2 - libalpm v15.0.0

### Describe the bug

This is the same bug that I reported in https://github.com/Morganamilo/paru/issues/1447, so I'll just repost the same description, along with later comments I did:

> Today I've started to get an error on `paru -Sl aur`.
> 
> It happens on both the latest release and `-git` version.
> 
> Also, can confirm it's not only a paru issue, since yay also shows the same gibberish. But pikaur does run the command correctly.
> 
> And also, I've checked the packages.gz file used by both made available by Arch and looks fine.

### Reproduction Steps

1. Install yay
2. Run `yay -Sl aur`

### Expected behavior

It should show all the available packages in AUR

### Output

[yay.txt](https://github.com/user-attachments/files/23733555/yay.txt)

espeer:
The current sync list code relies on the HTTP client library transparently decompressing the packages.gz file. This only works if the server provides the 'Content-Encoding: gzip" HTTP header, which has apparently changed recently.

This patch explicitly decompresses the package list if no encoding information is provided by the server.

Fixes: https://github.com/Jguer/yay/issues/2709

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bufio"
 	"context"
 	"errors"
 	"fmt"
@@ -427,18 +426,11 @@ func syncList(ctx context.Context, run *runtime.Runtime,
 	}
 
 	if run.Cfg.Mode.AtLeastAUR() && (len(cmdArgs.Targets) == 0 || aur) {
-		req, err := http.NewRequestWithContext(ctx, http.MethodGet, run.Cfg.AURURL+"/packages.gz", http.NoBody)
+		scanner, err := download.GetPackageScanner(ctx, httpClient, run.Cfg.AURURL)
 		if err != nil {
 			return err
 		}
-
-		resp, err := httpClient.Do(req)
-		if err != nil {
-			return err
-		}
-		defer resp.Body.Close()
-
-		scanner := bufio.NewScanner(resp.Body)
+		defer scanner.Close()
 
 		scanner.Scan()
 
--- a/cmd.go
+++ b/cmd.go
@@ -268,7 +268,7 @@ func handlePrint(ctx context.Context, run *runtime.Runtime, cmdArgs *parser.Argu
 			dbExecutor.LastBuildTime(), run.Cfg.BottomUp, double, quiet)
 	case cmdArgs.ExistsArg("c", "complete"):
 		return completion.Show(ctx, run.HTTPClient, dbExecutor,
-			run.Cfg.AURURL, run.Cfg.CompletionPath, run.Cfg.CompletionInterval, cmdArgs.ExistsDouble("c", "complete"))
+			run.Cfg.AURURL, run.Cfg.CompletionPath, run.Cfg.CompletionInterval, cmdArgs.ExistsDouble("c", "complete"), run.Logger)
 	case cmdArgs.ExistsArg("s", "stats"):
 		return localStatistics(ctx, run, dbExecutor)
 	}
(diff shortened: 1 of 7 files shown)


cmd.go
```diff
@@ -438,6 +439,14 @@ func syncList(ctx context.Context, run *runtime.Runtime,
...
 
+		if resp.Header.Get("Content-Encoding") == "" {
+			gzReader, err := gzip.NewReader(resp.Body)
+			if err == nil {
+				// Replace the raw binary body with the decompressing reader
+				resp.Body = gzReader
+			}
+		}
+
 		scanner := bufio.NewScanner(resp.Body)
```
textfuel: Good fix for handling the Content-Encoding header! A few technical issues to address:

  1. Error handling: Silently ignoring the error makes debugging harder. If decompression fails, we should know about it
  2. Resource management: gzip.Reader allocates internal buffers that need to be closed (per Go's compress/gzip docs), otherwise memory accumulates on repeated calls
  3. Type assignment: Assigning gzReader to resp.Body has type compatibility issues. Better to use a separate io.Reader variable


```suggestion
		var reader io.Reader = resp.Body
		if resp.Header.Get("Content-Encoding") == "" {
			gzipReader, err := gzip.NewReader(resp.Body)
			if err != nil {
				return fmt.Errorf("failed to decompress packages.gz: %w", err)
			}
			defer gzipReader.Close()
			reader = gzipReader
		}

		scanner := bufio.NewScanner(reader)
```

textfuel (changes requested):

cmd.go
espeer: Thanks @cockroach-eater! Go is not my native tongue ;)

The compiler didn't like the idea of assigning to the reader as you proposed. I've pushed a slightly different approach. Please take a look.

cmd.go
```diff
@@ -438,6 +458,17 @@ func syncList(ctx context.Context, run *runtime.Runtime,
 		}
 		defer resp.Body.Close()
 
+		if resp.Header.Get("Content-Encoding") == "" {
```
textfuel: One last question: is it actually correct that we check for an empty `Content-Encoding` instead of `"gzip"`?
Do we know what the server normally sends?
Just want to make sure we're handling the header the right way

cmd.go
textfuel: Thanks @espeer!
It's not my native language either, suffering a bit, but I'm really motivated to fix this bug.
And also, thanks a lot for starting and maintaining this pr in the first place 🙌

cmd.go
espeer: Yes, I believe so. We know the source file is a gzip archive, because that's what we requested. If there is no 'Content-Encoding' , then it should still be a gzip archive. The content encoding affects whether or not the HTTP client will transparently transform it. Note, the 'Content-Type' is still 'gzip', but I don't think we need to check that because we know what we're asking for.

espeer: Scratch that. I just double checked and it seems that the HTTP client also strips the 'Content-Encoding' header when it transparently unzips :/

Thus, the original version of the patch actually worked better than v2, since it still works correctly if the server does include the 'Content-Encoding' header.

I think the right approach is to blindly try to decompress it, irrespective of 'Content-Encoding', and then fall back to the original behavior if the decompression fails, since it must have already been decompressed in that case. I'll spin a 3rd version shortly.

textfuel (approved):

justbispo:
It seems like this function
https://github.com/Jguer/yay/blob/38397d1609cd01cb95d3ce2942beeec794eded6e/pkg/completion/completion.go#L79-L118
also needs to be patched to fix the issue reported in https://github.com/Jguer/yay/issues/2709#issuecomment-3580287522.
Can you also patch that function?

Jguer (approved):

Jguer:
Fixed some minor testing issues and added a missing test. Will merge now, thanks @espeer

Issue #2472, opened 2024-07-14
-Qi, -Si, -Qc, -Qg, -Qk, -Ql not printing exit status in case of an error


drws:
### Affected Version

yay v12.3.5 - libalpm v14.0.0

### Describe the bug

At least `-Qi`, `-Si`, `-Qc`, `-Qg`, `-Qk` and `-Ql` operations aren't printing exit status in case of an error as do other ones (`-Sl` for example). In addition, `yay -Si` is failing to pass Pacman's error message.

### Reproduction Steps

An expected output for reference:
```
$ yay -Sl nonexistent
error: repository "nonexistent" was not found.
 -> exit status 1
```

The issue:
```
$ yay -Si nonexistent
$ yay -Qi nonexistent
error: package 'nonexistent' was not found
$ yay -Qc nonexistent
error: package 'nonexistent' was not found
$ yay -Qg nonexistent
error: group 'nonexistent' was not found
$ yay -Qk nonexistent
error: package 'nonexistent' was not found
$ yay -Ql nonexistent
error: package 'nonexistent' was not found
```

As seen above there is no exit status printed with listed operations even though an error occurs and all the exit codes are 1.

In what is basically a separate, but a very **related issue**, already seen in the output above - `yay -Si` is also failing to pass Pacman's error message:
```
$ pacman -Si nonexistent
error: package 'nonexistent' was not found
```
Skipping the error message can also be observed with `yay -Sg` while that one seems to be a separate issue (in Pacman):
```
$ pacman -Sg nonexistent
$ yay -Sg nonexistent
 -> exit status 1
```

### Expected behavior

- Exit status line (` -> exit status 1`) to be printed in all cases where applicable. There might be other operations affected by the issue. One would expect that this part of the code is shared for all operations and that the status line is printed anytime the status code is non-zero, but I didn't look into it.
- `yay -Si` to pass Pacman's error messages. This additional issue might also be present in other places, a more generalized solution for passing these might be desired too.

PR #2718, opened 2025-12-02
Print an error message when a package isn't found when using -Si

jdholtz:
This fixes the second issue discussed in #2472. It uses the same solution to find missing packages as in #2717, so if both of these get merged, I can create a common function to not duplicate this logic.

A couple of notes on how `yay -Si <pkg>` differs from `pacman -Si <pkg>` (not huge deals IMO):
1. The exit code from yay is 1 when a package isn't found, whereas with pacman the exit code is the number of packages not found (mod 256)
2. pacman prints all packages in order of being specified, including errors. E.g specifying `cmatrix`, `nonexistent`, `firefox` will show the cmatrix info, then `'error: package 'nonexistent' not found`, then firefox info. Yay will print all packages not found first, then repo packages, then AUR packages due to the way the implementation is structured.

Both of these are not big deals, but wanted to point them out in case. I did not update/add any tests as this is just for error printing, but can do so if that is desired.

### DIFF
--- a/query.go
+++ b/query.go
@@ -9,6 +9,7 @@ import (
 	aur "github.com/Jguer/aur"
 	alpm "github.com/Jguer/go-alpm/v2"
 	mapset "github.com/deckarep/golang-set/v2"
+	"github.com/leonelquinteros/gotext"
 
 	"github.com/Jguer/yay/v12/pkg/db"
 	"github.com/Jguer/yay/v12/pkg/query"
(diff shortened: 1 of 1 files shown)


Jguer (approved):

Jguer:
I'm ok with it but there's a bunch of users that run CI environments and want yay to ignore all types of errors in missing packages. I think since this doesn't change the missing logic, we should be ok. 
Merging it in and we'll see based on feedback

PR #2739, opened 2025-12-19
[WIP] Add chroot build feature

androw:
This is a first try at adding the build in chroot feature in yay.
At the moment, it builds fine if no dependency from AUR is needed.
If someone has a idea to fix this last bit, I'm open to suggestion.
Edit: deps now correctly built

Inspired by paru

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -10,6 +10,7 @@ import (
 	alpm "github.com/Jguer/dyalpm"
 	"github.com/leonelquinteros/gotext"
 
+	"github.com/Jguer/yay/v12/pkg/chroot"
 	"github.com/Jguer/yay/v12/pkg/completion"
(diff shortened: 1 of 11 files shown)


Jguer:
This is a nice feature, the AUR struggles with some chroot building so I'm not sure it can ever be the default (plus caching). 

If this reaches completion with AUR dep and coverage I think it should be maintainable enough to merge

radiooperator-19:
I found out yay seems to compile most packages well in chroot, but I need to su to another user (throws `fatal: cannot change to '/tmp/yay': No such file or directory` error when trying to build as root)

androw:
Deps should now be built and installed in the chroot before building the target pkg if someone wants to test.

PR #2741, opened 2025-12-20
Fix building packages with incompatible architectures in makepkg 7.1.0+

jdholtz:
In pacman 7.1.0+, makepkg was changed to check for a package's architecture compatibility when running `--packagelist` (see https://gitlab.archlinux.org/pacman/pacman/-/merge_requests/307). This broke Yay's ability to ignore a package's architecture when installing it (running the package list command would fail with an incompatible architecture error).

`parsePackageList` is only called from the installer which already checks to make sure the user is okay with building the package on the incompatible architecture. In the future if this were to be called from more places it *could* be an issue to always pass `--ignorearch`, but I can't really think of a situation this would happen. I'm happy to add a parameter to `parsePackageList` so the caller has control over passing `--ignorearch` to makepkg.


Jguer (approved):
At some point this will be a problem indeed, keeping this working is hard due to the tests not catching this scenario

PR #2747, opened 2025-12-21
Add support for all sortyby options

### LINKED ISSUE #2539: First search result for "todoist" is unofficial (and unpopular) `todoist-git`
### Affected Version

```
yay v12.4.2 - libalpm v15.0.0
```

### Describe the bug

My mission is to install the official [Todoist](https://todoist.com/) client for Linux: https://todoist.com/downloads/linux.

First result from searching ["todist" on aur.archlinux.org](https://aur.archlinux.org/packages?O=0&K=todoist) is the official client package, [aur/todoist-appimage](https://aur.archlinux.org/packages/todoist-appimage) 9.9.0-1 (+25 0.74):

image

However, the first result from `yay -Ss todoist` is [aur/todoist-git](https://aur.archlinux.org/packages/todoist-git) r326.45738fb-1 (+0 0.00), an unofficial client with zero popularity:

Screenshot_2024-11-29_17-49

### Reproduction Steps

1. Run `yay -Ss todoist`.

### Expected behavior

First result is [aur/todoist-appimage](https://aur.archlinux.org/packages/todoist-appimage).

### Actual behavior

First result is [aur/todoist-git](https://aur.archlinux.org/packages/todoist-git).

### Output

Current yay config:

```json
{
  "buildDir": "/home/matt/.cache/yay",
  "editor": "",
  "editorflags": "",
  "makepkgbin": "makepkg",
  "makepkgconf": "",
  "pacmanbin": "pacman",
  "redownload": "no",
  "answerclean": "",
  "answerdiff": "",
  "answeredit": "",
  "answerupgrade": "",
  "gitbin": "git",
  "gpgbin": "gpg",
  "gpgflags": "",
  "mflags": "",
  "sortby": "popularity",
  "searchby": "name-desc",
  "gitflags": "",
  "removemake": "ask",
  "sudobin": "sudo",
  "sudoflags": "",
  "version": "12.4.2",
  "requestsplitn": 150,
  "completionrefreshtime": 7,
  "maxconcurrentdownloads": 1,
  "bottomup": true,
  "sudoloop": false,
  "timeupdate": false,
  "devel": true,
  "cleanAfter": true,
  "keepSrc": false,
  "provides": true,
  "pgpfetch": true,
  "cleanmenu": false,
  "diffmenu": false,
  "editmenu": false,
  "combinedupgrade": true,
  "useask": false,
  "batchinstall": true,
  "singlelineresults": true,
  "separatesources": true,
  "debug": false,
  "rpc": true,
  "doubleconfirm": true,
  "rebuild": "no"
}
```

### Thoughts

I suspect that renaming unofficial `todoist-git` to something other than `[companyname]`, `[companyname]-bin`, or `[companyname]-git` would fix this bug, amirite? Should be a rule against that, maybe there is. Regardless, the AUR search result parses it the right way (in my opinion).

jdholtz:
Fixes #2539.

Previously, only 'name' was a supported sortby option. Now, all sortby options are supported.

Whenever there is a tie for the primary sort, the metric is used as a tiebreaker (it is also the default when sortby is not specified). The only behavior that changed is that packages with the same metric are sorted oppositely when sorting bottomUp. This was actually a bug in the previous implementation which inverts the `cmpResult` for bottomUp and returns true if the metrics are the same. According to the docs for [Less](https://pkg.go.dev/sort#Interface), Less(i, i) should be false (the same element cannot sort before itself).

I mainly structured the code for efficiency by creating the sortby function once rather than checking `separateSources` and `bottomUp` every comparison, but I also think this improves readability compared to having a switch-case in the comparison function.

Going off of #2740, using a sortby metric besides the default (which uses `calculateMetric`) does not respect the repo order--but it didn't previously for `name` either. I can create a separate PR (or add to this one) to respect the repo order for non-default sortby options too.

### DIFF
--- a/completions/fish
+++ b/completions/fish
@@ -203,7 +203,7 @@ complete -c $progname -n "not $noopt" -l makepkgconf -d 'Use custom makepkg.conf
 complete -c $progname -n "not $noopt" -l nomakepkgconf -d 'Use default makepkg.conf' -f
 complete -c $progname -n "not $noopt" -l requestsplitn -d 'Max amount of packages to query per AUR request' -f
 complete -c $progname -n "not $noopt" -l completioninterval -d 'Refresh interval for completion cache' -f
-complete -c $progname -n "not $noopt" -l sortby -d 'Sort AUR results by a specific field during search' -xa "{votes,popularity,id,baseid,name,base,submitted,modified}"
+complete -c $progname -n "not $noopt" -l sortby -d 'Sort AUR results by a specific field during search' -xa "{votes,popularity,name,base,submitted,modified}"
 complete -c $progname -n "not $noopt" -l searchby -d 'Search for AUR packages by querying the specified field' -xa "{name,name-desc,maintainer,depends,checkdepends,makedepends,optdepends}"
 complete -c $progname -n "not $noopt" -l answerclean -d 'Set a predetermined answer for the clean build menu' -xa "{All,None,Installed,NotInstalled}"
 complete -c $progname -n "not $noopt" -l answerdiff -d 'Set a predetermined answer for the edit diff menu' -xa "{All,None,Installed,NotInstalled}"
--- a/completions/zsh
+++ b/completions/zsh
@@ -58,7 +58,7 @@ _pacman_opts_common=(
 	'--git[git command to use]:git:_files'
 	'--gpg[gpg command to use]:gpg:_files'
 
-	'--sortby[Sort AUR results by a specific field during search]:sortby options:(votes popularity id baseid name base submitted modified)'
+	'--sortby[Sort AUR results by a specific field during search]:sortby options:(votes popularity name base submitted modified)'
 	'--searchby[Search for packages using a specified field]:query'
 	'(--noanswerclean)--answerclean[Set a predetermined answer for the clean build menu]:answer'
 	'(--noanswerdiff)--answerdiff[Set a predetermined answer for the diff menu]:answer'
--- a/doc/yay.8
+++ b/doc/yay.8
@@ -245,7 +245,7 @@ the cache to be refreshed every time, while setting this to -1 will cause the
 cache to never be refreshed.
 
 .TP
-.B \-\-sortby <votes|popularity|id|baseid|name|base|submitted|modified>
+.B \-\-sortby <votes|popularity|name|base|submitted|modified>
 Sort AUR results by a specific field during search.
 
 .TP
--- a/pkg/query/query_builder.go
+++ b/pkg/query/query_builder.go
@@ -84,8 +84,6 @@ type abstractResult struct {
 	name           string
 	description    string
 	packageBase    string
-	packageId      int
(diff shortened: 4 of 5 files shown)


Jguer:
I've tried out the PR on a default config (with separate sources and bottom up). 

Only an exact match should break out of separate sources, which are now penalized

```
3 aur/yay-bin 12.5.7-1 (+339 13.16) (Installed)
    Yet another yogurt. Pacman wrapper and AUR helper written in go. Pre-compiled.
2 aur/yay 12.5.7-1 (+2460 34.60) 
    Yet another yogurt. Pacman wrapper and AUR helper written in go.
1 extra/ruby-yard 0.9.34-5 (807.6 KiB 3.8 MiB) 
    YARD is a Ruby Documentation tool. The Y stands for "Yay!"
==> Packages to install (eg: 1 2 3, 1-3 or ^4)
==> 
```

search such as linux or systemd now do not respect repo order and core repo appears far from 1

jdholtz:
I appreciate the review. I fixed the bug that would caused exact matches to be a part of the separated sources.

I also added tests specifically for this as well as ensuring the QueryBuilder follows the repo order. As an FYI, the existing tests that have a changed order are due to the change in how ties are handled in the sort (it didn't previously follow 'strict weak ordering' which Go's sort expects).

The main concern is that the default config is `sortby: "votes"`. In the next branch, yay uses the metric calculation (which means it doesn't actually sort by votes). As it stands in this PR, AUR packages are now sorted by votes. However, they don't follow the repo order or exact matching breaking out of the separated sources. I kept it this way because name (the only currently implemented sortby option) does not follow this and I am not sure if this is intended (but kept it for the other options to align with how name works).

I see a few different directions to go in (and we could take multiple). Happy to discuss them:
1. Change all sortby options to follow repo order and exact matching. The only change users (who use sortby votes) would see is that the search is now sorted by votes (but exact matches and repo order still work). However, sortby name would change because right now it doesn't follow repo order + exact matching
2. Change sortby votes to use the metric calculations. Users would see no change if they use the default config. However, packages still wouldn't be sorted by votes
3. Change the default config option to use the metric calculations. Without requiring user intervention, existing users would use votes as that is written in their config already

I think 1 would be a good option. I also think it would be good to use the metric calculations by default, but don't see an easy way for users to transition to that. Let me know your thoughts!

jdholtz:
Also, I’m wondering if sorting by ID and BaseID are even useful to provide to users? If not, I can just remove them from the documentation and no users would be affected as it was never actually implemented.

Jguer:
I think sort by ID and baseID are indeed useless for people and could be removed

jdholtz:
> I think sort by ID and baseID are indeed useless for people and could be removed

Removed and updated the documentation to remove them as well

Jguer (approved):
Let's merge this one in and iterate if we see results are not as useful to people. I have one large change to push related to alpm upgrades that changes a few things

PR #2751, opened 2025-12-24
Fix: Remove extra scanner.Scan() call in syncList

avasummer:
When running `yay -l` without arguments, the first AUR package ("010editor") was missing from the output. The list started from the second package("02engine-bin").

The `syncList` function in `cmd.go` had an extra `scanner.Scan()` call, before entering the main loop . This redundant call consumed the first line from the scanner, causing the first package to be skipped.


Jguer (approved):
This is probably due to the format change, well spotted, thanks for the PR

PR #2763, opened 2026-01-15
fix: properly handle install reason on upgrades

### LINKED ISSUE #2288: `yay -S` does not respect the option `--asexplicit`.
### Affected Version

yay v12.1.3.r2.g299aa1e - libalpm v13.0.2

### Describe the bug

`yay -S` does not respect the option `--asexplicit`. Debug output shows that the option is not passed to pacman.

### Reproduction Steps

Let `<package>` be a package that was installed as a dependency for another package.

1. `yay -Qi <package> | grep Reason`
2. `yay -S --asexplicit <package>`
3. The output of `yay -Qi <package> | grep Reason` is unchanged.

### Expected behavior

`yay -S` respects the option `--asexplicit`.

### Output

[`yay --debug -S --asexplicit js91`](https://paste.debian.net/plainh/9a8ac44c)
`yay -Pg`
```json
{
	"buildDir": "/home/x/.cache/yay",
	"editor": "",
	"editorflags": "",
	"makepkgbin": "makepkg",
	"makepkgconf": "",
	"pacmanbin": "pacman",
	"redownload": "no",
	"answerclean": "",
	"answerdiff": "",
	"answeredit": "",
	"answerupgrade": "",
	"gitbin": "git",
	"gpgbin": "gpg",
	"gpgflags": "",
	"mflags": "",
	"sortby": "votes",
	"searchby": "name-desc",
	"gitflags": "",
	"removemake": "ask",
	"sudobin": "sudo",
	"sudoflags": "",
	"version": "11.3.1.r60.g7612bb5",
	"requestsplitn": 150,
	"completionrefreshtime": 7,
	"maxconcurrentdownloads": 0,
	"bottomup": true,
	"sudoloop": true,
	"timeupdate": true,
	"devel": false,
	"cleanAfter": false,
	"keepSrc": false,
	"provides": false,
	"pgpfetch": true,
	"cleanmenu": true,
	"diffmenu": true,
	"editmenu": false,
	"combinedupgrade": false,
	"useask": false,
	"batchinstall": false,
	"singlelineresults": false,
	"separatesources": true,
	"debug": false,
	"rpc": true,
	"doubleconfirm": true,
	"rebuild": "no"
}
```

ironashram:
When a user explicitly re-installs a package with `yay -S <pkg>`, it should be marked as "Explicitly installed" regardless of how it was previously installed.

Right now if a package is already installed as a dependency and you do  `yay -S <pkg>` to explicitly install it, yay preserves the old install reason instead of marking it as explicit as pacman does.

**Before:**
```
$ yay -S vlc-plugin-vorbis
$ yay -Qi vlc-plugin-vorbis | grep "Install Reason"
Install Reason  : Installed as a dependency for another package
```

**After:**
```
$ yay -S vlc-plugin-vorbis
$ yay -Qi vlc-plugin-vorbis | grep "Install Reason"
Install Reason  : Explicitly installed
```

Test also was expecting the wrong behavior so i changed it to reflect this and renamed since it seemed misleading otherwise.

Fixes https://github.com/Jguer/yay/issues/2288

### DIFF
--- a/pkg/dep/dep_graph.go
+++ b/pkg/dep/dep_graph.go
@@ -332,11 +332,7 @@ func (g *Grapher) GraphSyncPkg(ctx context.Context,
 		SyncDBName: &dbName,
 	}
 
-	if upgradeInfo == nil {
-		if localPkg := g.dbExecutor.LocalPackage(pkg.Name()); localPkg != nil {
-			info.Reason = Reason(localPkg.Reason())
-		}
-	} else {
+	if upgradeInfo != nil {
 		info.Upgrade = true
 		info.Reason = Reason(upgradeInfo.Reason)
 		info.LocalVersion = upgradeInfo.LocalVersion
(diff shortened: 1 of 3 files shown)


ironashram:
sorry for the noise, i need more testing on this so i will close it and reopen once i am sure it does work properly

PR #2765, opened 2026-01-19
fix install of AUR package from interactive install menu

benjarobin:
When trying to install an AUR package the following error is printed:
 -> database AUR not found

This regression was introduce in commit
df80f397af63ce51044418c5c9cba0a29aa1c108.
In pkg/query/query_builder.go, sourceAUR was set to AUR instead of aur.

### DIFF
--- a/pkg/dep/dep_graph.go
+++ b/pkg/dep/dep_graph.go
@@ -158,6 +158,7 @@ func (g *Grapher) GraphFromTargets(ctx context.Context,
 
 			fallthrough
 		case "aur":
+		case "AUR":
 			aurTargets = append(aurTargets, target.Name)
 		default:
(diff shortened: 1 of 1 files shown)


Jguer:
:+1: thank you @benjarobin, I think a follow up could be actually using the SourceAUR constant if it does not cause a cyclic dep

Jguer:
@benjarobin I think some tests just need updating their source from "aur" to "AUR"

benjarobin:
I am sorry, but I will not be able to do that, I have no idea for the proper fix, ... I think it will be best that you fix it. You can close this merge request or cherry pick my commit in a branch of yours.

Jguer:
:+1: It will be fixed

PR #2786, opened 2026-02-18
bugfix: sudoloop not working with doas set as sudo executable

mossyguy:
yay-git version used:
yay v12.5.7.r34.g23ee3d3f - libalpm v16.0.1

yays sudoloop does not work with doas set as the sudo executable
```sh
[mossyguy@archlinux ~]$  yay --sudo "doas" --sudoloop
doas: invalid option -- 'v'
usage: doas [-Lns] [-C config] [-u user] command [args]
 -> exit status 1
 ```
 
fixed by instead running true command with elevated privileges


Jguer:
I don't use `doas`, doesn't this also need the user to set `persist` in their doas config?

PR #2796, opened 2026-03-18
Remove timeupdate option

Jguer:
## Summary
- remove the `timeupdate` option from code paths, default config, completions, and docs
- update tests to cover the removed behavior

## Why
`timeupdate` adds config and upgrade-selection complexity for a niche behavior. A quick sample of issue reports where users shared their yay config shows the option disabled, which suggests low real-world usage and weak justification for keeping the extra surface area around.

EDIT: also fixes accidental provides search enable by default

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,6 @@ pacman -S --needed git base-devel yay
 | `yay -Gp <AUR Package>`           | Print to stdout PKGBUILD from ABS or AUR.                                                                  |
 | `yay -Ps`                         | Print system statistics.                                                                                   |
 | `yay -Syu --devel`                | Perform system upgrade, but also check for development package updates.                                    |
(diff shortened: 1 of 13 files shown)


matteoipri:
Hi!
Just to add my 2 cents here.
I never configured yay with a config file, I only use command line interactively and my usual command line is `yay -Syu --devel --timeupdate` which got broken with the latest release.
So, my usage, constant and frequent, for me obviously, for the last years has gone under the radar.
I know that I started using it when some AUR packages got updated without version number change and I kept using it.
Is there any replacement command to check if this kind of update happens again?
Thanks

Gaxsun:
I support matteoipri's point too. I miss --timeupdate. Then again: https://xkcd.com/1172/

PR #2809, opened 2026-03-29
fix: detect stale db.lck caused by terminated process holding the lock

### LINKED ISSUE #2123: Mutual blocking of processes
### Affected Version

yay v12.0.3 - libalpm v13.0.2

### Describe the bug
the process is waiting for the process that started waiting for the first process

### Reproduction Steps
process1: run package update+AUR as user sudo
process2:While getting/checking packages from aur:starting the installation of packages with root.(lock on process2)
When getting/checking into process1 was completed. Process2 is blocked

pay attention to the events between (and including) >>``downloading+checking``<<
```
process1:yay                                              start:1
process2: yay -S package                                  start:2
-process2:lock                                            lock:2
-process1:downloading+checking...
--process2:close (ctrl c)                                 exit:2
-process2:start again                                     start:2
--process2:lock
-process1:aur:downloaded+checked ->ok
--process1:lock
```
finally states:
1)
```
process1:lock
process2:lock
```
2)
```
lock file /var/lib/pacman/db.lck
not served by these 2 processes: process1,process2
only manual removal
```
### Expected behavior

checks that the lock was caused by the parent process and continues its work

Uami-11:
Updated the waitLock() function so that yay no longer waits indefinitely when the db.lck is left behind by a terminated process.

**Changes:**

- Added a findLockOwner function that identifies which process holds db.lck
- Added Tests for findLockOwner that covers live and stale lock cases
- Modified waitLock to call findLockOwner to check if the process holding lock is live
- waitLock now prompts user to remove the lock if the lock is found to be stale

Closes #2123

### DIFF
--- a/pkg/settings/exe/cmd_builder.go
+++ b/pkg/settings/exe/cmd_builder.go
@@ -238,7 +238,7 @@ func (c *CmdBuilder) BuildPacmanCmd(ctx context.Context, args *parser.Arguments,
 	argArr = append(argArr, args.Targets...)
 
 	if needsRoot {
-		c.waitLock(c.PacmanDBPath)
+		c.waitLock(ctx, c.PacmanDBPath)
 
 		if os.Geteuid() != 0 {
 			return c.buildPrivilegeElevatorCommand(ctx, argArr)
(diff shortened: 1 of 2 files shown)


Uami-11:
There were some problems with the removal, so I fixed that. And I added a warning to show how many processes could not be seen due to permission. Here's a demonstration on my solution to a stale lock. 

fix-showcase

PR #2846, opened 2026-06-11
feat: Show AUR Package Age

rebelonion:
Related issue: #2824, specifically [this suggestion comment](https://github.com/Jguer/yay/issues/2824#issuecomment-4467797349).

### Adds AUR package age badges based on the package `LastModified` timestamp.

  The age badge is shown for AUR packages in:

  - `yay -Ss <package>`
  - `yay <package>`
  - `yay -Qum`
  - `yay -Qu`
  - `yay -Syu`

  Badges show at most 2 units (never seconds), for example `[23h59m]`, `[3h]`, `[54m]`, or `[38d8h]`.

### Colors For badges:
  - Less than 24 hours old: red + bold
  - Less than 7 days old: yellow + bold
  - 7 days or older: cyan

### Note:

`make test` fails (in my time zone) in `pkg/text` due to existing timezone-sensitive `TestFormatTime` behavior in `America/Chicago`: Unix timestamp `0` formats as `1969-12-31` locally instead of the test’s expected `1970-01-01`.

This is a preexisting issue not caused by this PR.

  ## Screenshots

  <details>
  <summary><code>yay -Ss &lt;package&gt;</code></summary>

  
image

  </details>

  <details>
  <summary><code>yay &lt;package&gt;</code></summary>

image

  </details>

  <details>
  <summary><code>yay -Qum</code></summary>

 image

  </details>

  <details>
  <summary><code>yay -Qu</code></summary>

image

  </details>

  <details>
  <summary><code>yay -Syu</code></summary>

  image

  </details>

PR #2848, opened 2026-06-12
feat: add --upgrademinage to delay AUR updates

mosajjal:
## What

Adds an `--upgrademinage <hours>` option that withholds an AUR package upgrade until its AUR `LastModified` timestamp is at least the given number of hours old.

The idea is a cooldown on fresh AUR pushes: instead of being offered an update the moment a maintainer pushes it, you only see it once it has survived in the AUR for a while. That leaves time for broken or malicious pushes to be noticed and corrected before they land on your machine.

## Behaviour

- Configurable via the `--upgrademinage` flag or the `upgrademinage` config key (JSON and Lua).
- `0` is the default and disables the delay, so existing behaviour is unchanged for anyone who does not set it.
- Only AUR packages are affected. Repo packages (handled by pacman) and devel/VCS packages are untouched, since their age is not a meaningful signal here.
- When an upgrade is held back, yay logs why, e.g. `hello: delaying upgrade (updated 1h0m0s ago, minimum age is 48h)`.

This builds directly on the recently added AUR package age feature (#2846) — the `LastModified` timestamp is already collected per package, so this just filters on it.

## Implementation notes

- The filter lives in `UpAUR`, which now takes the current time as a parameter rather than calling `time.Now()` internally, so the age logic is testable with a fixed clock.
- Option wiring mirrors the existing `requestsplitn` integer option (config struct + default, arg parser, `isArg`/`hasParam`).
- Man page, `--help` text, and bash/zsh/fish completions updated.
- New unit tests cover both the delayed and old-enough cases.

## Open question

Happy to adjust the scope or naming. If a duration string (e.g. `48h`) is preferred over a plain hours integer, I can switch to that — I kept it as an integer of hours for simplicity since Go's `time.ParseDuration` does not support days/weeks anyway.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -77,6 +77,7 @@ Permanent configuration options:
     --nomakepkgconf       Use the default makepkg.conf
 
     --requestsplitn <n>   Max amount of packages to query per AUR request
+    --upgrademinage <h>   Withhold AUR updates until the package is <h> hours old (0 to disable)
     --completioninterval  <n> Time in days to refresh completion cache
     --sortby    <field>   Sort AUR results by a specific field during search
     --searchby  <field>   Search for packages using a specified field
--- a/completions/bash
+++ b/completions/bash
@@ -71,7 +71,7 @@ _yay() {
   ##yay stuff
   common=('arch cachedir color config confirm dbpath debug gpgdir help hookdir logfile
           noconfirm noprogressbar noscriptlet quiet root verbose disable-download-timeout sysroot
-          save builddir editor editorflags makepkg pacman git gitflags gpg gpgflags mflags config requestsplitn sudoloop sudo sudoflags
+          save builddir editor editorflags makepkg pacman git gitflags gpg gpgflags mflags config requestsplitn upgrademinage sudoloop sudo sudoflags
           redownload noredownload redownloadall rebuild rebuildall rebuildtree norebuild sortby
           singlelineresults doublelineresults answerclean answerdiff answeredit answerupgrade noanswerclean noanswerdiff
           noansweredit noanswerupgrade cleanmenu diffmenu editmenu cleanafter keepsrc topdown bottomup devel separatesources
--- a/completions/fish
+++ b/completions/fish
@@ -204,6 +204,7 @@ complete -c $progname -n "not $noopt" -l config -d 'The pacman config file to us
 complete -c $progname -n "not $noopt" -l makepkgconf -d 'Use custom makepkg.conf location' -r
 complete -c $progname -n "not $noopt" -l nomakepkgconf -d 'Use default makepkg.conf' -f
 complete -c $progname -n "not $noopt" -l requestsplitn -d 'Max amount of packages to query per AUR request' -x
+complete -c $progname -n "not $noopt" -l upgrademinage -d 'Withhold AUR updates until the package is n hours old' -x
 complete -c $progname -n "not $noopt" -l completioninterval -d 'Refresh interval for completion cache' -x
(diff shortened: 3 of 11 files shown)


mosajjal:
Hi @Jguer, given the AUR incident and compromised packages, this would be a helpful option especially for auto update scripts. Most of the code is written by AI btw. I took a look myself and it doesn't look like slop hence pushing it to a PR.

jdholtz:
Hey @mosajjal, I believe this can be handled by the newly implemented Lua hooks (an example of this exact scenario is in the example config). See https://github.com/Jguer/yay/blob/4dbf124c51bcf11b7d8bcc86b832353fca7d66b2/doc/init.lua#L60C41-L74C6. Let me know if there are limitations to this that your approach doesn't have though.

mosajjal:
hey @jdholtz, thanks for this. I wasn't aware of this feature. Is it planned to be shipped in 13.x? If so, let's close this PR and document the hook instead.

jdholtz:
Not sure which release it is planned for, will defer to @Jguer.

Jguer:
https://jguer.space/blog/2026-06-15-yay-v13

v13 is out and includes lua hooks that allow implementing this. 
This allows users to configure the delay and even setup exemptions and other rules. 
Please create an issue if there's something missing from the lua hooks

mosajjal:
Lua hook to do the same:

```lua
(~/.config/yay/init.lua):

-- 48-hour cooldown on system upgrades.
--
-- Blocks `yay -Syu` (and bare `yay`) if the last successful upgrade was less
-- than 48h ago. Override a single run with: YAY_FORCE=1 yay -Syu
--
-- Replaces the old fish `yay` wrapper function. Uses yay v13 Lua hooks:
--   UpgradeSelect -> fires before the upgrade; aborts if still cooling down.
--   PostInstall   -> fires after a successful transaction; records the time.

local stamp = (os.getenv("HOME") or "") .. "/.cache/yay-cooldown.stamp"
local cooldown = 48 * 60 * 60 -- seconds

-- Set in UpgradeSelect so PostInstall only stamps actual sysupgrades, not a
-- plain `yay -S pkg`. Module-level state persists for the process lifetime.
local is_upgrade = false

local function last_upgrade()
    local f = io.open(stamp, "r")
    if not f then return nil end
    local t = tonumber(f:read("*l"))
    f:close()
    return t
end

yay.create_autocmd("UpgradeSelect", {
    desc = "48h cooldown between system upgrades",
    callback = function(event)
        is_upgrade = true

        if os.getenv("YAY_FORCE") then
            return { exclude = {}, skip_menu = false }
        end

        local last = last_upgrade()
        if last then
            local elapsed = os.time() - last
            if elapsed < cooldown then
                local elapsed_h = math.floor(elapsed / 3600)
                local remain_h = math.floor((cooldown - elapsed) / 3600)
                yay.abort(string.format(
                    "48h upgrade cooldown active — last upgrade %dh ago, ~%dh remaining. "
                        .. "Override: YAY_FORCE=1 yay -Syu",
                    elapsed_h, remain_h))
            end
        end

        return { exclude = {}, skip_menu = false }
    end,
})

yay.create_autocmd("PostInstall", {
    desc = "record successful upgrade time for the cooldown",
    callback = function(event)
        if not is_upgrade then return end
        local f = io.open(stamp, "w")
        if f then
            f:write(tostring(os.time()) .. "\n")
            f:close()
        end
    end,
})
```

adrinjalali:
For people who might end up here and want a bit more than the Lua hook, especially for packages which update too frequently resulting in a cooldown to never update them, I wrote https://github.com/adrinjalali/aur-cooldown , which you might find useful.

PR #2884, opened 2026-06-20
feat(lua): add lua-language-server type stubs for the yay global

### LINKED ISSUE #2876: lua stubs for config

```
Undefined global `yay` 
```

warning and lack of lsp

Ship lua stubs for config, similar how hyprland does it:

- https://wiki.hypr.land/Configuring/Start/#autocompletions
- https://github.com/hyprwm/Hyprland/blob/main/meta/generateLuaStubs.py

beweme11:
## What

Resolves #2876 — ships a `meta/yay.meta.lua` type definition file so [lua-language-server](https://luals.github.io/) users no longer get `Undefined global \`yay\`` warnings, and gain completion, hover docs, and typo detection for the whole Lua config API.

## Why

The issue asked for Lua stubs similar to how Hyprland ships them. yay's Lua API is small and stable (1 global, 4 top-level members, 45 opt fields, 5 autocmd events), so a hand-written `---@meta` file is the right fit — no generator needed.

## Changes

- **`meta/yay.meta.lua`** (new, 232 lines) — `---@meta` file declaring:
  - `yay` global with `opt`, `log`, `abort`, `create_autocmd`
  - `yay.opt` — all 45 fields from `Configuration` struct `lua:"..."` tags (string/boolean/integer/`RebuildMode`)
  - `yay.log` — debug/info/warn/error
  - `yay.abort(reason)`
  - `yay.create_autocmd` — 5 `---@overload` annotations, one per event, so each event string maps to its specific callback payload type and return contract
  - 14 event payload classes covering all 5 events (AURPreInstall, AURPostDownload, UpgradeSelect, PostInstall, SearchFilter)
  - `---@alias yay.RebuildMode` = `"no"|"yes"|"tree"|"all"`
  - Return contracts annotated for UpgradeSelect (`{exclude, skip_menu}`) and SearchFilter (`SearchResultRef[]`)

- **`doc/lua.md`** — new "Editor support" section with `.luarc.json` setup example pointing `workspace.library` at the `meta/` directory. Notes the `Lua 5.1` runtime requirement (yay uses gopher-lua).

- **`doc/init.lua`** — fixed PostInstall example: removed `if pkg.installed then` check referencing a field that does not exist on `PostInstallPackage` (actual fields: name, version, local_version, source, reason).

- **`Makefile`** — install/uninstall `meta/yay.meta.lua` to `$(PREFIX)/share/yay/meta/` so the documented `/usr/share/yay/meta` path exists on installed systems (same pattern already used for man page and shell completions).

### Field accuracy
Cross-checked every stub field against the Go source:
- `pkg/settings/config.go` — all 45 `lua:"..."` tags match (names, types, exclusions for `lua:"-"`)
- `pkg/settings/lua/autocmd.go` — all 45 `RawSetString` payload field names + 4 return-contract fields match exactly (snake_case, types)
- `pkg/settings/lua/log.go`, `abort.go`, `lua.go`, `parser/rebuild_mode.go` — all match

### lua-language-server tests
- **Positive**: realistic `init.lua` exercising the full API (all opt types, log, abort, all 5 autocmd events with full payload access, `require()` module) + `.luarc.json` pointing at the stub → **0 warnings**
- **Negative**: typos caught — undefined global (`yae`), unknown opt field (`yay.opt.botom_up` → inject-field), invalid event string (`"AURPreInstal"` → param-type-mismatch), misspelled payload field (`event.data.pkgbuld` → undefined-field)
- **Original issue**: `yay.opt.bottom_up = true` → no `Undefined global \`yay\`` warning ✓
- All repo Lua files (`doc/init.lua`, 4 examples in `doc/examples/`) pass clean with the stub

### Runtime tests (actual yay binary)
- `make build` → exit 0
- `go build ./... && go test -race ./...` → all packages pass, exit 0
- Ran the built binary with a realistic config (`XDG_CONFIG_HOME` pointing at a test dir with `init.lua` + required `hooks/maintainer_guard.lua`): `yay -Pg` → exit 0, every `init.lua` option applied correctly (sortby, devel, rebuild, maxconcurrentdownloads, removemake, editor, etc.)
- Typo config at runtime: `yay -Pg` → exit 1 with clear error (`unsupported event "AURPreInstal"` / `unknown yay.opt key: botom_up`)

### Known limitation
lua-language-server does not check value-type compatibility on assignments to global variable fields (e.g. `yay.opt.aururl = 123` will not warn about the type mismatch). This is a LuaLS design limitation for globals, not a stub defect. Field *existence* (typos) IS checked, and callback parameter types ARE inferred from the per-event overloads.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,7 @@ install: build ${MOFILES}
 	install -Dm644 completions/bash $(DESTDIR)$(PREFIX)/share/bash-completion/completions/${PKGNAME}
 	install -Dm644 completions/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_${PKGNAME}
 	install -Dm644 completions/fish $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/${PKGNAME}.fish
+	install -Dm644 meta/yay.meta.lua $(DESTDIR)$(PREFIX)/share/${PKGNAME}/meta/yay.meta.lua
 	for lang in ${LANGS}; do \
 		install -Dm644 ${LOCALEDIR}/$${lang}.mo $(DESTDIR)$(PREFIX)/share/locale/$$lang/LC_MESSAGES/${PKGNAME}.mo; \
 	done
@@ -106,6 +107,7 @@ uninstall:
 	rm -f $(DESTDIR)$(PREFIX)/share/bash-completion/completions/${PKGNAME}
 	rm -f $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_${PKGNAME}
 	rm -f $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/${PKGNAME}.fish
+	rm -f $(DESTDIR)$(PREFIX)/share/${PKGNAME}/meta/yay.meta.lua
 	for lang in ${LANGS}; do \
 		rm -f $(DESTDIR)$(PREFIX)/share/locale/$$lang/LC_MESSAGES/${PKGNAME}.mo; \
 	done
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ install: build ${MOFILES}
 	install -Dm644 completions/bash $(DESTDIR)$(PREFIX)/share/bash-completion/completions/${PKGNAME}
 	install -Dm644 completions/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_${PKGNAME}
 	install -Dm644 completions/fish $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/${PKGNAME}.fish
-	install -Dm644 meta/yay.meta.lua $(DESTDIR)$(PREFIX)/share/${PKGNAME}/meta/yay.meta.lua
+	install -Dm644 meta/yay.d.lua $(DESTDIR)$(PREFIX)/share/${PKGNAME}/meta/yay.d.lua
 	for lang in ${LANGS}; do \
 		install -Dm644 ${LOCALEDIR}/$${lang}.mo $(DESTDIR)$(PREFIX)/share/locale/$$lang/LC_MESSAGES/${PKGNAME}.mo; \
 	done
@@ -107,7 +107,7 @@ uninstall:
 	rm -f $(DESTDIR)$(PREFIX)/share/bash-completion/completions/${PKGNAME}
 	rm -f $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_${PKGNAME}
 	rm -f $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/${PKGNAME}.fish
-	rm -f $(DESTDIR)$(PREFIX)/share/${PKGNAME}/meta/yay.meta.lua
+	rm -f $(DESTDIR)$(PREFIX)/share/${PKGNAME}/meta/yay.d.lua
 	for lang in ${LANGS}; do \
 		rm -f $(DESTDIR)$(PREFIX)/share/locale/$$lang/LC_MESSAGES/${PKGNAME}.mo; \
 	done
--- a/doc/init.lua
+++ b/doc/init.lua
@@ -114,9 +114,7 @@ yay.opt.double_confirm = true -- Ask for confirmation before and after builds du
 --   desc = "log every package yay installed",
 --   callback = function(event)
 --     for _, pkg in ipairs(event.data.packages) do
---       if pkg.installed then
---         yay.log.info(pkg.name .. " " .. pkg.version .. " installed (" .. pkg.source .. ")")
---       end
+--       yay.log.info(pkg.name .. " " .. pkg.version .. " (" .. pkg.source .. ")")
 --     end
 --   end,
 -- })
--- a/doc/init.lua
+++ b/doc/init.lua
@@ -4,56 +4,53 @@
 -- or keep all of them and tune values. Command-line flags still override
 -- these values.
 
--- Strings
-yay.opt.aururl = "https://aur.archlinux.org" -- Base AUR URL.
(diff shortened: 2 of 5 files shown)


doc/lua.md
```diff
@@ -450,3 +450,40 @@ yay.create_autocmd("SearchFilter", {
   end,
 })
 ```
+
+## Editor support
+
+<p class="api-since">Available from yay v13.0.0</p>
```
danielwerg: latest version is 13.0.1

doc/init.lua
```diff
@@ -114,9 +114,7 @@ yay.opt.double_confirm = true -- Ask for confirmation before and after builds du
 --   desc = "log every package yay installed",
 --   callback = function(event)
 --     for _, pkg in ipairs(event.data.packages) do
---       if pkg.installed then
```
danielwerg, line 117: why was this check removed?

meta/yay.meta.lua
```diff
@@ -0,0 +1,232 @@
...
+---@field aurrpcurl string
+---@field build_dir string
+---@field editor string
+---@field editor_flags string
+---@field makepkg_bin string
+---@field makepkg_conf string
+---@field pacman_bin string
+---@field pacman_conf string
+---@field redownload string
+---@field answer_clean string
```
danielwerg: should be something like this:

```lua
---also accepts menu syntax: ranges, ^n, "abort"
---@field answer_clean ""|"All"|"None"|"Installed"|"NotInstalled"
```

instead of simple `string`. that way we can omit docs inside init.lua.

same goes for many other options below.

danielwerg (changes requested):
couple issues at a glance.

danielwerg:
@beweme11 add me as a contributor to the fork if u need help reformatting `meta/yay.meta.lua`. but first i want to hear @Jguer opinion on annotations style.

doc/init.lua
beweme11: pkg.installed was dead code, PR #2871 removed that field from PostInstallPackage but left the example stale.

meta/yay.meta.lua
beweme11: ok lets wait for his review

doc/lua.md
beweme11: have addressed this in latest commit, thanks!

meta/yay.meta.lua
Jguer: There's not much interest in using the the numbers allowed or exclusion here, except `abort`, so options seems better

Makefile
```diff
@@ -95,6 +95,7 @@ install: build ${MOFILES}
 	install -Dm644 completions/bash $(DESTDIR)$(PREFIX)/share/bash-completion/completions/${PKGNAME}
 	install -Dm644 completions/zsh $(DESTDIR)$(PREFIX)/share/zsh/site-functions/_${PKGNAME}
 	install -Dm644 completions/fish $(DESTDIR)$(PREFIX)/share/fish/vendor_completions.d/${PKGNAME}.fish
+	install -Dm644 meta/yay.meta.lua $(DESTDIR)$(PREFIX)/share/${PKGNAME}/meta/yay.meta.lua
```
Jguer: is this the usual location for this type of files?

meta/yay.meta.lua
Jguer: where we can provide choices, it's better than allowing any value (even if it's programmatically correct)

Jguer:
👍 not an expert on lua lsp type stubs, so once you and @danielwerg are in agreement it is good to merge, I will merge

Makefile
danielwerg: don't think so, but hyprland using `/usr/share/hypr/stubs/hl.meta.lua`. only thing i found in luals docs is this bit about file name:

https://luals.github.io/wiki/definition-files/

> Recommendation
> 
> Name your definition files *.d.lua, such as filesystem.d.lua, to make it clear which files are purely for documentation and not implementation code.

i think both `meta` and `stubs` are appropriate dir names, but maybe rename `yay.meta.lua` -> `yay.d.lua`?

doc/lua.md
```diff
@@ -453,7 +453,7 @@ yay.create_autocmd("SearchFilter", {
 
 ## Editor support
 
-<p class="api-since">Available from yay v13.0.0</p>
+<p class="api-since">Available from yay v13.0.1</p>
```
danielwerg: this unmerged feature can't be available in 13.0.1 which is already released. it's really up for maintainer to set this, we can't really know in which version this will be available.

danielwerg (changes requested):
i don't have push rights, but here are the patches:

typed `yay.opt` whenever possible, added description. do note that with [example `init.lua`](https://github.com/Jguer/yay/blob/next/doc/init.lua) user will get doubled description on hover docs, here is a patch that's strips description comments from `init.lua`: [init.lua.patch](https://github.com/user-attachments/files/29225753/init.lua.patch) (`Strings`, `Integers`, `Booleans` comments were polluting hover docs for following options)

[yay.opt.patch](https://github.com/user-attachments/files/29225468/yay.opt.patch)

added `create_autocmd` to `yay` global:

[yay.create_autocmd.patch](https://github.com/user-attachments/files/29225557/yay.create_autocmd.patch)

doc/lua.md
```diff
@@ -450,3 +450,40 @@ yay.create_autocmd("SearchFilter", {
   end,
 })
 ```
+
+## Editor support
+
+<p class="api-since">Available from yay v13.0.1</p>
```
Jguer:
```suggestion
<p class="api-since">Available from yay v13.0.2</p>
```

beweme11:
hey @danielwerg  have added your patches and added you as a contributor to this fork. you should have received an invite

danielwerg (approved):
lgtm!

Jguer (approved):

Jguer:
Merge it is then. Thank you both for picking this up

Issue #2573, opened 2025-02-02
Yay replaces pacman package with AUR package on package rename


agboom:
### Affected Version

yay v12.4.2 - libalpm v15.0.0

### Describe the bug

When package from pacman repo is renamed (e.g. `extra/sdl2` to `extra/sdl2-compat`), Yay tries to install a package with the old name from the AUR (e.g. `aur/sdl2`), even if its dependents are only pacman packages (non AUR).

### Reproduction Steps

1. Have `extra/sdl2` installed from official repos (e.g. through `extra/ffmpeg`)
2. Run `yay`
3. Observe line saying ` 1  aur/sdl2                            2.30.11-1              -> 2.30.11-2`

### Expected behavior

Yay should not substitute AUR packages for pacman packages with the same name (e.g. `extra/sdl2` to `aur/sdl2`).

### Output

Since my system is already updated I cannot reproduce it with debug enabled. If there's any other way please let me know.

```
yay -Pg
{
	"aururl": "https://aur.archlinux.org",
	"aurrpcurl": "https://aur.archlinux.org/rpc?",
	"buildDir": "/home/adriaan/.cache/yay",
	"editor": "",
	"editorflags": "",
	"makepkgbin": "makepkg",
	"makepkgconf": "",
	"pacmanbin": "pacman",
	"pacmanconf": "/etc/pacman.conf",
	"redownload": "no",
	"answerclean": "",
	"answerdiff": "",
	"answeredit": "",
	"answerupgrade": "",
	"gitbin": "git",
	"gpgbin": "gpg",
	"gpgflags": "",
	"mflags": "",
	"sortby": "votes",
	"searchby": "name-desc",
	"gitflags": "",
	"removemake": "ask",
	"sudobin": "sudo",
	"sudoflags": "",
	"version": "12.4.2",
	"requestsplitn": 150,
	"completionrefreshtime": 7,
	"maxconcurrentdownloads": 1,
	"bottomup": true,
	"sudoloop": false,
	"timeupdate": false,
	"devel": false,
	"cleanAfter": false,
	"keepSrc": false,
	"provides": true,
	"pgpfetch": true,
	"cleanmenu": true,
	"diffmenu": true,
	"editmenu": false,
	"combinedupgrade": true,
	"useask": false,
	"batchinstall": false,
	"singlelineresults": false,
	"separatesources": true,
	"debug": false,
	"rpc": true,
	"doubleconfirm": true,
	"rebuild": "no"
}
```

```
[...]

Issue #2375, opened 2024-02-20
If package was removed from official repos - it starts building AUR package with the same name without warning


Sidicer:
### Affected Version
yay v12.3.0 - libalpm v13.0.2

### Describe the bug
If you had a package installed in your system from official repositories (core/extra/multilib) and that package has been removed - YAY will automatically fetch and build an AUR package with the same package name without any warning.

This happened with [electron24](https://archlinux.org/packages/extra/x86_64/electron24/) package which was recently removed from `extra` repository. As [electron24](https://aur.archlinux.org/packages/electron24) exists in AUR it started automatically downloading after `yay -Syu`.
Due to the package having `chromium.git` as a dependency - we got 40+GB download.

### Reproduction Steps

1. Install package from Official Repos
2. Create package in AUR with the same name
3. Remove the original package from official repos
4. `yay -Syu`

### Expected behavior
I feel as this could have easily been a supply-chain attack and I feel that if a package was removed from official repos, but AUR has it - a warning or at least (y/n) prompt should  appear asking if you want to switch.

Jguer:
pacman doesn't store the source of the package so it's hard for us to know when it has changed except if we keep our own table (which is not great because if you stop using yay or do manual building it will become corrupt).

Looking at the information available to us, maybe we can use Packager as a flag for it but it will be hacky

PR #2910, opened 2026-07-18
Prefer repository replacements over matching AUR upgrades

### LINKED ISSUE #2579: yay doesn't ask to replace sdl2, instead it's compiling it
### Affected Version

yay v12.4.2 - libalpm v15.0.0

### Describe the bug

`yay -y` doesn't offer to replace sdl2 with sdl2-compat as it's described in the official PKGBUILD (sdl2-compat 2.32.50-1). Instead it wants to get sdl2 from AUR as illustrated below:

```
....
  3  extra/zix                            0.4.2-2                     -> 0.6.2-1
  2  extra/zvbi                           0.2.42-1                    -> 0.2.43-1
  1  aur/sdl2                             2.30.11-1                   -> 2.32.0-1
==> Packages to exclude: (eg: "1 2 3", "1-3", "^4" or repo name)
 -> Excluding packages may cause partial upgrades and break systems
==> ^C
```

At the same time `pacman -Syyu` command asks to replace sdl2 with sdl2-compat:

```
:: Replace sdl2 with extra/sdl2-compat? [Y/n] 
resolving dependencies...

```

### Reproduction Steps

If the installation is already upgraded to the current packages do 1-4 to get pre sdl2-compat installation:
1. edit pacman.conf to have 2025-01-15 snapshot repos (from Arch Linux Archive)
2. `pacman -Syyuu` to downgrade all packages
3. (`pacman -S sdl2` - reinstall sdl2 if you already have sdl2-compat to revert to sdl2)
4. revert pacman.conf to have normal core and extra repos.
5. `yay -y` doesn't ask about replacement for sdl2

### Expected behavior

`yay` should ask about sdl2 replacement like pacman does
```
$ yay -Pg
{
	"buildDir": "/home/user/.cache/yay",
	"editor": "",
	"editorflags": "",
	"makepkgbin": "makepkg",
	"makepkgconf": "",
	"pacmanbin": "pacman",
	"redownload": "no",
	"answerclean": "",
	"answerdiff": "",
	"answeredit": "",
	"answerupgrade": "",
	"gitbin": "git",
	"gpgbin": "gpg",
	"gpgflags": "",
	"mflags": "",
	"sortby": "votes",
	"searchby": "name-desc",
	"gitflags": "",
	"removemake": "ask",
	"sudobin": "sudo",
	"sudoflags": "",
	"version": "12.4.2",
	"requestsplitn": 150,
	"completionrefreshtime": 7,
	"maxconcurrentdownloads": 1,
	"bottomup": true,
	"sudoloop": false,
	"timeupdate": false,
	"devel": false,
	"cleanAfter": false,
	"keepSrc": false,
	"provides": true,
	"pgpfetch": true,
	"cleanmenu": true,
	"diffmenu": true,
	"editmenu": false,
	"combinedupgrade": true,
	"useask": false,
	"batchinstall": false,
	"singlelineresults": false,
	"separatesources": true,
	"debug": false,
	"rpc": true,
	"doubleconfirm": true,
	"rebuild": "no"
}
```

floze-the-genius:
## Summary

- compute the pacman sysupgrade plan before AUR upgrade discovery
- exclude installed package names that are replaced by repository upgrades from AUR/devel lookup
- expose package replacement metadata through the database executor and add an sdl2 -> sdl2-compat regression test

## Root cause

When a repository package was renamed, yay classified the installed old name as foreign before considering pacman's replacement plan. If an AUR package reused that old name, it was added to the upgrade graph and masked the repository replacement that pacman would otherwise offer.

The fix only suppresses AUR discovery for names explicitly covered by `Replaces` in the current repository sysupgrade transaction. Other foreign packages continue through the existing AUR path.

Fixes #2579.

Related duplicate context: #2573 and #2375.

## Validation

- focused upgrade regression tests
- `make test` in the repository's Arch Linux CI image (full race-enabled suite)
- `go vet ./...`
- `go build ./...`
- `golangci-lint v2.12.2 run --concurrency=8 ./...` (`0 issues`)
- `gofmt` and `git diff --check`

## AI assistance disclosure

This patch was developed with OpenAI Codex. I reviewed the resulting diff and test evidence and will respond to maintainer feedback.

### DIFF
--- a/pkg/db/executor.go
+++ b/pkg/db/executor.go
@@ -54,6 +54,7 @@ type Executor interface {
 	PackageGroups(IPackage) []string
 	PackageOptionalDepends(IPackage) []Depend
 	PackageProvides(IPackage) []Depend
+	PackageReplaces(IPackage) []Depend
 	PackagesFromGroup(string) []IPackage
 	PackagesFromGroupAndDB(string, string) ([]IPackage, error)
 	RefreshHandle() error
--- a/pkg/db/ialpm/alpm.go
+++ b/pkg/db/ialpm/alpm.go
@@ -415,6 +415,14 @@ func (ae *AlpmExecutor) PackageProvides(pkg alpm.Package) []alpm.Depend {
 	return pkg.Provides()
 }
 
+func (ae *AlpmExecutor) PackageReplaces(pkg alpm.Package) []alpm.Depend {
+	if pkgWithReplaces, ok := pkg.(interface{ Replaces() []alpm.Depend }); ok {
+		return pkgWithReplaces.Replaces()
+	}
+
+	return nil
+}
+
 func (ae *AlpmExecutor) PackageGroups(pkg alpm.Package) []string {
 	return pkg.Groups()
 }
--- a/pkg/db/mock/executor.go
+++ b/pkg/db/mock/executor.go
@@ -27,6 +27,7 @@ type DBExecutor struct {
 	PackageDependsFn              func(IPackage) []Depend
 	PackageOptionalDependsFn      func(alpm.Package) []alpm.Depend
 	PackageProvidesFn             func(IPackage) []Depend
+	PackageReplacesFn             func(IPackage) []Depend
 	PackagesFromGroupFn           func(string) []IPackage
 	PackagesFromGroupAndDBFn      func(string, string) ([]IPackage, error)
 	RefreshHandleFn               func() error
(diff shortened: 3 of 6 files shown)


Jguer:
Seems clean, There should be an issue on `yay -Qu --deps` as the filter happens later and as such the replaced will not show up but that seems a rather small case we can then fix.

Thanks for the fix. 
closes #2579

Jguer (approved):

Issue #2780, opened 2026-02-06
Add support for contextual help flag


Tiagoquix:
When I use `pacman -S --help`, it only displays the relevant flags for the `-S` option.

But when I use `yay -S --help`, it displays the same as if `pacman --help`/`yay --help` was used.

Could you please add support for contextual help flag so it matches default pacman behavior?

Thanks!

PR #2921, opened 2026-07-19
Add contextual help to Yay operations

jdholtz:
This is a proposal to add #2780. I thought this would be a good addition to yay because it's exactly how pacman operates. Happy to discuss this further too.

Instead of `yay -<op> --help` printing the entire help message, it will now print a contextual help message based on the operation being performed. 

The top-level help message shows available operations and global flags as before. It no longer shows operation-specific flags and `pacman -S --help` output. Like the pacman output, it shows that you can get operation-specific help by running `yay -<op> --help`.

The operation-specific help message shows the operation's flags and all pacman operation flags pass through to pacman.

Specifying `-V` and `--version` now always shows the version rather than having `yay -V --help` show the help message.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -26,6 +26,59 @@ import (
 	"github.com/Jguer/yay/v13/pkg/vcs"
 )
 
+type helpOption struct {
+	flags       string
+	description string
+}
+
+type operationHelp struct {
+	short   string
+	long    string
+	targets string
+	options []helpOption
+}
+
+var yayOperationHelp = []operationHelp{
+	{"B", "build", "[options] [dir]", []helpOption{
+		{"-i --install", "Build and install a PKGBUILD"},
+	}},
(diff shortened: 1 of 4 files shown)


Jguer (approved):

Jguer:
I agree @jdholtz , let's merge it.

Questions unanswered for the future:
- How to ensure this is kept up to date
- Does pacman support localization of these and should we do the same

jdholtz:
> Questions unanswered for the future:
>
>  - How to ensure this is kept up to date
>  - Does pacman support localization of these and should we do the same

For keeping up to date, I don't think the process would necessarily because of this PR. The options still need to be updated in this help text, the man page, and the completions. Since that's so many places, I don't see a simple static check being feasible. That'd be a good option for an AI reviewer though.

For localization, yes pacman does support localization. Since we support localization in most (all?) text areas, we should support it for the help message too. I created #2944 to solve that.

Issue #1243, opened 2020-04-17
Git diffs use pagination inconsistently


ZimbiX:
#### Affected Version

yay v9.4.6 - libalpm v12.0.1

#### Issue

When I'm reviewing AUR package changes while running `yay`, the diffs are paginated inconsistently. If there's a package with a diff longer than the screen height, it'll show only that diff in what seems to be `less`. When I press Q to continue reviewing, and what remains is a bunch of packages with small diffs, they end up in the scrollback, and I have to scroll back up to see them all, trying to work out where I was up to.

It seems like the current functionality is something akin to sequentially running `git diff ...` in each package's repo, which has the effect of each one separately being possibly paginated.

I think the desired functionality would be to collect all the diffs, and then paginate that. This could be done by adding the `--no-pager` flag to the Git command (`git --no-pager diff`), and piping that into `less` (but then needing to use the `--color` Git flag to keep colour enabled, and `less -R` to interpret ANSI colour safely).

Also, including titles to separate the packages diffs more clearly would be great.

ZimbiX:
Right, sorry, I didn't realise this was addressed in the readme:

> #### Sometimes diffs are printed to the terminal, and other times they are paged via less. How do I fix this?
> 
> Yay uses `git diff` to display diffs, which by default tells less not to
> page if the output can fit into one terminal length. This behavior can be
> overridden by exporting your own flags (`export LESS=SRX`).

Using this workaround, I think I prefer `LESS=SR`, which forces each package's diff to be separately paginated on a clean screen, no matter the length.

A quick improvement could be for Yay to default the `LESS` environment variable to `SR` if it's unset. Although this could make it tricky to quit early if you disagree with the content of a diff.

Still, I think it would be preferable to show all the diffs at once, separated by title.

For now at least, I will be adding this alias to my dotfiles:

```bash
alias yay='LESS=SR yay'
```

HaleTom:
I think the option you're looking for is:
```
       -F or --quit-if-one-screen
              Causes less to automatically exit if the entire file can be displayed on the first screen.
```
If `less` were invoked with `-+F`, this would reset this to the default value. (I tested, it works on a simple file view)

So, maybe you could invoke `git` as `LESS="$LESS -+F" git diff ...`

Because invoking as `LESS="$LESS -+F" yay ...` sadly doesn't work.

Perhaps each call to git diff needs to be piped to `less -+F` rather than being printed directly to the screen?

grandchild:
[...]

PR #2933, opened 2026-07-25
fix(diff): place --color flag before pathspec separator

danny-kuehn:
Since #2905, PKGBUILD diffs lose color even with `Color` enabled in pacman.conf.

`collectPkgbuildDiffs` appends `--color=always`/`never` after the `--`
separator, so git treats it as a pathspec and ignores it. This predates
#2905 but was masked: `Show()` gave git a TTY, so auto color worked. With
`Capture()` git sees a pipe and disables color.

Fix: append the color flag before the range and pathspecs.

Tested: `TZ=UTC make test`, `make lint`, and manual diff review with color
on/off. Related: #2905, #1243.

### DIFF
--- a/pkg/menus/diff_menu.go
+++ b/pkg/menus/diff_menu.go
@@ -78,16 +78,17 @@ func collectPkgbuildDiffs(ctx context.Context, cmdBuilder exe.ICmdBuilder, logge
 			}
 		}
 
-		args := []string{
-			"--no-pager", "diff",
-			start + "..HEAD@{upstream}", "--src-prefix",
-			dir + "/", "--dst-prefix", dir + "/", "--", ".", ":(exclude).SRCINFO",
-		}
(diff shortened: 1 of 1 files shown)


Jguer:
Awesome find @danny-kuehn , thanks for fixing this.

Jguer (approved):

radon-at-beeper:
Just for anyone else who finds this: the behavior still changes compared to before https://github.com/Jguer/yay/issues/1243, when diffs would be colorized under `git config color.ui auto`, even if `Color` was not enabled in `pacman.conf`. Now, it is required to enable `Color` in `pacman.conf` (just uncomment the line that says `Color`), if you want color in Yay diffs.

This is arguably more consistent, although it is surprising since the behavior changed, and some people might expect that configuring Git to color diffs would also apply to Git diffs within Yay.

PR #2936, opened 2026-07-27
feat: next add minage option to prefer older package versions

vlasov-y:
Add --minage option so upgrades prefer package versions that are at least N days old. This lets users avoid pulling in freshly released builds that had no time to get bug reports.

- Track a last-modified timestamp on synced packages to compute age.
- On sysupgrade or dry-run print, walk the dependency graph and, for each package younger than the minimum age, keep the local version, downgrade to an older cached or AUR build, or skip it along with its dependents.
- Fetch older AUR versions by checking package git history and pin the resolved commit before build.
- Install downgraded sync packages from the local pacman cache and restore their original install reason.
- Report kept, downgraded, and skipped packages before the actual upgrade runs.
- Exclude minage-adjusted packages from pacman own upgrade pass so the chosen older versions are not overwritten.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -79,6 +79,7 @@ Permanent configuration options:
 
     --requestsplitn <n>   Max amount of packages to query per AUR request
     --completioninterval  <n> Time in days to refresh completion cache
+    --minage      <days>  Prefer package versions at least this old on upgrade (0 disables)
     --sortby    <field>   Sort AUR results by a specific field during search
     --searchby  <field>   Search for packages using a specified field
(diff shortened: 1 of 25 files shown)


vlasov-y:
image
image

vlasov-y:
@Jguer will appreciate your review :pray:

Jguer:
Hey @vlasov-y , since this feature can be implement via hooks it won't be merged to keep yay's core lean. 
Thanks for submitting the PR and do feel free to contribute to the hook design if you think there's something missing compared to this PR

PR #2944, opened 2026-08-01
feat(translate): add translation for help message

jdholtz:
A follow-up to #2921. Allow for the help message to be localized. This changes the help message printing so that every message is wrapped in `gotext.Get()`.

There are a couple things different than the previous help message which I can revert if the old way is preferred.
1. The flag arguments (`<xxx>`) are no longer aligned with each other. This aligns with pacman's help message and makes the code a tiny bit simpler.
2. Since pacman prints global options for every `pacman -<op> --help` message but yay doesn't, I added the message `use 'yay {-h --help}' for global options` to operation-specific help messages (maybe add `without an operation`?). This matches the message in the global menu: `use 'yay {-h --help}' with an operation for available options` and I think it is a better option than cluttering every operation-specific message with the global options.

I did not run `make locale` as that would create thousands of lines of diff, but I think that should be run to update the .po files (I'm not exactly sure how they work). Note that when running `make locale`, it only changes `en.po` as `mv po/default.pot po/en.po` should be a `cp` if I understand correctly.

Last, the only misaligned flag is `--completioninterval`'s arg:
```
    --requestsplitn <n>   Max amount of packages to query per AUR request
    --completioninterval <n> Time in days to refresh completion cache
    --sortby <field>      Sort AUR results by a specific field during search
```
If the following is preferred, I can change the description to be a newline when arguments overrun the limit (what pacman does) or just shift all the other descriptions over by 4 characters:
```
    --requestsplitn <n>   Max amount of packages to query per AUR request
    --completioninterval <n> 
                          Time in days to refresh completion cache
    --sortby <field>      Sort AUR results by a specific field during search
```


Jguer:
Part of me wishes we had a "cobra like" https://cobra.dev/docs/tutorials/12-factor-app/ cmd for our cli pattern. 
Thanks @jdholtz , I'll get this merged up and make the `en.po` generation.

Jguer:
the misaligned flag I leave to your best judgement if you want to change it

Jguer (approved):

jdholtz:
> Part of me wishes we had a "cobra like" https://cobra.dev/docs/tutorials/12-factor-app/ cmd for our cli pattern.

Yeah that would be nice, although the it is probably just more hassle at this point to migrate. We probably could mimic pacman's cli using cobra's customization, but at a certain point that'd probably be almost as much work as just maintaining the current cmd.

> the misaligned flag I leave to your best judgement if you want to change it

I added a solution in #2978.

PR #2945, opened 2026-08-01
feat: add minage lua hook example for aur upgrade filtering

vlasov-y:
- Add a lua hook example that excludes aur and devel upgrades newer
than a configurable minimum age.
- Document the hook limitations, including missing repo timestamps,
pulled dependency handling, and lack of version downgrade support.
- Reference the new example from init.lua comments and lua hook docs
so users can discover and enable it.

### DIFF
--- a/doc/examples/minage.lua
+++ b/doc/examples/minage.lua
@@ -1,0 +1,90 @@
+-- Exclude AUR upgrades younger than a minimum age (minage policy).
+--
+-- Enable from init.lua:
+--   require("hooks.minage")
+-- after copying this file to <config_dir>/hooks/minage.lua
+--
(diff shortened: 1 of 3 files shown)


vlasov-y:
@Jguer followed your reply https://github.com/Jguer/yay/pull/2936#issuecomment-5134038018

Jguer:
Hey @vlasov-y  how does this differ from https://github.com/Jguer/yay/blob/next/doc/examples/recently_modified.lua

vlasov-y:
> Hey @vlasov-y how does this differ from https://github.com/Jguer/yay/blob/next/doc/examples/recently_modified.lua

Probably you are right, the only diff is filtering devel packages as well, but that is minor. I will close the PR.

PR #2947, opened 2026-08-03
Add warnings for recent AUR maintainer changes 

benjamindaines:
image
image

Adds a warning to AUR packages that have recently changed maintainers.
Caches packages-meta-ext-v1 for maintainer and lastModified. Refreshed
and diff'd upon sync or for individual packages upon transaction.

Currently flags any maintainership change within the last 30 days, pulls
new maintainer database if it hasn't done so within the last 24 hours.
Both are configurable.

Covers:
- Maintainer A -> Maintainer B
- Orphan -> New Maintainer
- Orphan -> Malicious Maintainer -> Orphan (flags an orphan that has
  been recently updated, yet still has no maintainer)

Should be fairly easy to expand upon and add more conditions to be
flagged.

Minimally tested, as the AUR is currently frozen.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -391,6 +391,10 @@ func handleSync(ctx context.Context, run *runtime.Runtime, cmdArgs *parser.Argum
 
 	switch {
 	case cmdArgs.ExistsArg("s", "search"):
+		run.RefreshAdoption(ctx, cmdArgs.ExistsArg("y", "refresh"))
+		if b, ok := run.QueryBuilder.(*query.SourceQueryBuilder); ok {
+			b.SetAdoptions(run.AdoptionStore.Active())
+		}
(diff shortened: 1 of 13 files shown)


Jguer:
Thanks @benjamindaines for the PR. This can be done through hooks as it stands and won't be added to the yay core
https://github.com/Jguer/yay/blob/next/doc/examples/maintainer_change.lua

benjamindaines:
Yea, no worries. I didn't even think to check for existing hooks.  However, regardless of how it's implemented, it's a good conversation to have. 

At this point the AUR has proven itself to be wide open to being taken advantage of, and we really haven't seen any effective policy changes to mitigate that. Honestly,  are there really any meaningful changes to be made that wouldn't just result in it being essentially a managed repo?  

I think that moves the line of defense to the Helpers and I see this as an essential security feature worth being part of core yay.

